PR #25, opened 2022-05-16
Copy over verify_tag logic

ra0x3:
### Description
- This PR copies over the `verify_tag.sh` logic using [fuel-core's verify_tag.sh](https://github.com/FuelLabs/fuel-core/blob/master/.github/workflows/scripts/verify_tag.sh) as a reference 

### Testing steps
- [ ] This addition will only run on a `publish` event.
  - This logic actually already existed in a job, but the job never triggered, so there was no error thrown for `verify_tag.sh` not existing.
  - This PR simply just adds the script
  - Should be a pretty easy PR to eyeball


adlerjohn (approved):

mitchmindtree (approved):
Good catch!

PR #31, opened 2022-05-20
Refactor CI into concurrent jobs. Add mdbook and udeps checks.

mitchmindtree:
Also adds `--locked` to several cargo jobs to ensure that the `Cargo.lock` file is valid.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,73 +14,190 @@ env:
   REGISTRY: ghcr.io
 
 jobs:
-  build:
+  cancel-previous-runs:
     runs-on: ubuntu-latest
-
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.9.1
         with:
           access_token: ${{ github.token }}
 
-      - name: Checkout repository
-        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


mitchmindtree:
OK, looks like running these jobs concurrently cuts the overall CI duration from ~25 mins down to around ~15 mins. Still really long, but better!

I've removed the `cargo udeps` check in favour of a future PR that checks packages more granularly as there were some false positives in there. #62.

We'll also want to test the `--all-features` set, but there were some issues with this so I've opened #63.

ra0x3 (dismissed):
- Looks good as is
- https://github.com/FuelLabs/fuel-indexer/issues/64 can be handled in a follow up (if not this PR)

mitchmindtree:
Sorry @ra0x3, just noticed the `publish` job is behaving a little strangely (triggered when it shouldn't have, but failed anyway) and I've tried properly forwarding on the success state in the `validation-complete` job. Will re-ping once ready again :)

mitchmindtree:
OK, fixed the odd `publish` behaviour, I think this is good to go.

@adlerjohn could you change the "Required" job from the old `build` job to the new `validation-complete` one?

PR #32, opened 2022-05-23
Update deps, move wasm indexers over

tjsharp1:
Dependencies updated, this still needs a few things from `fuels` to become optional dependencies since there are crates that will not compile to wasm, like `socket2`.
The libraries that won't compile to wasm are `fuels-contract`, `fuels-signers` and `fuels-test-helpers`. 
If we can set these to optional deps, then I can add `no-default-features` and everything compiles.


mitchmindtree (commented):
Nice, excited to see this happening!

> The libraries that won't compile to wasm are fuels-contract, fuels-signers and fuels-test-helpers.

Are we able to workaround this in the meantime by using only the `fuels-*` crates that we need without using `fuels` directly? If that helps to land this PR, that could be a beneficial step.

Either way, we should also:

1. Open an issue with this list of known `fuels-*` crates that are causing issues with WASM compilation. We should identify exactly which types/APIs/crates within each are the source of the WASM incompatibility.
2. Open an issue at `fuels-rs` about these crates to notify the SDK team that links to the issue we create here. We should also close the old `no_std` issue there as we no longer require it.
3. If everyone on the SDK team is busy with existing issues, we should address the incompatibilities in the `fuels-*` crates ourselves as this is a blocker for `fuel-indexer`. We can ask for Qs and for guidance from the SDK team in the `fuels-rs` tracking issue if need be.
4. Add a basic WASM target compilation check to the `fuels-rs` repo CI to ensure we maintain compatibility in the future.

ra0x3:
@tjsharp1 @JoshuaBatty @mitchmindtree Let me know what you think about this
- PR is pretty big. A non-insignificant amount of the added code is examples (thanks @tjsharp1 🥳 )
- Obviously [@mitchmindtree's comment here](https://github.com/FuelLabs/fuel-indexer/pull/32#pullrequestreview-982444937) is probably the biggest outstanding question
  - This is super interesting
  - In an ideal world, fuel-indexer would sit directly on top of the SDK and not have its own adjacent SDK-ish logic
  - I agree with the approach mentioned by @mitchmindtree in terms of working with the SDK team to figure out how to resolve the SDK/WASM compatibility (with this having some priority)
- @mitchmindtree I agree _ideally_ the `api_server` package would be reviewed separately 
  - However, it's footprint seems relatively small in this PR, so given the blocking nature of this work, I could see the case for letting it stay
- Personally, I'd prefer the "merge what we have" (as long as its in some workable state), then immediately start working in smaller PRs to get to our _ideal_ state
  - Versus keeping such a big PR hanging on the vine for too long
  - If we could somehow get CI passing that would be a big help -- not sure how much additional work that would be in this PR

Cargo.toml
```diff
@@ -2,9 +2,11 @@
 # Use the new resolver to prevent dev-deps and build-deps from enabling debugging or test features in production.
 resolver = "2"
 members = [
+   "api_server",
```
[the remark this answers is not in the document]
tjsharp1: yeah, this just got moved from indexer.

tjsharp1:
> 

okay, issue opened. https://github.com/FuelLabs/fuels-rs/issues/317
I think one way or another we will need a change in fuels-rs to land this PR, the codegen macro will output `use` statements that have `fuels::core` instead of `fuels_core`.
The changes are pretty straightforward, I already had to do it, so maybe I just submit that as a PR as well.

tjsharp1:
Looks like CI is failing on ` git ls-files | grep Cargo.toml$ | xargs --verbose -n 1 cargo-toml-lint`, it says it can't find manifest?
Not sure what's happening here, but that step is passing on my machine.

mitchmindtree:
> I agree ideally the api_server package would be reviewed separately. However, it's footprint seems relatively small in this PR, so given the blocking nature of this work, I could see the case for letting it stay.

@ra0x3 yes I'm happy for this to stay as it's just being moved from another location which tj clarified is the case [here](https://github.com/FuelLabs/fuel-indexer/pull/32).

> Looks like CI is failing on git ls-files | grep Cargo.toml$ | xargs --verbose -n 1 cargo-toml-lint, it says it can't find manifest?

@tjsharp1 Yeah that cargo lint step works on my machine too, I tried both `0.1.0` and `0.1.1` but both appear to work fine :thinking: Unless someone else chimes in with a solution in the meantime, I say we remove the Cargo toml lint step for now as landing this is more important. I'll try to re-introduce the lint as a part of #31 after this lands.

> I think one way or another we will need a change in fuels-rs to land this PR, the codegen macro will output use statements that have fuels::core instead of fuels_core.

I've opened up https://github.com/FuelLabs/fuels-rs/issues/318 to clarify our needs around WASM support from `fuels-rs`. If you could provide a bit more info there on what dependencies of those `fuels-*` crates are not compatible with the WASM target that would be a big help, i.e. is it just socket2 or are there more?

If https://github.com/FuelLabs/fuels-rs/issues/317 makes it easier to make progress in the meantime then that sounds OK to me.

tjsharp1:
> I've opened up [FuelLabs/fuels-rs#318](https://github.com/FuelLabs/fuels-rs/issues/318) to clarify our needs around WASM support from `fuels-rs`. If you could provide a bit more info there on what dependencies of those `fuels-*` crates are not compatible with the WASM target that would be a big help, i.e. is it just socket2 or are there more?
I think the compile failures would prevent further crates from being compiled, so not sure we can get that full list.
One of the major ones is fuel-gql-client, which has sockets as a sub-dependency, I don't see a need for having the fuel client compiled to WASM so we could leave that out.
But I think what we need from fuels is very basic anyway, all I need to do is have the type definitions from the contract ABI json and I need to be able to deserialize them once I've passed them into WASM memory.

tjsharp1:
@mitchmindtree I have a PR for the fuels-rs fix here: https://github.com/FuelLabs/fuels-rs/pull/351

Currently pointing the indexer deps at that branch to get CI passing. Looks like if we get that one merged, this will be unblocked.

indexer/Cargo.toml
```diff
@@ -34,21 +34,19 @@ tracing = "0.1"
...
+fuel-core = { version = "0.6.4", features = ["test-helpers"] }
+fuel-gql-client = { version = "0.6.4", features = ["test-helpers"] }
 fuel-indexer = { path = "../lib" }
 fuel-storage = { version = "0.1" }
-fuel-types = { version = "0.3", features = ["serde-types"] }
-fuels-abigen-macro = { version = "0.2" }
-fuels-rs = { version = "0.2" }
+fuel-types = "0.3"
+fuels = "0.12"
+fuels-abigen-macro = { version = "0.12" }
```
mitchmindtree, line 47: Looks like `fuels` and `fuels-abigen-macro` are on `0.12` here whereas other crates in this workspace are pointing at `0.15` - should we update these too?

It might be worth double-checking the `Cargo.lock` for any dupes that can be resolved. Normally you can spot these by checking the `dependencies` lists within `Cargo.lock`, as duplicate crates need to have their semver version appended for disambiguation.

mitchmindtree (commented):
Looks like this is super close :partying_face: I've left one more comment that we might want to address first, otherwise looking good!

derive/Cargo.toml
```diff
@@ -19,6 +19,7 @@ syn = { version = "1.0", features = ["full"] }
 
 [dev-dependencies]
 fuel-indexer = { path = "../lib" }
-fuels-core = { version = "0.2" }
+fuels = { version = "0.12" }
```
mitchmindtree, line 22:
```suggestion
fuels = { version = "0.15" }
```

mitchmindtree (approved):
LGTM :partying_face:

ra0x3 (approved):
🚢

PR #37, opened 2022-05-25
Update CI

ra0x3:


tjsharp1:
@ra0x3 
I think there's a few things we get out of using WASM:
- we handle all the boilerplate that would be common to any indexer, like database management, setting up a fuel client, etc. That would simplify development for the app developers, since they don't have to think about those bits.
- while allowing the app developer to write arbitrary indexing code, we can execute it in a sandbox.
The functions themselves don't need to do much, it's really just taking the data types out of the contract ABI and allowing the developer to map them onto types in their graphql schema.
The only thing that we need out of fuels-rs is to generate rust structs from the abi.json file so we can compile them down to wasm, and we need a way to deserialize those types (we have to allocate wasm memory and then copy the bytes in). When we have that, we can unpack those structures in the WASM handler code and from there we can insert it into the database tables. So the `wasm_abigen` macro just needs to output data structures, the derives for `Tokenizable` and related stuff, but not much more. I think we can probably just make the output of that macro not depend on `fuels::core` and instead do `fuels_core`. Then we can get rid of the `fuels` dependency in the wasm modules and just have the `fuels_core` dependency we'll be good to go.

mitchmindtree:
@ra0x3 it looks like the intention for this PR is to be based on @tjsharp1's work in #32 - is that correct?

If so, would you mind opening this PR against TJ's `update-deps` branch instead of the `master` branch?

That will make it much easier to review the diff, and for TJ to merge the changes into his PR if desired.

Cargo.toml
```diff
@@ -6,9 +6,9 @@ members = [
    "derive",
    "lib",
    "indexer",
-   "indexer-samples/*",
    "schema"
 ]
+exclude = ["examples"]
```
mitchmindtree, line 11: Could you elaborate a little on why we want to remove these from the main workspace? It would be nice to share a workspace if possible in order to share build artifacts, otherwise every example will require it's own target directory and building the entire set of `fuel-indexer` deps each time.

.github/workflows/ci.yml
```diff
@@ -34,7 +34,19 @@ jobs:
...
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock', '**/Cargo.toml') }}
+            ${{ runner.os }}-cargo-
```
mitchmindtree, line 49: Could you elaborate a little on what these changes do? Were we having some kind of issue with `rust-cache` in CI?

mitchmindtree (commented):
Great job on getting CI working for #32 @ra0x3!

I've left a couple comments but after addressing those it looks like this should be ready for review from @tjsharp1 for merging into #32? Do you want to convert this from draft?

Cargo.toml
ra0x3: @mitchmindtree So yea,
- When I try to `cargo check` in an example directory, without the examples exclusion, I get the `current package believes it's in a workspace when it's not` error
- Further the `Run Cargo.toml linter` CI check fails because there's no `Cargo.toml` in the `examples/` directory root (if the exclusion is not included)
- I noticed that we also do this (exclude examples) [in the sway repo](https://github.com/FuelLabs/sway/blob/master/Cargo.toml#L25) so I thought it made since to exclude here as well
- I also figured we'd build these example projects on CI later so that we can ensure they're functional/workable separately from the fuel-indexer project, but let me know if you think we should be including these examples

ra0x3: - Ah, now seeing that `examples` in the `workspace` does not work. But `examples/*` does 👌 
- However, given the state of the project (still a WIP) I would want to see what you think about temporarily keeping `examples/*` out of the `workspace` until https://github.com/FuelLabs/fuel-indexer/pull/32 is merged?
  - `cargo` complains a bit about the examples, as is

mitchmindtree: > cargo complains a bit about the examples, as is

Can you post the errors? Or remove this exclude line and re-add `examples/*` to the members list so we can see them show up in CI? That way we might be able to help resolve whatever errors are occurring.

I'd prefer not to exclude the examples if we can avoid it as it will result in a lot of `Cargo.lock` noise and greatly increase the CI build times.

ra0x3: @mitchmindtree Sounds good. Feel free to check the CI errors

adlerjohn (commented):
CI failing

ra0x3:
> CI failing

@adlerjohn Yea we're gonna play around with it and see if we can get it to pass while including `examples/` in the workspace. But yes, I will return it to Draft now 😅

mitchmindtree:
@ra0x3 it looks like TJ has since amended the cause of this CI failure on the `update-deps` branch, which looks like it might have been caused by accidentally including some local paths in a couple of the fuel dependencies. It looks like merging `update-deps` again should solve this?

ra0x3:
@mitchmindtree Now that I think about it, since TJ's PR is passing on CI, I'm happy to close this one out. Any further CI updates we want to throw in, I think we can do after TJ's PR lands

PR #38, opened 2022-05-28
PoC for non-WASM handler

ra0x3:


examples/simple-handler/src/schema.rs
```diff
@@ -0,0 +1,327 @@
+extern crate proc_macro;
```
ra0x3, line 1: This is just a copy/paste of `derive/schema.rs`

PR #44, opened 2022-05-31
Add non-wasm execution

ra0x3:
### Description

### Testing Steps

### Reviewer Notes


ra0x3:
Closing this as WASM is unfortunately pretty baked into the indexer. 😭

mitchmindtree (commented):
Just had a quick first look, epic stuff!

I've left a few comments. I think it might also be easier to review some of this after addressing these and after we land #32 and merge it in. Doing so might also help to resolve some of the version differences etc.

Also looks like a `.DS_Store` file might have slipped in.

Issue #54, opened 2022-06-07
Clarify which tool version(s) are preferred


ra0x3:
- The indexer uses quite a few tools (GraphQL, Postgres, etc)
- We should clarify what our preferred/supported version of these tools are

Additional context
- Currently, the `diesel` migrations do not work as is, on Postgres major 14
  - There is no mention of this and you only see it when you try `diesel migration revert`

adlerjohn:
Shouldn't that be captured by dependency versions? Or are you talking about system dependencies?

ra0x3:
> Shouldn't that be captured by dependency versions? Or are you talking about system dependencies?

@adlerjohn system dependencies (e.g., Postgres)

PR #60, opened 2022-06-14
Add non-wasm execution

ra0x3:
More up-to-date version of https://github.com/FuelLabs/fuel-indexer/pull/44

### Description

### Testing Steps

### Reviewer Notes


mitchmindtree (commented):
Small reminder to remove the left-over `Cargo.lock`s from the examples directory now that they're back in the main workspace.

Otherwise I think some of my comments on #44 still stand w.r.t. versions and waiting for #32 to land first :+1:

ra0x3:
> Small reminder....

Thanks. Will push this change (still understanding workspaces and such) 👌🏽. 

I'll wait til #32 is merged before putting this in review (will make it a draft now)

PR #61, opened 2022-06-14
Update out-of-date lock file following #32

mitchmindtree:
This slipped through in #32 as we don't yet use `--locked` in the current CI. I've added the `--locked` check in #31, but need to land this first.


JoshuaBatty (approved):
Nice, good spot.

Issue #58, opened 2022-06-13
Allow receipts to implement Tokenizable


ra0x3:
**This issue is only a reference issue, since the actual fix would come from fuel-tx**

- This functionality can be seen [here](https://github.com/FuelLabs/fuel-indexer/pull/44/files#diff-a4823faf4a276409ae568a0298b89751bddfaae8694fc1db61750890de0cc5d3R86)
- It would be great if `Receipt`s implemented the `Tokenizable` trait
- Ideally, the user would just call `Receipt::from_token` and would be returned a `Receipt` which could be used to grab the `data` via `Receipt.data()`
- This prevents the user from having to keep track of each `Token` in a `Receipt`, and also cuts down on code written

### Example 1: Deserializing data in handler functions
Ideal functionality

```rust
// Where from_token() takes in a serialized receipt and returns a receipt struct
let receipt: Receipt = Receipt::ReturnData::from_token(data)?;
```

Current functionality

```rust
let tokens: Vec<Token> = ABIDecoder::new()
    .decode(
        &[
            ParamType::B256,
            ParamType::U64,
            ParamType::U64,
            ParamType::B256,
            ParamType::Array(Box::new(ParamType::U8), 16),
            ParamType::U64,
            ParamType::U64,
        ],
        data,
    )
    .expect("Bad encoding!");
```

### Example 2: Serializing data when receiving transaction events

Ideal functionality

```rust
Receipt::ReturnData {
    id,
    ptr,
    len,
    digest,
    data,
    pc,
    is,
} => {
    let token: Token = Token::Struct::into_token([id, ptr, len, digest, data, pc, is])?;
    let args = ABIEncoder::new()
        .encode(&[token.clone()])
        .expect("Bad Encoding!");
}
```

Current functionality

```rust
Receipt::ReturnData {
    id,
    ptr,
    len,
    digest,
    data,
    pc,
    is,
} => {
    // TODO: might be nice to have Receipt type impl Tokenizable.
    let token = Token::Struct(vec![
        id.into_token(),
        ptr.into_token(),
        len.into_token(),
        digest.into_token(),
        data.into_token(),
        pc.into_token(),
        is.into_token(),
    ]);

    let args = ABIEncoder::new()
        .encode(&[token.clone()])
        .expect("Bad Encoding!");
```

PR #65, opened 2022-06-14
Add non-wasm execution

ra0x3:
More up-to-date version of https://github.com/FuelLabs/fuel-indexer/pull/44

### Description

### Testing Steps

### Reviewer Notes


api_server/src/lib.rs
```diff
@@ -61,7 +61,7 @@ async fn query_graph(
     Extension(manager): Extension<Arc<RwLock<SchemaManager>>>,
 ) -> (StatusCode, Json<Value>) {
     if !manager.read().await.contains_key(&name) {
-        if let Ok(Some(schema)) = load_schema(&database_url, &name).await {
+        if let Ok(Some(schema)) = load_schema_wasm(&database_url, &name).await {
```
ra0x3, line 64: You'll notice regular name changes such as this. Where "wasm" was added for clarity

examples/simple-non-wasm/programs/counter-rs/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+fuels-abigen-macro = "0.15"
+json = "0.12"
+log = "0.4"
+pretty_env_logger = "0.4"
+rand = "0.8"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+tokio = { version = "1", features = ["full"] }
+
+[workspace]
```
ra0x3, line 26: This lockfile is not included in the repo. This project needs not be included in the workspace (it's a separate little app only used for this specific example)

examples/simple-non-wasm/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let schema = manifest.graphql_schema().await?;
+
+    // In this example, we've already started our fuel node on another process
+    info!("Fuel node listening on {}", config.fuel_node_addr);
+    let api_handle = tokio::spawn(GraphQlApi::run(config.clone()));
+
+    // Create a new service to run
+    let mut service = IndexerService::new(config.clone())?;
+    let _ = service.build_schema(&manifest, &schema)?;
```
ra0x3, line 179: - For non-wasm execution, the schema needs to be loaded _before_ the executor is created with a database (this also works for wasm execution)

examples/simple-non-wasm/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+        ))
+        .handler(CustomHandler::new(
+            ReceiptEvent::LogData,
+            manifest.namespace.clone(),
+            "AnotherCount".to_owned(),
+            another_count_handler,
+        ));
+
+    // Add the executor to our service
+    if let Err(e) = service.add_executor(executor, "an_unused_field", manifest, false) {
```
ra0x3, line 200: You'll notice throughout I left fields untouched, such as `"an_unused_field"` in an attempt to change as little of the existing API as possible

indexer/src/executor.rs
```diff
@@ -34,6 +43,96 @@ pub struct IndexEnv {
...
+    manifest: Manifest,
+}
+
+impl CustomIndexExecutor {
+    pub fn new(db_conn: &str, manifest: Manifest) -> IndexerResult<Self> {
+        let events = HashMap::new();
+        let db = Arc::new(Mutex::new(Database::new(db_conn).unwrap()));
+
+        db.lock()
+            .expect("Lock poisoned from schema load")
```
ra0x3, line 59: Didn't improve the error messaging. Figured we could do this in a follow-up pass, in a smaller PR (as this one is rather large, as is)

indexer/src/executor.rs
```diff
@@ -90,29 +189,31 @@ impl IndexExecutor {
...
+impl Executor for WasmIndexExecutor {
     /// Restore index from wasm file
-    pub fn from_file(_index: &Path) -> IndexerResult<IndexExecutor> {
+    fn from_file(_index: &Path) -> IndexerResult<Self> {
         unimplemented!()
     }
 
     /// Trigger a WASM event handler, passing in a serialized event struct.
-    pub fn trigger_event(&self, event_name: &str, bytes: Vec<Vec<u8>>) -> IndexerResult<()> {
+    fn trigger_event(&self, event_name: ReceiptEvent, bytes: Vec<Vec<u8>>) -> IndexerResult<()> {
```
ra0x3, line 209: Changed the input type for `trigger_event` to be a `ReceiptEvent`
- The idea is that a `ReceiptEvent` is just a type, whereas `fuel_tx::Receipt` is a value
- Fuel contracts output values (`fuel_tx::Receipt`) and we trigger execution based on the type of `ReceiptType`
  - We can obviously add filters later to narrow down what type of specific `Receipt` we want to execute on

indexer/src/executor.rs
```diff
@@ -209,7 +310,7 @@ mod tests {
                 .expect("Failed to encode"),
         ];
 
-        let result = executor.trigger_event("an_event_name", encoded);
+        let result = executor.trigger_event(ReceiptEvent::Log, encoded);
```
ra0x3, line 313: Since this was just previously a dummy value, I used `ReceiptEvent::Log` and updated the tests accordingly

indexer/src/handler.rs
```diff
@@ -0,0 +1,62 @@
...
+        match e {
+            ReceiptEvent::LogData => "LogData".to_owned(),
+            ReceiptEvent::Log => "Log".to_owned(),
+            ReceiptEvent::ReturnData => "ReturnDataa".to_owned(),
+            _ => "Other".to_owned(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
ra0x3, line 59: Unused

indexer/src/service.rs
```diff
@@ -151,7 +203,71 @@ impl IndexerService {
...
+                                ptr,
+                                len,
+                                digest,
+                                data,
+                                pc,
+                                is,
+                            } => {
+                                // TODO: might be nice to have Receipt type impl Tokenizable.
+                                let token = Token::Struct(vec![
+                                    id.into_token(),
```
ra0x3, line 223: Again, https://github.com/FuelLabs/fuel-indexer/issues/58 might avoid some of this boiler plate (but this isn't terrible since it's abstracted away from the user)

indexer/src/handler.rs
```diff
@@ -0,0 +1,62 @@
+use anyhow::Result;
+use fuel_indexer::EntityResult;
+use fuel_indexer_schema::type_id;
+use serde::{Deserialize, Serialize};
+
+pub type Handle = fn(Vec<Vec<u8>>) -> Result<Option<EntityResult>>;
```
ra0x3, line 6: - Might need some help here
- Ideally, this would be 
```rust
pub type Handle = fn(Vec<Vec<u8>>) -> Result<Option<Box<dyn Entity>>>;
``` 

OR 

```rust
pub type Handle = fn(Vec<Vec<u8>>) -> Result<Option<Box<T>>>;
```

- But since `T: Entity` isn't a safe object, I get a compiler error

lib/src/lib.rs
```diff
@@ -42,13 +41,29 @@ impl Logger {
     }
 }
 
+pub struct EntityResult {
```
ra0x3, line 44: - See [this comment](https://github.com/FuelLabs/fuel-indexer/pull/65/files#r897328575)
- This feels....hacky :(

schema/src/db/models.rs
```diff
@@ -74,29 +74,41 @@ impl GraphRoot {
...
         version: &str,
-    ) -> QueryResult<Vec<TypeIds>> {
+    ) -> QueryResult<Vec<TypeId>> {
         use gr::type_ids::dsl::*;
         type_ids
             .filter(schema_name.eq(name).and(schema_version.eq(version)))
             .load(conn)
     }
 
+    pub fn latest_version(schema_name: &str, conn: &PgConnection) -> QueryResult<String> {
```
ra0x3, line 100: `latest_version` returns the latest schema version for the given namespace

ra0x3:
@mitchmindtree @tjsharp1 @JoshuaBatty 

- Ping'ing for review
- I see CI is failing, but not so concerned about that atm since I'm sure there will be quite a bit of feedback
- I went through and left thoughts/comments on what I was thinking with some of my decisions, hopefully this helps the review
- Keep in mind this is my 2nd month of writing Rust -- so if you see something and wonder "Why'd he do it like that?" - chances are, I didn't know of a more idiomatic way
  - So definitely  feel free to comment on any best practices or leave any pro tips as well
- Acceptance criteria should possibly be code quality (obviously) and following the README steps to get similar output? 🤔
- @mitchmindtree So help me God I've tried to clean up all the lockfiles and what not 😆  -- let me know if I'm missing anything

examples/simple-non-wasm/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+        Ok(Self { count, timestamp })
+    }
+
+    fn into_token(self) -> Token {
+        Token::Array(vec![Token::U64(self.count), Token::U64(self.timestamp)])
+    }
+}
+
+// Implement Entity trait for object being indexed
+impl Entity for CountEvent {
```
ra0x3, line 54: Ideally there would be macro to automagically derive `CountEvent` from GraphQL schema and make it implement `Entity` (not too versed in macros myself). 

Similar to what's in [derive/macro](https://github.com/FuelLabs/fuel-indexer/blob/master/derive/src/schema.rs)

examples/simple-non-wasm/README.md
```diff
@@ -0,0 +1,81 @@
...
+9 directories, 13 files
+```
+
+## Usage
+
+#### Run migrations
+  - In this example we're using an `indexer` database owned by the `postgres` user without a password
+
+```bash
+DATABASE_URL="postgres://postgres@127.0.0.1:5432/indexer" diesel migration list --migration-dir=./schema/migrations
```
ra0x3, line 41: See https://github.com/FuelLabs/fuel-indexer/issues/54 if you run into issues here

mitchmindtree:
@ra0x3 If you could address the CI stuff first that would be a huge help! Otherwise we need to spend review time adding comments on things that the compiler or tests can probably pick up anyways.

> Keep in mind this is my 2nd month of writing Rust

:exploding_head: TIL! (or maybe forgot and relearned?) Great progress for being so new to Rust @ra0x3, it can definitely have a bit of a steep curve. I'll try better to keep this in mind!

ra0x3:
@mitchmindtree That's definitely fair. CI should be fixed now

examples/simple-non-wasm/programs/counter-rs/Cargo.toml
mitchmindtree: Is it possible to include this within the workspace regardless?

If we are to not include this under the same workspace, we'll need to build it from scratch (including all it's dependencies) when testing locally or during CI.

Having it under the same workspace means they can share build artifacts which can speed up builds a lot! It also ensures that this example will get tested when running `cargo test` etc from the root of the repo.

.gitignore
```diff
@@ -1,4 +1,9 @@
+.DS_Store
+
 **/target
 
 # Don't add the generated MDBook artifacts
 docs/book/
+
+# Any rust programs in examples/
+examples/simple-non-wasm/programs/counter-rs/Cargo.lock
```
mitchmindtree, line 9: We should add

```
out
```

to this `.gitignore`.

in order to disclude the `out` directory generated by `forc build` which just contains the compiler output artifacts. This should help to stop the `examples/simple-non-wasm/programs/counter/out/` from being included, as well as any future forc project output artifacts.

examples/simple-non-wasm/programs/counter/Forc.lock
```diff
@@ -0,0 +1,12 @@
...
+dependencies = []
+
+[[package]]
+name = 'counter'
+dependencies = ['std git+https://github.com/fuellabs/sway?tag=v0.14.5#60c626cf486d5c53c44d84db1ec81cb90174cad3']
+
+[[package]]
+name = 'std'
+source = 'git+https://github.com/fuellabs/sway?tag=v0.14.5#60c626cf486d5c53c44d84db1ec81cb90174cad3'
+dependencies = ['core']
```
mitchmindtree, line 12: It looks like this file was generated by forc `v0.14.5` - just a friendly reminder to update your installed `forc` version :)

indexer/Cargo.toml
```diff
@@ -14,8 +14,10 @@ cfg-if = "1.0"
 diesel = { version = "1.4", features = ["postgres", "serde_json"] }
 fuel-core = { version = "0.6.4" }
 fuel-gql-client = { version = "0.6.4" }
+fuel-indexer = { path = "../lib" }
```
mitchmindtree, line 17: Could you also add the `version` here? This will be needed when publishing the crate.

indexer/src/database.rs
```diff
@@ -165,7 +170,21 @@ impl Database {
...
+            false => self.schema[table]
+                .iter()
+                .filter_map(|col| {
+                    if col == "id" {
+                        None
+                    } else {
+                        Some(col.to_owned())
+                    }
+                })
+                .collect(),
```
mitchmindtree, line 184: Should be possible to simplify this branch a little to something like: 
```rust
            false => self.schema[table]
                .iter()
                .filter(|&col| col != "id")
                .cloned()
                .collect(),
```
I've leave the choice on whether it's acceptable to use `include_id` like this up to @tjsharp1.

indexer/src/executor.rs
```diff
@@ -34,6 +43,96 @@ pub struct IndexEnv {
...
+        let event = handler.event.clone();
+        match self.events.get_mut(&event) {
+            Some(handles) => {
+                handles.push(handler);
+            }
+            None => {
+                let h = vec![handler];
+                self.events.insert(event, h);
+            }
+        }
```
mitchmindtree, line 81: This match could be simplified to something like:

```rust
self.events.entry(event).or_insert_with(Vec::new).push(handler);
```

mitchmindtree (commented):
I've added a few comments related to Rust conventions, but would like to get @tjsharp1's thoughts on the design approach here.

examples/simple-non-wasm/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    // Create a new service to run
+    let mut service = IndexerService::new(config.clone())?;
+    let _ = service.build_schema(&manifest, &schema)?;
+
+    // Create a new executor to run on the service
+    let mut executor = CustomIndexExecutor::new(&config.database_url, manifest.clone())?;
+
+    // Add some handlers to the executor, in order to process events
+    executor
```
tjsharp1, line 183: Yeah, I think with the macro we can hide a lot of these details and make it look similar to the WASM stuff, where the developer only writes handlers.

indexer/src/executor.rs
tjsharp1: I'd say we should ship the whole list of receipts to the module that's interested in events from a contract, and then have the all boilerplate of receipt handling happen in the modules. The service side would then just maintain a generic subscription list for <contract type> -> <indexer module>  and/or <contract address> -> <indexer module> mappings. We keep the service code pretty general that way and the modules can get more specific. So trigger_event should maybe become private, and then we'd have a dispatch loop in the executor.

examples/simple-non-wasm/src/main.rs
```diff
@@ -26,119 +25,35 @@ pub struct Args {
...
+// Load structs from abigen
+abigen!(
+    Counter,
+    "examples/simple-non-wasm/programs/counter/out/debug/counter-abi.json"
+);
 
-    fn to_row(&self) -> Vec<FtColumn> {
-        vec![FtColumn::UInt8(self.count), FtColumn::UInt8(self.timestamp)]
-    }
+fn count_handler(data: Vec<u8>) -> Result<Option<Count>> {
```
ra0x3, line 36: - After the discussion [here](https://fuellabs.slack.com/archives/C02CX4JVCSJ/p1655834736318249) I feel like this is as simple as I can get a handler at this point. 
- The user will just have to know what `ParamType` they need
- If anyone has a way to automagically do this, I'm open to ideas

schema/src/db/tables.rs
```diff
@@ -316,15 +316,15 @@ mod tests {
     const CREATE_THING1: &str = concat!(
         "CREATE TABLE IF NOT EXISTS\n",
         " test_namespace.thing1 (\n",
-        " id bigserial primary key not null,\n",
+        " id bigint primary key not null,\n",
```
ra0x3, line 319: - Revert this per a conversation with TJ
- We should create an issue to make this concept explicit -- and add this to the markdown book

indexer/src/executor.rs
```diff
@@ -66,30 +76,33 @@ impl CustomIndexExecutor {
...
     fn from_file(_index: &Path) -> IndexerResult<Self> {
         unimplemented!()
     }
 
-    fn trigger_event(&self, event: ReceiptEvent, bytes: Vec<Vec<u8>>) -> IndexerResult<()> {
+    fn trigger_event(
+        &self,
+        event: ReceiptEvent,
+        _bytes: Vec<Vec<u8>>,
+        receipt: Option<Receipt>,
```
ra0x3, line 104: - I didn't want to change the `trigger_event` signature here because the already-merged WASM implementation expects things to be a certain way, and I'd rather add a few extra/unused params here than change something that was _just_ merged
- `event` is a type used to identify what event occurred, while `receipt` is the actual `Receipt` value that is used to get the `data`

ra0x3 (commented):
@tjsharp1 
- Left some comments when you're ready to take another pass
- There's still a _tiny_ bit of manual work regarding reconstructing the abi-json struct from the `data` payload in the `Receipt` (left a comment)
- Still using `ReceiptType` as the identifier for triggered events (we briefly discussed this, but might have to discuss again)
  - Btw, this `ReceiptType` could also work with your `manifest` implementation since `"LogData"` in the manifest file, would map to `ReceiptType::LogData`

.gitignore
ra0x3: @mitchmindtree  
- I'll defer to you on whether or not that's a forc convention (I can go either way).
- I only kept the out/ because it includes the abi-json (which is needed by the example)
  - I could definitely always copy the abi-json out of that folder, but wanted the user to see the full structure of the example

examples/simple-non-wasm/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+
+// Load structs from abigen
+abigen!(
+    Counter,
+    "examples/simple-non-wasm/programs/counter/out/debug/counter-abi.json"
+);
+
+fn count_handler(data: Vec<u8>, pg: Arc<Mutex<Database>>) -> IndexerResult<()> {
+    // Define which params we expect (using the counter-abi.json as a reference)
+    let params = ParamType::Struct(vec![ParamType::U64, ParamType::U64, ParamType::U64]);
```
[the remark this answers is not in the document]
ra0x3: @tjsharp1 Yea it's been ready for a few days (I probably asked for a re-review but didn't move it out of draft)

ra0x3:
@tjsharp1
- That's fair
- Yea, I think just the large slighly blocking nature of this PR lends itself to merging now, with quick follow ups to come
  - The follow up PRs will be smaller and easier to test/review
- Tracking the API exposure in https://github.com/FuelLabs/fuel-indexer/issues/83

ra0x3:
@tjsharp1 Need that ✅ before I can merge (looks like you requested changes)

PR #66, opened 2022-06-14
Add database for tests

tjsharp1:
I think this should allow us to add all-features. The postgres tests just need a pg backend to talk to.


.github/workflows/ci.yml
```diff
@@ -94,6 +96,39 @@ jobs:
...
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - run: cargo install diesel_cli
+      - run: |
+          cd schema
+          diesel migration run
+          cargo fmt
+      - uses: Swatinem/rust-cache@v1
```
mitchmindtree, line 123: I think if we put this cache step right before the `cargo install diesel_cli` step, it might also cache diesel too.

mitchmindtree (commented):
Nice one! Don't forget to add to the `needs` list of `validation-complete` job too

.github/workflows/ci.yml
tjsharp1: ok, I moved it up. Looks like it works though, tests are passing.

mitchmindtree: Nice :+1: just need to add the new job to the list of `needs` under the `validation-complete` job now.

ra0x3 (approved):

mitchmindtree (approved):
:ok_hand:

Issue #28, opened 2022-05-19
Investigate use of README vars


ra0x3:
- In the [Getting started](https://github.com/FuelLabs/fuel-indexer/blob/master/docs/src/getting-started/setup.md) section of the Fuel Indexer book, we are using vars such as `OPENSSL_LIB_DIR` and `OPENSSL_INCLUDE_DIR` for installation on Arch Linux
- We'd like to investigate this in order to determine whether or not we actually need these vars
- Documentation should be updated according to whatever conclusion is reached

deekerno:
Closing as it looks like this was completed by way of #77.

PR #72, opened 2022-06-22
Update mdbook outline

ra0x3:


docs/src/getting-started/system-dependencies.md
```diff
@@ -0,0 +1,52 @@
...
+### Installation
+
+On all architectures, install the following Fuel crates via `cargo`
+
+```bash
+cargo install forc fuel-core
+```
+
+## Other system dependencies
+
```
mitchmindtree, line 20: It might be worth mentioning generally what the main dependencies are and why they're needed before going on to how to install them. It could help us to keep track too in case we'd like to try and simplify the installation process in the future.

It looks like the main ones are:

- **cmake** + **clang** + **pkg-config** - I guess these must be for building the bindings for one of our crate deps?
- **OpenSSL** - We might not need this anymore. I think `forc` was updated a little while ago to link it statically. That said, there might be another dep that needs it?
- **postgresql** - DB stuff.

All that said, we can add something like this in a future PR if you'd prefer to land the rest first :+1:

docs/src/the-fuel-indexer.md
```diff
@@ -2,4 +2,6 @@
 
 ![Fuel Logo](./img/fuel.png)
 
-Fuel Indexer uses WASM to index transactions and state within a fuel network, allowing for high-performance read-only access to the blockchain for advanced dApp use-cases.
+The Fuel Indexer is a standalone binary that can be used to index various components of [the blockchain]. These indexable components include blocks, transactions, and [receipts](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/tx_format.md) and state within a fuel network, allowing for high-performance read-only access to the blockchain for advanced dApp use-cases.
```
mitchmindtree, line 5: Looks like this might be missing a link - maybe needs to be copied from the README again?

mitchmindtree (commented):
Looking really good @ra0x3, I'm liking the new layout!

docs/src/getting-started/system-dependencies.md
ra0x3: @mitchmindtree 
- Agreed, it would be better to clarify why these deps are needed. 
- https://github.com/FuelLabs/fuel-indexer/issues/28 was created to address this. 
  - I'd vote to merge as is, and put a deadline on #28 (I'll actually tackle that while I'm doing this doc work)

mitchmindtree (approved):
Lgtm!

adlerjohn (approved):

PR #87, opened 2022-06-28
Add gql api server config

ra0x3:
This will probably be overwritten by whatever comes out of https://github.com/FuelLabs/fuel-indexer/issues/89, but until that is decided, still need this cleanup for now

### Description
- This PR adds an example configuration file for the GraphQL API server
- This example config file serves as a means of describing what each config option does
  - Obviously we only have a few right now
- Renamed `api_server/configs` to `api_server/config`
- Removed the `sqlite_config.yaml` file and added it's differential variable to the example config
- Updates the mdbook accordingly

### Testing steps
- [ ] Start the GraphQL API server `RUST_LOG=debug cargo run --bin api_server -- --config ./api_server/config.yaml`
- [ ] Try out a query to ensure all works


api_server/config.yaml.example
```diff
@@ -0,0 +1,9 @@
+## The endpoint on which to serve the Fuel Indexer GraphQL API
+listen_address: 0.0.0.0:29987
+
+## Postgres connection URL. Note that this URI follows the format specified in
+## section 34.1.1.2 of https://www.postgresql.org/docs/current/libpq-connect.html
+database_url: postgres://postgres@127.0.0.1:5432/indexer
+
+## Local path to the local SQLite database
+sqlite_db_path: ../main.db
```
mitchmindtree, line 9: Awh, it's a pity the trailing `.example` disables yaml highlighting. Perhaps we should just name this `config.yaml` and then mention its purpose with a comment at the top of the file along the lines of:

```yaml
# The following is an example Fuel Indexer GraphQL configuration.
```

mitchmindtree (dismissed):
Looking good! utACK.

I guess we may want to revisit some of this after #89, but still worth having this in the meantime :+1:

ra0x3:
@mitchmindtree Re-requesting a quick review after upgrading to `rustc 1.62` (locally). Had to fix some `clippy` things

PR #94, opened 2022-07-01
Embedded migrations

tjsharp1:
First PR for #15. Need to move postgres schemas into a submodule, and embed & run migrations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,9 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-features --all-targets -- -D warnings
+          # After rustc 1.62 upgrade, the `Insertable` trait in the diesel crate triggers an 
+          # unused lifetime warning, so only allow that single warning until diesel is upgraded
+          args: --all-features --all-targets -- -D warnings -Aclippy::extra-unused-lifetimes
 
   mdbook-build:
     needs: cancel-previous-runs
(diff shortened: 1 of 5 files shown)


schema/src/db/models.rs.orig
```diff
@@ -0,0 +1,286 @@
...
+
+        Ok(result)
+    }
+}
+
+#[derive(Debug, QueryableByName)]
+pub struct EntityData {
+    #[sql_type = "Binary"]
+    pub object: Vec<u8>,
+}
```
JoshuaBatty, line 286: Is there any reason why we still need to keep this file? seems to have git conflicts baked into the file as well...

JoshuaBatty (commented):
Seems relativly straight forward, just left a comment asking why we need to keep `model.rs.orig` file.

schema/Cargo.toml
```diff
@@ -8,6 +8,7 @@ publish = false
 
 [dependencies]
 diesel = { version = "1.4", optional = true, features = ["postgres", "serde_json"] }
+diesel_migrations = { version = "1.4.0", features = ["postgres", "sqlite"] }
```
ra0x3, line 11: @tjsharp1 
- I'm assuming this is added `diesel_migrations` crate is responsible for the migration file changes [like this one here](https://github.com/FuelLabs/fuel-indexer/pull/94/files#diff-5ba27fcf941f34a3f45e0a5c999a2be4817bc9c1edd8593bd4b485be7dc693feR6) ?

ra0x3 (approved):

Issue #89, opened 2022-06-29
Should we have a single indexer config?


ra0x3:
- We currently use a few configuration files to run an indexer
  - A node config ([example](https://github.com/FuelLabs/fuel-indexer/blob/master/examples/simple-non-wasm/config.yaml))
  - An executor config ([example](https://github.com/FuelLabs/fuel-indexer/blob/master/examples/simple-wasm/manifest.yaml))
  - An api_server config ([example](https://github.com/FuelLabs/fuel-indexer/blob/master/api_server/configs/pg_config.yaml))
 
- Should we merge all these configs into a single `config.yaml` in the `fuel-indexer/` root?
  - Pros:
    - Users would just have to update a single file for config changes
    - Could neatly partition each config section into parts (e.g., "database", "api_server", etc)
  - Cons:
    - Config file might be slightly large once we get all the options in there (I'm personally OK with this)
    - [Insert other cons...]

Additional context
- Mdbook would have to be update according to whatever is decided
- [See this PR comment](https://github.com/FuelLabs/fuel-indexer/pull/108#discussion_r915438840)
  - We should use default paths for the config, the user can specify a config path if it's outside the default location

ra0x3:
CC @tjsharp1 @mitchmindtree @JoshuaBatty @SilentCicero for comments

adlerjohn:
Since the indexer can run without contract source code/Forc, _requiring_ configuration of the indexer to live in `Forc.toml` doesn't make sense. It can be added _optionally_ later.

Voxelot:
Would it also make sense to allow for dynamic configuration of the indexer to hot-reload wasm modules without downtime?

ra0x3:
> Would it also make sense to allow for dynamic configuration of the indexer to hot-reload wasm modules without downtime?
- +1 on being able to update the config and restart the process(es) with little-to-no downtime
- But I'm thinking, would that be the responsibility of the user? Meaning, outside the scope of the indexer?
  - **Example:** 
    - If I'm running the indexer as a `systemd` process, I would just update my config file (whether that be new WASM module paths, new database config, etc). From there I'd just `systemctl restart fuel-indexer` (or whatever)
- So I guess my question is, how much of that restarting/reloading/etc is the responsibility of the indexer versus the responsibility of the user and whatever service manager they're using? (systemd in this case)

Voxelot:
[...]

Issue #81, opened 2022-06-27
Use a single set of CLI opts


ra0x3:
- Related to https://github.com/FuelLabs/fuel-indexer/issues/47
- As of now, WASM execution and non-WASM execution use different CLI args
  - [WASM execution CLI args](https://github.com/FuelLabs/fuel-indexer/blob/master/indexer/src/bin/service.rs)
  - [non-WASM execution CLI args](https://github.com/FuelLabs/fuel-indexer/pull/65/files#diff-a4823faf4a276409ae568a0298b89751bddfaae8694fc1db61750890de0cc5d3R23)
- Both execution implementations should use the same set of CLI args

Additional notes:

- https://github.com/FuelLabs/fuel-indexer/pull/65 is already big enough as is. We should fix this in a separate PR

Issue #47, opened 2022-06-02
Allow service configuration via CLI args


ra0x3:
- We currently use a small manifest file to keep track of service config options
- Service config options should also be allowed via CLI args
- A bit of inspiration can be found [here](https://github.com/graphprotocol/indexer) and [here](https://github.com/graphprotocol/graph-node)

Issue #46, opened 2022-06-02
Docker support


ra0x3:
- We should add Docker support
- These Docker assets should be publicly available
- Support should constitute installing a the indexer service, adding dependencies, and service start commands

PR #116, opened 2022-07-13
Add Docker support

ra0x3:
### Description
- This PR:
  - Allows a user to start the service/node via a config file _or_ command line opts, or using nothing at all
  - Adds Docker support (the config stuff was a prerequisite)
  - Changes the name of the executor binary from `service` to `fuel-executor`

Thus addressing:
- https://github.com/FuelLabs/fuel-indexer/issues/47
- https://github.com/FuelLabs/fuel-indexer/issues/33
  - Respecting the discussion in https://github.com/FuelLabs/fuel-indexer/issues/89
- https://github.com/FuelLabs/fuel-indexer/issues/46
- https://github.com/FuelLabs/fuel-indexer/issues/81

### Notes
- I created `fuel-indexer-lib` as a utility all the other crates can pull from (for logic that is not unique to a single crate)
- If someone feels strongly we shouldn't have a config file, I will remove it

#### Executor
- [ ] Try to start the service a few different ways:
  - [ ] Just by running `cargo run --bin fuel-executor ` 
    - The service should use the specified defaults (using sensible defaults for everything)
  - [ ] Update the `config.yaml` with some non-default options and `cargo run --bin fuel-executor -- -c config.yaml`
    - Service should run using these updated options
    - Try adding secrets from the env like `$POSTGRES_PASSWORD` in the config.yaml and these should work as well
  - [ ] Run the service by just passing in CLI opts `cargo r --bin fuel-executor -- --fuel-node-host 127.0.0.1 --fuel-node-port 4004 --graphql-api-host 0.0.0.0 --graphql-api-port 23232 --postgres-user postgres --postgres-host localhost --postgres-port 5432`
    - Service should run using these CLI opts

### Docker
- [ ] Try building the image `docker build -t local/fuel-indexer:latest -f deployment/Dockerfile .`
  - Open an `ngrok` terminal for a public Fuel-node `ngrok http [fuel-node-port]`
  - Open an `ngrok` for a public Postgres instance `ngrok tcp 5432`
- [ ] Run the Docker image with `docker run $IMAGE --fuel-node-host $FUEL_NGROK_HOST --fuel-node-port $FUEL_NGROK_PORT --graphql-api-host 0.0.0.0 --graphql-api-port 23232 --postgres-user postgres --postgres-host $POSTGRES_NGROK_HOST --postgres-port $POSTGRES_NGROK_PORT`


ra0x3:
@tjsharp1 
- I ping'd @rfuelsh to review the Docker components (I realize that CI step is failing).
- Pinging you for a initial look at the rest of the changes

.env.example
```diff
@@ -0,0 +1,5 @@
+POSTGRES_USER=
+POSTGRES_DATABASE=
+POSTGRES_PASSWORD=
+POSTGRES_HOST=
+POSTRES_PORT=
```
rfuelsh, line 5: @ra0x3 - will values be placed here eventually?

.github/workflows/ci.yml
```diff
@@ -226,6 +226,71 @@ jobs:
...
+
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: |
+            ghcr.io/fuellabs/fuel-core
```
rfuelsh, line 252: @ra0x3 - you wanna change this to gchr.io repo for fuel-indexer if you intend to create the image e.g. ghcr.io/fuellabs/fuel-indexert

deployment/Dockerfile
```diff
@@ -0,0 +1,51 @@
+# Stage 1: Build
+FROM lukemathwalker/cargo-chef:latest-rust-1.61 as chef
+
+WORKDIR /build/
+
+# hadolint ignore=DL3008
+RUN apt-get update && \
```
rfuelsh, line 7: apt-get update -y

.env.example
ra0x3: @rfuelsh Yea the idea is that .env.example will show configuration (primarily for secrets) via a .env file

ra0x3:
@Voxelot  @tjsharp1  Just FYI this is officially ready for review

indexer-lib/src/lib.rs
```diff
@@ -0,0 +1,258 @@
...
+    };
+    use anyhow::Result;
+    use serde::Deserialize;
+    use std::fmt::Write;
+    use std::net::SocketAddr;
+    use std::path::PathBuf;
+    use structopt::StructOpt;
+
+    #[derive(Debug, StructOpt)]
+    #[structopt(
```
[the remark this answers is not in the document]
ra0x3: @Voxelot Interesting. I noticed this was different than what was used for `forc` but didn't think "it's not the same" was a good enough reason to change it. I can make this update

Voxelot:
@ra0x3 

For easier and automated testing, there's two strategies I'd suggest.

near-term goals
a.a) Use docker-compose to stand up an environment for localized testing. Docker-compose will handle the tricky port forwarding and networking details between services and is reproducible in CI or other developer machines. This will make it easy to setup a mock installation using all docker images for fuel-core, indexer, and postgres.
a.b) Setup some very basic "Acceptance Tests" that do the bare minimum to ensure the indexer is behaving correctly when connected with all the other live components (e.g. the main functional usecase like indexing events and making that data available via the API)

longer-term goals
b.a)  Have a k8s environment that all merges to master deploy the latest docker image to. Setup a ci flow to trigger your Acceptance Tests to run against every new deployment and provide feedback / slack alerts if something is amiss. 
b.c) AT + Ephemeral env: Once ephemeral environments are better supported, each PR could have its' own unique CD environment that also run the Acceptance Tests post-deploy.

ra0x3:
@Voxelot 
- docker-compose makes a lot of sense
- I've created https://github.com/FuelLabs/fuel-indexer/issues/126 to track this
- For the sake of not keeping this work hanging on the vine too long -- and since we're not using these images at the moment -- I would ask if we're good to merge this (if it looks OK) and then follow this up with #126  ?
  - Obviously I could also see the argument for "do these in one PR" but my thinking here is that let's get as many quick wins as possible while at the same time not putting the follow up work on the shelf for too long (given everything else that needs to be done)

Voxelot:
Does the api-server also need a separate docker image? Or does the fuel-executor binary include both?

ra0x3:
@Voxelot 
- The `fuel-api-server` can indeed be its own standalone binary, but yea `fuel-executor` includes a GraphQL endpoint as well
- I created https://github.com/FuelLabs/fuel-indexer/issues/128 which ideally I'd like to tackle after https://github.com/FuelLabs/fuel-indexer/pull/127

Voxelot (dismissed):

Voxelot:
There's a lot of changes here besides the dockerfile which I didn't have time to extensively review, but the dockerfile itself looks good.

deployment/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+COPY . .
+RUN cargo chef prepare --recipe-path recipe.json
+
+
+FROM chef as builder
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
+COPY --from=planner /build/recipe.json recipe.json
+RUN cargo chef cook --release -p fuel-indexer --recipe-path recipe.json
+COPY . .
+RUN cargo build --release -p fuel-executor --all-targets
```
Voxelot, line 32:
```suggestion
RUN cargo build --release -p fuel-indexer --all-targets
```

Issue #121, opened 2022-07-19
Archive assets on release


ra0x3:
- [Similar to how it's done in fuel-core](https://github.com/FuelLabs/fuel-core/blob/master/.github/workflows/ci.yml#L223)
- This would include the `indexer` and `api_server` binaries
- [Would use the same architecture matrix as well](https://github.com/FuelLabs/fuel-core/blob/master/.github/workflows/ci.yml#L232)

PR #125, opened 2022-07-21
Adjustments to #121

ra0x3:
Testing asset building here https://github.com/FuelLabs/ci-dummy-test/commits/master


ra0x3:
@Voxelot 
- So to get the archiving assets working, I started by just copying over what we had from fuel-core. That PR is here https://github.com/FuelLabs/fuel-indexer/issues/121
- However, when testing the asset creation and archiving, I noticed some things weren't functioning as expected (lots of linker errors, fuel-executor expecting different dependency versions than were included in the base Dockerfiles, etc)
  - Did a lot of experimentation over at https://github.com/FuelLabs/ci-dummy-test/commits/master to see what works and what didn't work
  - For example, some dependencies of fuel-indexer's Postgres require glibc 2.18 (but the glibc version in `FROM ghcr.io/cross-rs/x86_64-unknown-linux-gnu:main-centos` is 2.17)
- Eventually managed to get a base `ubuntu` image to build (which is a mirror of the Dockerfile from  https://github.com/FuelLabs/fuel-indexer/pull/116)
- So I created this PR, with its 2 assets, as a PoC of what's currently working for me, with the idea being we discuss what needs to stay/go

CC @tjsharp1

PR #127, opened 2022-07-21
Standardize crate naming

ra0x3:
### Description
- This PR standardizes the fuel-indexer crate names
- Everything should work as normal
- Renamings 
  - `indexer` -> `fuel-executor`
  - `lib` -> `fuel-indexer`
  - `api_server` -> `fuel-api-server`
  - `schema` -> `fuel-indexer-schema`
  - `derive` -> `fuel-indexer-derive`
  - `indexer-lib` -> `fuel-indexer-lib`

### Notes
- Waiting on approval before making a PR to reserve  the`fuel-indexer-lib` crate


JoshuaBatty (approved):

mitchmindtree:
Nice, for the most part I think these seem like a big improvement!

> lib -> fuel-indexer
> indexer-lib -> fuel-indexer-lib

Could you elaborate a little on how the role of each of these crates differs?

Also, it might be overly verbose, but I wonder if `fuel-executor` and `fuel-api-server` should be `fuel-indexer-executor` and `fuel-indexer-api-server`, as they're quite specific to the indexer?

ra0x3:
@mitchmindtree 
- `fuel-indexer` is just a container for some higher level FFI logic
  - This crate name is a slight misnomer, as `fuel-indexer` is really the entirety of the workspace (rather than a single crate), but this doesn't present any issues at the moment
- `fuel-indexer-lib` is a bunch of utility components that are used by, but don't belong to any particular other crate
- You're right that these crates are specific to the indexer, so at first I considered naming `fuel-executor` `fuel-indexer-executor`, but after looking over [the fuel-core repo](https://github.com/FuelLabs/fuel-core) I saw that there are several crates named `fuel-*` and not `fuel-core-*` so decided to just go with `fuel-executor`

mitchmindtree:
Thanks @ra0x3! This helps to clear things up in my mind a bit.

> fuel-indexer is just a container for some higher level FFI logic 

It looks like this crate is purely used via code that is generated by `fuel-indexer-derive` within fuel indexer plugin implementations - does this sound right?

> fuel-indexer-lib is a bunch of utility components that are used by, but don't belong to any particular other crate

What are your thoughts on using something like `fuel-indexer-util` for these kinds of miscellaneous, shared utilities? We have something very similar in the sway repo as `forc-util` that we share between the different forc crates.

> You're right that these crates are specific to the indexer, so at first I considered naming fuel-executor fuel-indexer-executor, but after looking over [the fuel-core repo](https://github.com/FuelLabs/fuel-core) I saw that there are several crates named fuel-* and not fuel-core-* so decided to just go with fuel-executor

I think I'm particularly concerned about the `fuel-executor` and `fuel-api-server` names as they're quite ambiguous on their own. E.g. `fuel-executor` almost sounds more like a fuel node runner or block executor, when in reality it executes indexer plugins to do the indexing.

I think the crates in the `fuel-core` repo might be a bit of a special case seeing as they are all related directly to the `fuel` node itself which is very much the foundation of the fuel ecosystem. In our case, all our crates are very specific to the indexer.

---

After taking all this into consideration, what are your thoughts on the following names?

| Old Name | New Name | Description |
| --- | --- | --- |
| `indexer` | `fuel-indexer` | The core binary that performs indexing by calling into plugins. |
| `indexer-lib` | `fuel-indexer-util` | Common utilities shared between the indexer and API server crates. |
| `schema` | `fuel-indexer-schema` | DB schema definition, shared between indexer, api-server and plugins. |
| `api_server` | `fuel-indexer-api-server` | Allows for running queries on an indexer DB. |
| `lib` | `fuel-indexer-plugin` | The top-level lib required for implementing a custom indexer plugin. |
| `derive` | `fuel-indexer-plugin-macros` | Implementations of proc macros useful to plugin implementations. |


### `fuel-indexer-plugin` and `fuel-indexer-plugin-macros`

My thinking with these names is to clarify that these are specific to custom plugin implementations.

A follow-up step that we can take so that plugin writers do not have to import `fuel-indexer-plugin-macros` is to re-export its macros via the `fuel-indexer-plugin` crate. You can see an example of how to do this in the `fuels` crate where [they re-export the `abigen` macro from `fuels`](https://github.com/FuelLabs/fuels-rs/blob/f8aea8e0dead36ec97cfe410c616c7252bb9c92b/packages/fuels/src/lib.rs#L62). This would mean the user only needs to add the `fuel-indexer-plugin` dependency when writing a custom fuel indexer plugin. We can track this in a separate issue, but I just wanted to mention this as it's relevant to the name selection.

Also, I noticed the `derive` crate doesn't appear to implement any derive macros, but instead implements some regular proc macros (`handler`, `graphql_schema`). That's what lead me to think of renaming the `derive` suffix to `macros`.

@ra0x3 curious to get your thoughts!

ra0x3:
@mitchmindtree  
- Thanks for taking the time to write that.
- Just to be clear, I have no strong feelings on what any of these names should be (chime in if you have some opinions @tjsharp1 @adlerjohn )
- I'd be ok with changing `fuel-executor` to `fuel-indexer`, and `fuel-indexer` to `fuel-indexer-plugins` I reckon. Other than that I feel like we're in personal preference territory 😅 so unless there's a super strong case or someone else says the names should change, I'd suggest leaving as-is (meaning whatever is on the base branch)

Going by [Nick's comment](https://fuellabs.slack.com/archives/C02RU9MNVKN/p1658846633381449?thread_ts=1658812178.534629&cid=C02RU9MNVKN) it seems `fuel-*` is an OK prefix, as opposed to `fuel-[project]-*` just FYI

PR #143, opened 2022-08-06
Add e2e docker compose test

ra0x3:
### Description
- This is an end-to-end test of the indexer service using a reproducible environment. 
- This PR does a few things
  - Simplifies the `simple-native` example (just removing code)
    - Also updates the `simple-native` `fuels` crates to `0.15` (that's the version we use everywhere else, not sure why it was `0.16`)
  - Adds a `composable-indexer` crate in `tests/e2e` that contains:
    - A fuel node
    - A simple web api (this is a  stand in for any web app -- e.g., SwaySwap)
    - An indexer project `composable-indexer-lib` with a WASM module
      - This is what users would build
  - Updates the `deployment/Dockerfile` with `rustc 1.62` as well as a `tzinfo` package
  - Manually reverts `anyhow` from `1.0.60` to `1.0.53` (doesn't build on stable)
  - Updates CI actions to use `toolchain: 1.62` from `toolchain: stable` (stable now 1.63)

### Testing steps
- [ ] Pull down the PR
- [ ] Build images and docker-compose up `bash tests/e2e/composable-indexer/compose-up.bash` (go grab some ☕)
- [ ] Simulate an event `curl -X POST http://0.0.0.0:8000/ping`
- [ ] Verify the event was indexed:

```bash
curl -X POST http://0.0.0.0:29987/graph/composability_test \
   -H 'content-type: application/json' \
   -d '{"query": "query { message { id ping pong message }}", "params": "b"}' \
   | json_pp
```

Should output

```json
[
   {
      "id" : 1,
      "message" : "6161616173646673646661736466736466616173646673646661736466736466",
      "ping" : 123,
      "pong" : 456
   }
]
```


ra0x3:
@tjsharp1 Got some questions for you on this one.
- I managed to get this working, but if you look at [my handler function](https://github.com/FuelLabs/fuel-indexer/pull/143/files#diff-68ebc2b8ace5c49a51241a885de1e93e3c51dc5f1b4bd934a75fc2c13e2a27ebR14) I'm just creating new data in the function -- not passing the data from the Fuel client (due to some issues I was having) 👎 
- Some issues I'm having:

> Issue 1: If I add a `str[]` type to my smart contract, I get those WASM issues I was discussing with you earlier
  - I went back and double checked that my ABI-JSON and my contract were correct
  - For whatever reason, WASM panicked on my `str[4]` field (I've since changed that field)
  - As you mentioned [this line](https://github.com/FuelLabs/fuel-indexer/blob/d09cd342362577c73b441e9297952472d117cab0/fuel-indexer-macros/src/handler.rs#L31) could be the culprit but I didn't debug that far

> Issue 2. I get a Postgres error when I try to save the `id` field from my `ping` event as is. 
  - After inspecting the `INSERT` statement, for whatever reason the `id` field that the `function_one` handler gets is a _really_ big integer.
  - It seems that [this ID set in the smart contract](https://github.com/FuelLabs/fuel-indexer/pull/143/files#diff-7e5da5219180357b52554785c235e951bbfbb02155a42b81fac656551611c5cdR23) is being read as the integer `10428267242562419082` which is obviously too big
    - Could this be an endianess issue? 

If you pull down this PR I wrote some pretty thorough reproduction steps [in this README](https://github.com/FuelLabs/fuel-indexer/pull/143/files#diff-db77e085991f0288be91da0d0bc86db14c186d983be45c5912068a19232c0f8eR1) that you can follow. Curious to see if these issues affect you as well. 

I planned to add the dockercompose element after I get these issues sorted out -- so you don't have to run each service individually

PR #176, opened 2022-08-28
Adding remote dispatch for fuel indexer deployment

rfuelsh:
This will automatically deploy fuel indexer to the EKS when a master build is ongoing


ra0x3 (approved):
utAck :shipit:

PR #178, opened 2022-08-29
Send block data to indexer

tjsharp1:
Still have some tests to fix up, but opening the draft now, in case people want to start looking at it.

### DIFF
--- a/fuel-indexer/src/executor.rs
+++ b/fuel-indexer/src/executor.rs
@@ -54,11 +54,7 @@ pub struct NativeIndexExecutor {
 }
 
 impl NativeIndexExecutor {
-    pub async fn new(
-        db_conn: &str,
-        manifest: Manifest,
-        path: String,
-    ) -> IndexerResult<Self> {
+    pub async fn new(db_conn: &str, manifest: Manifest, path: String) -> IndexerResult<Self> {
         let db = Arc::new(Mutex::new(
             Database::new(db_conn)
                 .await


fuel-indexer/src/manifest.rs
```diff
@@ -8,24 +7,29 @@ use std::{fs::File, io::Read};
 pub struct Manifest {
     pub namespace: String,
     pub graphql_schema: String,
-    pub wasm_module: Option<String>,
+    pub module: Module,
     pub start_block: Option<u64>,
-    pub handlers: Vec<Handle>,
     pub test_events: Option<Vec<Event>>,
+    pub indexer_port: Option<u16>,
```
ra0x3, line 13: Can we not use `IndexerConfig` for this?

fuel-indexer/src/service.rs
```diff
@@ -204,103 +186,81 @@ impl IndexerService {
...
-                    }
-                })
-                .await;
+                let result = executor.handle_events(block_info).await;
 
                 if let Err(e) = result {
-                    error!("Indexer executor failed {e:?}");
+                    error!("Indexer executor failed {e:?}, retrying.");
+                    sleep(Duration::from_secs(5)).await;
+                    retry_count += 1;
```
ra0x3, line 249: https://github.com/FuelLabs/fuel-indexer/issues/194

ra0x3 (changes requested):
@tjsharp1 

- Thanks for this 🙏 
- Obviously it's a really big/impactful API change so left a few more comments than usual
  - Mainly some questions 
- I pulled this down to test, and managed to get the `simple-native` example working (but had to change it to use WASM cause native wasn't working -- which is fine)
- Still trying to get the E2E test working

tests/e2e/composable-indexer/composable-indexer-lib/src/lib.rs
```diff
@@ -1,29 +1,26 @@
...
-    no_name,
-    "tests/e2e/composable-indexer/composable-indexer-lib/contracts/ping/out/debug/ping-abi.json"
-);
+#[indexer(
+    abi = "tests/e2e/composable-indexer/composable-indexer-lib/contracts/ping/out/debug/ping-abi.json",
+    namespace = "composability_test",
+    schema = "schema/schema.graphql"
+)]
+mod composability_test {
+    fn function_one(ping: Ping) {
```
ra0x3, line 10: Last thing (forgot to leave this in the original review).
- So how is block/transaction indexing going to work here?
  - Would we write special handlers for `Block`s and/or `Transaction`s ?
  - For example, what if I wanted access to a `Block` here for indexing (e.g., SwaySwap)
    - Or would this behavior lie somewhere else entirely? 🤔

deekerno (commented):
Mostly the same concerns as @ra0x3 regarding leftover files and potentially leftover comments, but it looks good overall so far.

fuel-indexer/src/service.rs
ra0x3: Yea I didn't mean do this in this PR. Down the road

tests/e2e/composable-indexer/composable-indexer-lib/src/lib.rs
ra0x3: Yea if that's too much for this PR we can always follow up. But will need that before we can start indexing blocks I believe

fuel-indexer-macros/test_data/fail_args.stderr
```diff
@@ -1,7 +1,13 @@
-error: handler macro does not take arguments
- --> test_data/fail_args.rs:9:1
+error: abi specification is required in indexer definition!
```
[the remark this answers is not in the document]
ra0x3: @tjsharp1 
- Are they needed for the service to run? They look like dev files (files you keep locally while you're working on things).
- If at all possible, maybe we can `.gitignore` them, remove them from the repo, that way everyone just has them locally and not in the repo?

fuel-indexer/src/executor.rs
```diff
@@ -81,35 +82,43 @@ impl Executor for NativeIndexExecutor {
...
+        let msg = serialize(&IndexerMessage::Blocks(blocks));
+
+        self.stream.write_u64(msg.len() as u64).await?;
+        self.stream.write_all(&msg).await?;
+
+        loop {
+            let size = self.stream.read_u64().await? as usize;
+
+            if self.stream.read_exact(&mut buf[..size]).await? < size {
+                return Err(IndexerError::HandlerError);
```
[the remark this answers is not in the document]
ra0x3: You are correct. My mistake here.

ra0x3:
@tjsharp1 
- Sorry for the conflicts :(
- I have a local branch with a few minor changes to this work (updates to an example, and to the e2e test) that I will probably push up soon
- But locally (with those mentioned changes) this PR seem to work for me 👌 
- Approved :shipit: 
- Can give the official ✅ when the conflicts are fixed

PR #190, opened 2022-09-02
changed unwrap

Jade-Jones-dev:
changed unwrap to expect


deekerno:
Thanks for the contribution! We happened to address this in a separate PR, but we appreciate you taking the time to have a go at this. Feel free to look at any of the other issues labeled "Good first issue", and if you choose one, let us know by adding a comment so we don't duplicate work. :smile:

Also, here's [a link](https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#shortcuts-for-panic-on-error-unwrap-and-expect) about the differences between `unwrap` and `expect`, and why one may choose one over the other.

Closing, superseded by #191.

PR #201, opened 2022-09-08
Update test & example for new api + fix Docker build

ra0x3:
### Description 
- This PR does a few things:
  - Fixes an unexpected bug in the Docker image (due to some recent changes in Ubuntu 22.04. I believe [This SO question](https://askubuntu.com/questions/1424442/libssl1-1-is-deprecated-in-ubuntu-22-04-what-to-do-now) is in the ballpark of what I'm referring to)
  - Updates the `simple-native` example
    - Native execution was giving grief - but we're focusing on WASM only anyway for now so I just renamed the example and switched it to WASM 
      - Also updated to mdbook
      - Updated some folder names for clarity
  - Updates the e2e test

### Testing steps

Docker e2e test should work

- [ ] The simple end-to-end testing steps [here](https://github.com/FuelLabs/fuel-indexer/blob/master/tests/e2e/composable-indexer/README.md) should pass


docs/src/examples/counter.md
```diff
@@ -0,0 +1,117 @@
+# Counter
+
+> In this example project we'll walk through how to run a Fuel Indexer service that uses WASM-based execution. We'll go through project structure, how to actually use the service, as well as a few key concepts.
```
deekerno, line 3: This line states that the project uses WASM-based execution, but the last line states that the project used native Rust execution.

deekerno (changes requested):
Nice and clear overall; left a nit on `examples/counter.md`.

docs/src/examples/counter.md
ra0x3: Ahh good catch @deekerno. Pushing a change

deekerno (approved):
🚢

ra0x3:
Went ahead and merged this cause the Docker changes are needed to test some helm stuff

Issue #185, opened 2022-09-02
Optimize healthcheck's uptime implementation


ra0x3:
- Right now the Indexer uses `ps` via a `Command` to grab the service's uptime
- As noted in [this comment](https://github.com/FuelLabs/fuel-indexer/pull/174#discussion_r961047876) we can optimize this process a bit by just passing some `$CURRENT_TIME_IN_MILLISECONDS` argument to the service at start time
- We would then just calculate uptime using this starting value instead of running a process every 10secs

Additional context
- The healthcheck runs every 10 seconds so this isn't a huge issue for now

PR #202, opened 2022-09-08
Calculate uptime by using time elapsed since instant

deekerno:
## Description
As detailed in #185, this PR adjusts the calculation of API uptime to use elapsed time.

### DIFF
--- a/fuel-indexer/src/api.rs
+++ b/fuel-indexer/src/api.rs
@@ -20,7 +20,7 @@ use hyper::Client;
 use hyper_tls::HttpsConnector;
 use serde::Deserialize;
 use serde_json::{json, Value};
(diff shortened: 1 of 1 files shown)


ra0x3 (changes requested):
@deekerno Looks good. My only comment is just to update the `"uptime"` key in the JSON response to clarify what interval the time is in (e.g., `"uptime(seconds)"` or `"uptime(millis)"` or whatever the time granularity is)

deekerno:
@ra0x3 Good point. Added the interval, and changed to seconds instead of milliseconds as the former is probably more useful for uptime reasons.

ra0x3 (approved):
LGTM 👍

PR #212, opened 2022-09-14
Hotfix/helm

uberscott:
fixed issue where helm chart was not supporting multiple upgrades.   (not tested in github actions until we merge)


ra0x3 (approved):
With https://github.com/FuelLabs/fuel-indexer/pull/213 passing env vars in the start command should work now

```bash
POSTGRES_PASSWORD=foo ./fuel-indexer --other-arg
```

.gitignore
```diff
@@ -8,3 +8,5 @@ docs/book/
 *.db
 *.db*
 
+
+.idea
```
Voxelot, line 12: typically @adlerjohn has wanted us to avoid ide specific settings in the shared .gitignore, and for ppl to update their local `.git/info/exclude` instead.

PR #216, opened 2022-09-14
Allow indexing blocks & transactions

ra0x3:
### Description
- This PR does a few things:
  - Adds the ability to index `BlockData` 
    - By using a `#[block]` attribute, a user can have access to `BlockData` in the function signature
      - This is needed by SwaSwap
  - Adds a `explorer.graphl` schema (owned by Fuel) that describes the layout of the eventual entities
  - Adds a `Jsonb` schema type that we can use to store arbitrary data (needed by the explorer)
  - Adds a few more properties to `BlockData`

### Notes

Here's the example usage

```graphql
# project/schema/block.graphql
schema {
    query: QueryRoot
}

type QueryRoot {
    block: Block
    transaction: Transaction
    count: Count
}

type Block {
    id: ID!
    height: UInt8!
    producer: Address!
    hash: Bytes32!
    timestamp: Int8!
}


type Transaction {
    id: ID!
    hash: Bytes32!
    block: Block!
    input_data: Jsonb!
    timestamp: Int8!
}


type Count {
    id: ID!
    count: UInt8!
    timestamp: UInt8!
}
```

```rust
// project/block-indexer/src/lib.rs
extern crate alloc;
use fuel_indexer_macros::indexer;

#[indexer(
    abi = "contract-abi/block.json",
    namespace = "block_explorer",
    identifier = "block_index",
    schema = "schema/block.graphql"
)]
mod block_indexer {

    fn counter_thing(event: CountEvent) {
        let c = Count{ id: 1, count: event.count, timestamp: 1 };
        c.save();
    }

    #[block]
    fn block_handler(block: BlockData, count: CountEvent) {
        let blk = Block {
            id: block.height,
            hash: block.id,
            height: block.height,
            producer: block.producer,
            timestamp: block.time,
        };
        blk.save();

        let input_data = r#"{"foo":"bar"}"#.to_string();

        for (i, _receipts) in block.transactions.iter().enumerate() {
            let tx = Transaction {
                id: i as u64,
                hash: [0u8; 32].into(),
                block: blk.id,
                timestamp: block.time,
                input_data: Jsonb(input_data.clone()),
            };
            tx.save();
        }
    }
}
```

### Testing steps
- [ ]  Copy and past the above ☝️ into a project, trigger an event, etc (the usual)


fuel-indexer-macros/src/schema.rs
```diff
@@ -56,6 +56,32 @@ fn process_field<'a>(
     (typ, ident, extractor)
 }
 
+fn process_fk_field<'a>(
```
ra0x3, line 59: Found a bug while I was in there related to FKs

fuel-indexer-macros/src/indexer.rs
```diff
@@ -101,6 +102,41 @@ fn decode_snippet(ty_id: u64, ty: &Ident, name: &Ident) -> proc_macro2::TokenStr
     }
 }
 
+fn process_block_fn_items(input: &ItemFn) -> TokenStream2 {
```
ra0x3, line 105: I could've hacked this into `process_fn_items`, but wanted to be explicit and keep this separate

deekerno (dismissed):

fuel-indexer/src/executor.rs
```diff
@@ -228,166 +228,169 @@ impl Executor for WasmIndexExecutor {
...
-                            is: 0,
-                        },
-                    ],
-                ],
-            }])
-            .await;
-        assert!(result.is_ok());
-    }
-}
+// #[cfg(test)]
```
ra0x3, line 231: @tjsharp1 
- Obviously would prefer to leave these tests in as they're valuable (and we already don't have that much testing)
- But absent a good resolution, can't let them block this work for too much longer. 
- I can create a high priority tracking issue to make sure we don't forget about this, but yea -- I think I've reached the end of my debugging ideas with this 🤷

deekerno (approved):
LGTM. 👌🏽

PR #220, opened 2022-09-16
Change HTTP endpoints to gracefully fail 

### LINKED ISSUE #215: Allow all HTTP endpoints to fail gracefully
- We're currently using a bit of panic/expect logic on web API routes
- The server should not stop if anything goes wrong with an HTTP request
- We should allow all requests to fail gracefully, log errors, and return a generic `INTERNAL_SERVER_ERROR`

Additional context
- Looks like axum has some type of error handling functionality but not sure how easy it is to use https://docs.rs/axum/latest/axum/error_handling/index.html

deekerno:
## Description
This changes the behavior of the HTTP endpoints from panicking to returning a 500 status code with a message. This will close #215.

I wrestled with `axum` and its error handling functionality for awhile, but couldn't get the types to fully work. I may try it again if feedback points towards going that way.

### DIFF
--- a/fuel-indexer/src/api.rs
+++ b/fuel-indexer/src/api.rs
@@ -82,21 +82,50 @@ pub async fn health_check(
     // Get fuel-core status
     let https = HttpsConnector::new();
     let client = Client::builder().build::<_, hyper::Body>(https);
-    let resp = client
-        .get(
-            format!("{}/health", config.fuel_node.http_url())
-                .parse()
-                .expect("Failed to parse string into URI"),
-        )
-        .await
-        .expect("Failed to get fuel-client status.");
 
-    let body_bytes = hyper::body::to_bytes(resp.into_body())
-        .await
-        .expect("Failed to parse response body.");
+    let uri = match format!("{}/health", config.fuel_node.http_url()).parse() {
+        Ok(s) => s,
+        Err(e) => {
+            error!("Failed to parse string into URI: {}", e);
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(Value::String("Failed to parse string into URI".into())),
+            );
+        }
+    };
 
-    let fuel_node_health: FuelNodeHealthResponse =
-        serde_json::from_slice(&body_bytes).expect("Failed to parse response.");
+    let resp = match client.get(uri).await {
+        Ok(response) => response,
+        Err(e) => {
+            error!("Failed to get fuel-client status: {}", e);
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(Value::String("Failed to get fuel-client status".into())),
+            );
+        }
+    };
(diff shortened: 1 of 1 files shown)


ra0x3:
Hey @deekerno thanks for jumping into this.

- From what I can see, looks like you're `expect`'ing various `panic`'s in the route handlers, then returning error responses from there
- This seems to work OK for now, but ideally we'd use an error handler that would act as a catch-all for _all_ errors. The benefits of this are:
  1. We don't have to `expect` errors in the route handlers. As long as our top-level error (e.g., `APIError`) implements the errors we `panic` on (e.g., `SqlxError`) the catch-all handler will be triggered thus making the routes infallible 
  2. We don't have to write lots of line of codes on a all of these `expect`'s -- we can just write this logic in a single place
- I took a few mins to put up a PoC of what I'm referring to here https://github.com/FuelLabs/fuel-indexer/pull/221
- Happy to chat if you have any questions

ra0x3 (changes requested):
Hey @deekerno thanks for jumping into this.

- From what I can see, looks like you're `expect`'ing various `panic`'s in the route handlers, then returning error responses from there
- This seems to work OK for now, but ideally we'd use an error handler that would act as a catch-all for _all_ errors. The benefits of this are:
  1. We don't have to `expect` errors in the route handlers. As long as our top-level error (e.g., `APIError`) implements the errors we `panic` on (e.g., `SqlxError`) the catch-all handler will be triggered thus making the routes infallible 
  2. We don't have to write lots of line of codes on a all of these `expect`'s -- we can just write this logic in a single place
- I took a few mins to put up a PoC of what I'm referring to here https://github.com/FuelLabs/fuel-indexer/pull/221
- Happy to chat if you have any questions

deekerno:
> Hey @deekerno thanks for jumping into this.
> 
>     * From what I can see, looks like you're `expect`'ing various `panic`'s in the route handlers, then returning error responses from there
> 
>     * This seems to work OK for now, but ideally we'd use an error handler that would act as a catch-all for _all_ errors. The benefits of this are:
>       
>       1. We don't have to `expect` errors in the route handlers. As long as our top-level error (e.g., `APIError`) implements the errors we `panic` on (e.g., `SqlxError`) the catch-all handler will be triggered thus making the routes infallible
>       2. We don't have to write lots of line of codes on a all of these `expect`'s -- we can just write this logic in a single place
> 
>     * I took a few mins to put up a PoC of what I'm referring to here #221
> 
>     * Happy to chat if you have any questions

Yeah, I'll hit you up out-of-band for a chat 👌🏽 I did a short attempt with the catch-all handler already, but it seemed a bit intractable, so I'm curious to see if we have similar thoughts.

ra0x3:
@deekerno and I synced. Closing in favor of https://github.com/FuelLabs/fuel-indexer/pull/222

PR #223, opened 2022-09-19
Add database option explanation to Database page

### LINKED ISSUE #160: Add explanation for `DbType`s
- When https://github.com/FuelLabs/fuel-indexer/pull/134 lands, we'll have support for multiple database types
- It would be nice to clarify some of the differences between the two (obviously the end-user won't really see the difference)
- This should go under section 4 "Database"

deekerno:
## Description
This PR adds a neutral comparison of the capabilities of both Postgres and SQLite to the Database section of the book in order to aid the reader in choosing the best solution for their use case. This will close #160.


deekerno:
Closing; will open a new one with more of a focus on implementation nuances rather than conceptual differences.

PR #249, opened 2022-10-06
Add CI job for wasm example

### LINKED ISSUE #239: Add CI check for WASM builds
- Sometimes when making dependency changes to Indexer crates, we unexpectedly break WASM compilation
- However, this might get overlooked since WASM builds are not a part of CI
- We should add WASM compilation to CI to prevent this

Additional context
- Could be something as simple as a new `step` with a `cargo build` in an indexer example project

deekerno:
## Description
Adds CI job for to ensure proper compilation of WASM example. Closes #239.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,36 @@ jobs:
         env:
           RUSTFLAGS: '-D warnings'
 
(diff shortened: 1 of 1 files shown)


ra0x3 (approved):
Perfect 🚢

PR #256, opened 2022-10-10
(WIP) Allow for asset registration without restart

### LINKED ISSUE #237: Allow uploaded assets to be registered in an executor without service restart
- After https://github.com/FuelLabs/fuel-indexer/pull/203 we have the ability to upload assets via an HTTP endpoint
- However, this still requires the indexer service to be restarted once the assets are uploaded, so that new executors can be created from the uploaded assets
- We should implement logic that allows for new executors to be created and registered with the indexer service, as soon as assets are successfully uploaded -- without requiring a service restart

Additional context
- Might be tricky since the indexer service keeps state of executors when the service starts, and this (by definition) would require mutating that state at runtime with some communication between the GraphQL component and the service component

deekerno:
## Description
This PR adds the ability for assets to be registered and executed without a restart. Closes #237.

_Note: Currently a work in progress._

### DIFF
--- a/fuel-indexer/Cargo.toml
+++ b/fuel-indexer/Cargo.toml
@@ -34,7 +34,7 @@ serde_json = { version = "1.0", features = ["raw_value"] }
 serde_yaml = "0.8"
 sqlx = "0.6"
 thiserror = "1.0"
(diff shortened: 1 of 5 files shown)


deekerno:
Spoke to @ra0x3; closing this in favor #258. 👌🏽

PR #257, opened 2022-10-10
Move GraphQL API component to fuel-indexer-api-server

### LINKED ISSUE #175: Move `GraphQLApi` component to `fuel-indexer-api-server`
- The `GraphQLApi` component is tucked in the `fuel-indexer` crate
- The `GraphQLApi` component depends `fuel-indexer-database-types`, `fuel-indexer-lib`, and `fuel-indexer-schema`
- We should move the `GraphQLApi` component to `fuel-indexer-api-server`, which already depends on `fuel-indexer-schema`, `fuel-indexer-lib`, and `fuel-indexer`
  - At most, `fuel-indexer-api-server` would have 1 more dependency (i.e., `fuel-indexer-database-types`)
  - This would in turn cause `fuel-indexer` to depend on `fuel-indexer-api-server` (reverses the dependency)
  - Would make `fuel-indexer-api-server` more of an independent crate

Additional context
- This might require a bit more thought

deekerno:
## Description
This PR moves the GraphQL API component to the `fuel-indexer-api-server` crate. Closes #175.

Given comments on #175, we may want to look into decomposing `IndexerError` as it's taken on a bit of a "god object" status; however, I did not do that here in order to keep this PR short and focused.


ra0x3 (approved):
@deekerno 
- Looks good, approved ✅ . I retriggered the udeps job one more time in hopes that the issue has been fixed. 
- If not, we can either use the same version of nightly as we're using on stable (1.64). 
  - I see this issue seems to be stemming from 1.66 nightly

ra0x3 (approved):
🚢

PR #273, opened 2022-10-18
Swap manual trait impl for strum

ra0x3:
### Description
- This PR does a bit of cleanup and swaps manual `ToString` implementations (and the like) for `strum::EnumString` usage

### Testing steps
- N/A


deekerno (approved):
Very nice. 👏🏽

Issue #230, opened 2022-09-20
Allow indexing `Receipt::TransferOut`


ra0x3:
- https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#transferout-receipt

PR #274, opened 2022-10-19
Add more receipt types

### LINKED ISSUE #228: Allow indexing `Receipt::ScriptResult`
- https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#scriptresult-receipt

deekerno:
Closes #228 and #230. 
## Changelog
- Adds support for `ScriptResult` receipt
- Adds support for `TransferOut` receipt

I may include `MessageOut` in this PR, but I'm still doing research on that; I thought it was best to not block the other two.

## Testing Plan
Added new e2e tests; however, like the `Transfer` receipt, the `TransferOut` e2e test needs some work.


fuel-indexer-macros/src/indexer.rs
```diff
@@ -448,6 +487,16 @@ fn process_fn_items(
                                 decoder.decode_logdata(rb, data);
 
                             }
+                            Receipt::ScriptResult { result, gas_used } => {
+                                #contract_conditional
+                                let data = ScriptResult{ result: u64::from(result), gas_used };
```
ra0x3, line 492: I believe you can remove the `u64::from`, as the decimal string will be automatically coerced to `u64`

fuel-indexer-tests/tests/e2e.rs
```diff
@@ -145,3 +145,79 @@ async fn test_can_trigger_and_index_log_event() {
...
 //     assert_eq!(1, 1);
 // }
+
+// TODO: Need to find a better way to test ScriptResult receipt indexing as
+// the expected values for this test will be overwritten by any test
+// with an event that involves a ScriptResult receipt, which may
+// make the assertions fail as tests are not called in a particular order;
+// one can verify that indexing works by running this test by itself
+
+// #[tokio::test]
```
ra0x3, line 155: @deekerno 
- I see the `TODO` but we should still have a test for `ScriptResult` (since that's actually the easiest to generate)
- Not a blocker to the PR getting merged -- but as a very last resort, to test this in isolation, we could just remove all other `ScriptResult`s in this test before calling `client.post()`
  - We could even just have this test `sqlx::query` a previous test's `ScriptResult`

ra0x3 (changes requested):
Looks mostly fine, the `u64::from` comment is the only real requested change

ra0x3 (approved):
- Dismissing my last request for changes
- the fuel-specs/abi docs are a little out of date -- @deekerno and I discussed offline

PR #276, opened 2022-10-20
Add tests for transfer receipt type

deekerno:
## Changelog
- Adds e2e test for `Transfer` receipt type

## Testing Plan
Added new e2e tests; CI should pass.


fuel-indexer-tests/src/lib.rs
```diff
@@ -14,6 +14,8 @@ pub mod defaults {
     pub const WEB_API_ADDR: &str = "127.0.0.1:8000";
     pub const PING_CONTRACT_ID: &str =
         "68518c3ba3768c863e0d945aa18249f9516d3aa1338083ba79467aa393de109c";
+    pub const TRANSFER_BASE_ASSET_ID: &str =
```
ra0x3, line 17: - This is already exported by `fuels::prelude::BASE_ASSET_ID`.
- Can probably just import that directly instead of vial `fuel_indexer_tests::defaults`

fuel-indexer-tests/tests/e2e.rs
```diff
@@ -94,8 +94,19 @@ async fn test_can_trigger_and_index_transfer_event() {
...
     sleep(Duration::from_secs(defaults::INDEXED_EVENT_WAIT)).await;
 
-    // FIXME: Still need to trigger an actual receipt
-    assert_eq!(1, 1);
+    let row = sqlx::query("SELECT * FROM fuel_indexer_test.transferentity where id = 1")
+        .fetch_one(&mut conn)
+        .await
+        .unwrap();
+
+    let _id: i64 = row.get(0);
```
ra0x3, line 102: (nit) If not using these vars with underscore (e.g., `_id`) can remove them

ra0x3 (approved):
- Works as expected so ✅ 
- Left a piece of feedback and a nit, not blockers though (if you wanna follow up with another PR)

fuel-indexer-tests/src/lib.rs
deekerno: Oops, that was left over; I'll remove it in another PR.

PR #297, opened 2022-10-30
Update explorer example

ra0x3:
### Description

- This PR updates the block-explorer example to be slightly better understood within the context of how dApp developers can leverage not only blocks/transactions, but also receipts.

### Testing steps
- [ ] Same as https://github.com/FuelLabs/fuel-indexer/pull/279


deekerno (approved):
🚢

PR #306, opened 2022-10-31
Add new steps to release

ra0x3:
Just FYI there's two ways to debug this:
1. Just keep merging into master until it works (might have to do this 1 or 2 more times)
2. Pull every thing over the FuelLabs/ci-dummy-test and test there, then pull it back over the fuel-indexer when it works

Don't feel like putting in the effort for number 2 🙄  so going with number 1


deekerno (approved):
> "Laziness is the first step towards efficiency."

PR #319, opened 2022-11-03
v0.1.0-isolate-crates

ra0x3:
### Description
- Yet another v0.1.0 release error
- Turns out this was related to an issue where although the crates build OK and pass CI -- when built in isolation they would fail (thus the release error)
- This PR used [`cargo-make`](https://github.com/sagiegurari/cargo-make) to individually build each crate
  - As you can see, a few manifest files had to be updated
- Also introduces the `get-workspace-members` step (same as fuel-rs) in order to build each crate individually on CI going forward

### Testing steps
- [ ] CI should pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -521,6 +521,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+    
       - name: Install toolchain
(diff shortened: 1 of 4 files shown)


deekerno (approved):
LGTM. Thanks for working on this; I know it's not easy.

PR #327, opened 2022-11-07
feature: native execution

ra0x3:


examples/block-explorer/explorer-index/src/lib.rs
```diff
@@ -247,10 +249,12 @@ mod explorer_index {
...
                         });
 
-                        Logger::info("LogData types are unused in this example. (>'')>");
+                        WasmLogger::info(
+                            "LogData types are unused in this example. (>'')>",
+                        );
                     }
                     _ => {
-                        Logger::info("This type is not handled yet.");
+                        WasmLogger::info("This type is not handled yet.");
```
ra0x3, line 257: `WasmLogger` --> `Logger::Wasm::info`

PR #328, opened 2022-11-07
Make entire config optional

ra0x3:
### Description
- This PR makes each piece of the config file optional, so devs are free to pass as much or as little in the config file as they want. 
- Whatever vars are not present in the file, defaults are used in their place

### Testing steps


deekerno (approved):
Nice, clean implementation. 👌🏽

PR #357, opened 2022-11-24
Add popular example

### LINKED ISSUE #291: Create a popular/interesting/easy-to-understand use-case indexing example
- We have a few indexer examples, but we need one that is less-techy and able to be easily understood by dApp developers and the broader community
- Ideally we'll have cleaned a few of our hacks/issues up by the time this issue is started

Additional context
- We should actually work with @camiinthisthang on this
- This might have include a frontend

deekerno:
Closes #291.

## Changelog
- Adds an example in which a user can use a dashboard to monitor what kinds of assets have been transferred to a contract
- Adds permissive `CORS` configuration to the API server
- Updates documentation with dashboard example

## Testing

1. `cargo build -p dashboard-example --release`
3. `bash scripts/utils/build_test_wasm_module.bash`
4. `bash scripts/utils/start_test_components.bash`
5. `cargo run --bin fuel-indexer -- --manifest examples/dashboard/manifest.yaml`
6. `bash examples/dashboard/go.bash`
7. Wait a couple seconds while random transfers are generated.
8. Open `examples/dashboard/dashboard-index/components/frontend/index.html` in a web browser to see the transfers.

## Notes
After implementing this example, I wanted to share some things:
- The frontend for this example is deliberately simple as I didn't want to pull in `npm`/`yarn`/the entire Node ecosystem in order to create a simple interface; in fact, the example just pulls in React from a CDN and uses one component. However, if we think that it should be more complex, I can certainly put up another PR after this.
- I learned a bit more about what does and doesn't compile to WASM. Namely, the `wasm32-unknown-unknown` target does not support random generation or generating timestamps without quite a bit of work; the WASM modules compile but will panic and cause an `UnreachableCode` error in the WASM executor. Perhaps it may be prudent to come to a consensus on whether we want to (if at all) endorse a workaround for these use cases; I know having timestamps in this particular example would make it more valuable.
- One of the first things we should do in this next milestone is prioritize GraphQL at the user interaction layer. It's very easy to create incredibly long interface elements and having pagination would be ideal.
- I had to add a CORS configuration to the API server to allow for a browser to properly request information from the API. We should make a decision on whether we want to add a configuration option for this.
- Finally, I added separate node and API binaries that the ones present in our testing suite. This was to generate and demonstrate multiple asset usage. I'm not tied to this methodology; we can certainly use the testing binaries and create a generation script for the example using those. If there's concern about its current implementation, let me know and I can change it.


deekerno:
We've spoken with the app dev team and decided it'll be more valuable to do an indexing backend for one of their Sway apps. Closing for now.

PR #362, opened 2022-11-28
Fix `SocketAddr` issue while connecting to beta-2 testnet

### LINKED ISSUE #361: Running indexer on beta-2 testnet
The indexer fails to run when the host is pointed to the beta-2 testnet. Following is the config and error log.

- The `config.yaml` file
``` yaml

fuel_node:
  # tested with `http` and without `http` also `https`
  host: http://node-beta-2.fuel.network
  port: 80

```

- The error log
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: AddrParseError(Socket)',
fuel-indexer-lib/src/config.rs:164:56
```

- Command used to run
```shell
cargo run -p fuel-indexer -- --config config.yaml

# also
./target/debug/fuel-indexer --config config.yaml
```

- If the config is using `host: <ip_address>`  it works fine as in with read `blocks` or will fail `fetching blocks` but not the above error.

I have also searched this issue and found #168 but that too doesn't work.

atulpatare:
Fixes #361 

- Removed `SocketAddr` using `&str` instead
- Fuel Client auto checks for node URL so no explicit type cast is required
- Ran all tests and followed [contributing.md](https://github.com/FuelLabs/fuel-indexer/blob/a5753c425798a787f4d8aec888f16e4a23d567d9/docs/CONTRIBUTING.md)


ra0x3:
utAck

Looks like unfortunately our testnet domains are not being resolved

```rust
    let addrs: Vec<_> = "node-beta-2.fuel.network"
                .to_socket_addrs()
                .expect("Unable to resolve domain.")
                .collect();
```

@atulpatare Need to pull it down a do a quick test but looks fine

Also will require a quick change on my part to get around the Docker image CI step

AP-Atul:
@ra0x3 done

fuel-indexer/src/service.rs
```diff
@@ -123,14 +123,19 @@ async fn spawn_executor_from_index_asset_registry(
...
-    fuel_node_addr: SocketAddr,
+    fuel_node_addr: &str,
     kill_switch: Arc<AtomicBool>,
     mut executor: T,
     start_block: Option<u64>,
 ) -> impl Future<Output = ()> {
     let mut next_cursor = None;
     let mut next_block = start_block.unwrap_or(1);
-    let client = FuelClient::from(fuel_node_addr);
+    let client = FuelClient::from_str(fuel_node_addr).unwrap_or_else(|_| {
```
ra0x3, line 133: @AP-Atul Since `FuelNodeConfig` implements `ToString` maybe this PR could just be limited to:

```rust
let client = FuelClient::from_str(fuel_node_addr).unwrap_or_else(|e| panic!());
```

Just instantiating from the string instead of the socket?

ra0x3:
@AP-Atul Just tested this locally. Approved pending merge conflict fixes (sorry for the delay).

PR #363, opened 2022-11-28
(WIP) Clean up SQLite type inference

### LINKED ISSUE #347: Cleanup sqlite query usage
- When writing SQLite queries we return a `SqliteRow` from the query and extract each field in the query out manually `row.get(index);`
- We didn't have to do this in Postgres because type inference was supported with Postgres
- Now that https://github.com/launchbadge/sqlx/pull/1960 has landed we should update SQLite queries to remove this temporary/manual type inference functionality

deekerno:
Closes #347.

## Changelog
- Cleans up SQLite type inference


ra0x3:
@deekerno looks good, it's pending https://github.com/FuelLabs/fuel-indexer/pull/366

ra0x3 (approved):
Thanks for this 👌

PR #367, opened 2022-11-28
Reorganize packages

ra0x3:


packages/fuel-indexer-tests/tests/executor.rs
```diff
@@ -0,0 +1,233 @@
+// use fuel_indexer::{
```
ra0x3, line 1: - This test has been giving problems since day 1. 
- It's an integration test. 
- We don't need it. 
- As long as e2e tests work, this test is unnecessary.
- Commenting it out for now, so as to keep it without it being a blocker

deekerno (approved):

PR #370, opened 2022-11-28
Update CI

ra0x3:


.github/workflows/ci.yml
```diff
@@ -212,6 +212,7 @@ jobs:
 
   # TODO: https://github.com/FuelLabs/fuel-indexer/issues/269
   cargo-test-workspace-all-features:
+    if: github.event_name != 'release' && github.event.action != 'published'
```
ra0x3, line 215: If we're cutting a release then CI must've already passed (since release PR's only include version bumps), so we don't need to do this long CI step again

.github/workflows/ci.yml
```diff
@@ -415,8 +416,7 @@ jobs:
 
   publish-docker-image:
     needs: cargo-toml-fmt-check
-
-    if: always()
+    if: github.event_name == 'release' && github.event.action == 'published'
```
ra0x3, line 419: External contributors can't merge PRs because this step requires Fuel org creds. Only Fuel org associates will merge to publish releases.

deekerno (dismissed):

deekerno (approved):

PR #385, opened 2022-12-02
Fix `SocketAddr` issue while connecting to beta-2 testnet

### LINKED ISSUE #361: Running indexer on beta-2 testnet

atulpatare:
@ra0x3 
Fixes #361 
Continues #362 

- Removed `SocketAddr` using `&str` instead
- Fuel Client auto checks for node URL so no explicit type cast is required
- Ran all tests and followed [contributing.md](https://github.com/FuelLabs/fuel-indexer/blob/a5753c425798a787f4d8aec888f16e4a23d567d9/docs/CONTRIBUTING.md)

` I accidentally closed the other PR and had some conflicts locally. Here's a fresh one. `

### DIFF
--- a/packages/fuel-indexer/src/service.rs
+++ b/packages/fuel-indexer/src/service.rs
@@ -26,7 +26,7 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 use std::future::Future;
 use std::marker::{Send, Sync};
-use std::net::SocketAddr;
+use std::str::FromStr;
 use tokio::{
     sync::mpsc::Receiver,
     task::JoinHandle,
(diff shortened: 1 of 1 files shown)


ra0x3 (approved):
Our first external contribution. Thanks 🥳

PR #387, opened 2022-12-03
Add feature `start_block` and fix `next_cursor` logic

### LINKED ISSUE #380: Suggestion regarding `start_block` option
I believe currently if the `start_block` is set in the manifest, the indexer [skips](https://github.com/FuelLabs/fuel-indexer/blob/303d20063835213e18204e1efc143fa58c19d75d/packages/fuel-indexer/src/service.rs#L157) blocks which are not yet reached the `start_block` in this case until `next_block`.

So If I want to index from `1000000` block it will `continue` that many times.

If you initialize [`next_cursor`](https://github.com/FuelLabs/fuel-indexer/blob/303d20063835213e18204e1efc143fa58c19d75d/packages/fuel-indexer/src/service.rs#L131) to `Some(start_block.unwrap_or(1))` then you can just jump to that block due to the `cursor` logic in `PaginatedResult`

AP-Atul:
Closes #380 

Changes
- Starts indexing from block no if defined by `start_block`
```
handler following block values

0 => starts from 1st block
1 => starts from 1st block
>1 => starts from that specific block
>max current block => waits until that block appears
```
- Fixed `next_cursor` logic, if there were no new blocks, the indexer started paginating from `block=1`

### DIFF
--- a/packages/fuel-indexer/src/service.rs
+++ b/packages/fuel-indexer/src/service.rs
@@ -252,11 +252,13 @@ fn make_task<T: 'static + Executor + Send + Sync>(
                 }
             }
 
-            next_cursor = cursor;
-            if next_cursor.is_none() {
-                info!("No next page, sleeping.");
+            if cursor.is_none() {
+                info!("No new blocks to process, sleeping.");
                 sleep(Duration::from_secs(DELAY_FOR_EMPTY_PAGE)).await;
-            };
+            } else {
+                next_cursor = cursor;
+            }
+
             retry_count = 0;
         }
     }
--- a/packages/fuel-indexer/src/service.rs
+++ b/packages/fuel-indexer/src/service.rs
@@ -103,8 +103,14 @@ fn make_task<T: 'static + Executor + Send + Sync>(
     mut executor: T,
     start_block: Option<u64>,
 ) -> impl Future<Output = ()> {
-    let mut next_cursor = None;
-    let mut next_block = start_block.unwrap_or(1);
+    let start_block_value = start_block.unwrap_or(1);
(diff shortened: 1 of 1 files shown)


packages/fuel-indexer/src/service.rs
```diff
@@ -251,11 +252,13 @@ fn make_task<T: 'static + Executor + Send + Sync>(
...
 
-            next_cursor = cursor;
-            if next_cursor.is_none() {
-                info!("No next page, sleeping.");
+            if cursor.is_none() {
+                info!("No new blocks to process, sleeping.");
                 sleep(Duration::from_secs(DELAY_FOR_EMPTY_PAGE)).await;
-            };
+            } else {
+                next_cursor = cursor;
```
AP-Atul, line 259: update `next_cursor` only when the `cursor` guarantees to have data

AP-Atul:
@ra0x3  got it

ra0x3:
Gonna convert this to Draft in the meantime :)

AP-Atul:
@ra0x3 this PR is ready for review btw

ra0x3:
> @ra0x3 this PR is ready for review btw

@AP-Atul Are you not able to request a review from me? (in the Github UI on the upper right). Asking because if you can't then other contributors won't be able to either 🤔

AP-Atul:
@ra0x3  I can't actually see those options. Positive can request review, only option to convert to draft 
Screenshot from 2022-12-05 22-25-27

ra0x3:
@AP-Atul Ok hm, need to update the CONTRIBUTING doc to reflect this. Thanks. I'll review

ra0x3 (approved):

PR #395, opened 2022-12-05
update release ci

ra0x3:
### Description
- Release CI time is almost 2 hours. An untenable position
- Don't run cargo tests when publishing release
  - Release PRs only contain version changes -- they don't contain new code.


deekerno (approved):
Good catch!

PR #403, opened 2022-12-06
Fix pool usage across the indexer

deekerno:
_Note: This does not fix SQLite; this is just a cleanup PR while I continue to fix the issue._

## Changelog
- Consolidates pool usage across application; we were creating multiple pools across the indexer, which meant that `max_connections()` and other settings weren't being respected
- Adds pool and connection options; we're now able to change pool and connection settings (e.g. log level), if so desired
- Changes some `unwrap()`s to `expect()`s

## Testing Plan
CI should pass. Starting the service using Postgres should also work as expected.

### DIFF
--- a/packages/fuel-indexer-api-server/src/api.rs
+++ b/packages/fuel-indexer-api-server/src/api.rs
@@ -105,15 +105,9 @@ pub struct GraphQlApi;
 impl GraphQlApi {
     pub async fn run(
         config: IndexerConfig,
-        connection_pool: Option<IndexerConnectionPool>,
+        pool: IndexerConnectionPool,
         tx: Option<Sender<ServiceRequest>>,
     ) {
-        let pool = match connection_pool {
-            Some(p) => p,
-            None => IndexerConnectionPool::connect(&config.database.to_string())
-                .await
-                .expect("Failed to establish connection pool"),
-        };
         let sm = SchemaManager::new(pool.clone());
         let schema_manager = Arc::new(RwLock::new(sm));
         let config = config.clone();
--- a/packages/fuel-indexer-api-server/src/api.rs
+++ b/packages/fuel-indexer-api-server/src/api.rs
@@ -103,18 +103,22 @@ impl IntoResponse for ApiError {
 pub struct GraphQlApi;
 
 impl GraphQlApi {
-    pub async fn run(config: IndexerConfig, tx: Option<Sender<ServiceRequest>>) {
-        let sm = SchemaManager::new(&config.database.to_string())
-            .await
-            .expect("SchemaManager create failed");
+    pub async fn run(
+        config: IndexerConfig,
+        connection_pool: Option<IndexerConnectionPool>,
+        tx: Option<Sender<ServiceRequest>>,
+    ) {
+        let pool = match connection_pool {
+            Some(p) => p,
+            None => IndexerConnectionPool::connect(&config.database.to_string())
+                .await
+                .expect("Failed to establish connection pool"),
+        };
+        let sm = SchemaManager::new(pool.clone());
         let schema_manager = Arc::new(RwLock::new(sm));
         let config = config.clone();
         let start_time = Arc::new(Instant::now());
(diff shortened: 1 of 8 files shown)


packages/fuel-indexer-api-server/src/api.rs
```diff
@@ -103,19 +103,23 @@ impl IntoResponse for ApiError {
...
 impl GraphQlApi {
-    pub async fn run(config: IndexerConfig, tx: Option<Sender<ServiceRequest>>) {
-        let sm = SchemaManager::new(&config.database.to_string())
-            .await
-            .expect("SchemaManager create failed");
+    pub async fn run(
+        config: IndexerConfig,
+        connection_pool: Option<IndexerConnectionPool>,
+        tx: Option<Sender<ServiceRequest>>,
+    ) {
```
ra0x3, line 110: What is the case where we don't pass a pool, but need to create a pool? The `GraphQLApi` can't really do anything w/o a DB connection so we can change `Option<IndexerConnectionPool>` to `IndexerConnectionPool`

packages/fuel-indexer-api-server/src/bin/fuel-indexer-api-server.rs
```diff
@@ -27,7 +27,7 @@ pub async fn main() -> Result<()> {
 
     info!("Configuration: {:?}", config);
 
-    GraphQlApi::run(config.clone(), None).await;
+    GraphQlApi::run(config.clone(), None, None).await;
```
ra0x3, line 30: Mentioned above ☝️ but the `GraphQlApi` w/o a DB connection can't really do anything so we can pass one here.

packages/fuel-indexer-database/src/lib.rs
```diff
@@ -75,15 +78,20 @@ impl IndexerConnectionPool {
...
                 })
                 .await;
 
                 Ok(IndexerConnectionPool::Postgres(pool))
             }
             "sqlite" => {
                 let pool = attempt_database_connection(|| {
-                    sqlx::Pool::<sqlx::Sqlite>::connect(database_url)
+                    let connection_options = SqliteConnectOptions::from_str(database_url)
+                        .expect("Failed to parse SQLite connection options");
```
ra0x3, line 92: Same here with regard to return `Error` more and `expect`'ing less

packages/fuel-indexer/src/bin/fuel-indexer.rs
```diff
@@ -63,7 +63,11 @@ pub async fn main() -> Result<()> {
         () => (None, None),
     };
 
-    let mut service = IndexerService::new(config.clone(), rx).await?;
+    let pool = IndexerConnectionPool::connect(&config.database.to_string())
+        .await
+        .expect("Failed to open connection pool");
```
ra0x3, line 68: To the "return `Error` more and `expect` less comment" - we can just `?` here

ra0x3 (changes requested):
- Looks fine. 
- Some minor code quality comments

packages/fuel-indexer-database/src/lib.rs
```diff
@@ -75,15 +78,20 @@ impl IndexerConnectionPool {
         match url.scheme() {
             "postgres" => {
                 let pool = attempt_database_connection(|| {
-                    sqlx::Pool::<sqlx::Postgres>::connect(database_url)
+                    let connection_options = PgConnectOptions::from_str(database_url)
+                        .expect("Failed to parse Postgres connection options");
```
ra0x3, line 82: - Looks like my original comment didn't go through, but mentioned that we wanna start `expect`'ing less and start `return`'ing `Error`s more where we can
- This is not just for this PR but in general -- if you see `expect`s try converting them unless there's a good reason they're there

packages/fuel-indexer-api-server/src/api.rs
deekerno: Currently? No case.

I'll change it. 👍🏽

ra0x3 (approved):
✅

PR #404, opened 2022-12-06
general code cleanup

ra0x3:


packages/fuel-indexer/src/service.rs
```diff
@@ -295,97 +289,100 @@ impl IndexerService {
...
+    pub async fn register_indices_from_registry(&mut self) -> IndexerResult<()> {
+        let mut conn = self.pool.acquire().await?;
+
+        let _ = queries::start_transaction(&mut conn)
+            .await
+            .expect("Could not start database transaction");
+
+        let indices = queries::registered_indices(&mut conn).await?;
+
+        println!(">>> INDICIES : {:?}", indices);
```
deekerno, line 365: Is this meant to be left in?

deekerno (approved):

packages/fuel-indexer/src/service.rs
ra0x3: whoops, nope. will remove in follow up

PR #405, opened 2022-12-07
Allow for optional ABI JSON tokens in #[indexer] codegen

### LINKED ISSUE #289: Why do I have to use ABI JSON tokens in `#[indexer]` codegen?
- Whilst working on https://github.com/FuelLabs/fuel-indexer/pull/279 I'm noticing some weirdness
- It appears that you are _required_ to have ABI JSON tokens in the `#[indexer]` codegen
  - I have deugged this quite a bit, and it seems not only must I have an ABI JSON file - but my ABI JSON must at least have 1 item in `types`, and at least 1 item in `functions` -- even if I'm not actually using/referencing these data whatsoever
- I'm not sure what this weirdness is related to - but I'm creating this tracking issue so as to not block #279 
  - The current workaround is to just `default-abi.json` (which is a copy/paste of our `fuel-indexer-test-abi.json` asset), when the user does not specify `abi` in the index manifest
    - This allows us to still mark https://github.com/FuelLabs/fuel-indexer/issues/217 as complete until we fix this current issue
- Ideally, if the user does not pass an `abi` param in their index manifest - we don't read any ABI files and don't generate any ABI tokens

deekerno:
Closes #289.

## Changelog
- Enable optional ABI JSON tokens in #[indexer] codegen
- Remove default ABI JSON file
- Add additional logging to WASM executor
- Fix path in block explorer example schema

## Testing Plan
_No ABI path (using block explorer example)_
1. `cargo build -p fuel-indexer-test-web --release`
2. Build block explorer WASM or use build test WASM script
3. `bash scripts/utils/start_test_components.bash`
4. `cargo run --bin fuel-indexer -- --manifest examples/block-explorer/explorer_index.manifest.yaml`
5. `bash examples/block-explorer/go.bash` to generate blocks
6. Watch the blocks get picked up by the indexer.
7. Check database to ensure that they are present.

_With ABI path_
- Check other examples as one would normally do

## Notes
I spent quite awhile trying to figure out why the WASM module wasn't working, and then I realized that we were panicking on any value that didn't match the types in the ABI. Thus, when there is no content in the ABI, everything is caught by the `_` match arm. To prevent this (and to create a slightly more ergonomic experience), I've removed the `panic!`s from the decoder struct and replaced them with instructional messages in order to help users determine why a type is not being indexed. We should probably figure out a nicer way to do this, but that can wait for a later PR.


packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -478,7 +494,7 @@ fn process_fn_items(
             pub fn decode_logdata(&mut self, rb: u64, data: Vec<u8>) {
                 match rb {
                     #(#logged_types),*
-                    _ => panic!("Unknown logged type id '{}'.", rb),
+                    _ => Logger::warn("Unknown logged type ID; check ABI to make sure that logged types are well-formed")
```
ra0x3, line 497: - So we still want to keep these compile time panics
- If the user is trying to decode unrecognized logged types that's a sign of a bigger issue -- you don't wanna fail silently there

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -442,15 +454,19 @@ fn process_fn_items(
...
+                    }
                 }
             }
 
             fn decode_type(&mut self, ty_id: usize, data: Vec<u8>) {
                 match ty_id {
                     #(#abi_decoders),*
-                    _ => panic!("Unkown type id '{}'.", ty_id),
+                    _ => {
+                        Logger::warn("Unknown type ID; check ABI to make sure types are well-formed");
```
ra0x3, line 468: - Same here. If we're trying to decode unknown type IDs that's a sign of a bigger issue. 
- Still want to panic

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -442,15 +454,19 @@ fn process_fn_items(
             fn selector_to_type_id(&self, sel: u64) -> usize {
                 match sel {
                     #(#abi_selectors)*
-                    // TODO: should handle this a little more gently
-                    _ => panic!("Unknown type id."),
+                    _ => {
+                        Logger::warn("Unknown selector; check ABI to make sure function outputs match to types");
```
ra0x3, line 458: - Same here.
- In all these cases we want the best/most-informative error possible -- but still wanna panic.

packages/fuel-indexer/src/executor.rs
```diff
@@ -225,7 +225,17 @@ impl Executor for WasmIndexExecutor {
         let res = spawn_blocking(move || fun.call(ptr, len)).await?;
 
         if let Err(e) = res {
-            error!("WasmIndexExecutor handle_events failed: {e:?}.");
+            error!("WasmIndexExecutor handle_events failed: {}.", e.message());
+            let frames = e.trace();
+            for (i, frame) in frames.iter().enumerate() {
```
ra0x3, line 230: Nice 👌

ra0x3 (changes requested):
- Left some comments regarding keeping the `panic` since failing silently might lead to undefined behavior
- Will pull this down and test it now

ra0x3:
Copy/pasting from Slack

> So after playing around with it, I think `Logger::warn` is fine. In the selectors function , I think we should return `size::MAX` for an unknown selector. Lastly, we can create a tracking issue for “why are selectors unknown for the explorer example?” — off the top of my head I would expect all selctors to be known. Then :white_check_mark: . Thanks for cleaning this up

ra0x3 (approved):
Glad this got cleaned up :shipit:

PR #413, opened 2022-12-08
Add `forc index stop`

### LINKED ISSUE #396: Add `forc index stop` command
- With the completion of https://github.com/FuelLabs/fuel-indexer/pull/373 we have the ability to stop indices via API requests
- We should add this functionality to the `forc index` plugin

Additional context
- I imagine this might work somewhat similarly to:

```bash
forc index stop --manifest path/to/my_index.manifest.yaml --auth my-foo
```
- The `--auth` flag should be similar to the `--auth` flag used for `forc index deploy`
  - The flag is in the the CLI opts but it is unused

deekerno:
Closes #396.

## Changelog
- Add `forc index stop` command
- Move `extract_manifest_fields` to utils
- Clean up docstrings for other commands

## Testing Steps
1. `cargo build --release -p forc-index`
2. `cp target/release/forc-index ~/.fuelup/bin`
3. Start an index and verify that it indexes data
4. `forc index stop --manifest [same_manifest_as_running_index] --auth foo`
5. Logging should say that index has been stopped.
6. Verify that indexer no longer indexes data.

### DIFF
--- a/plugins/forc-index/src/ops/forc_index_stop.rs
+++ b/plugins/forc-index/src/ops/forc_index_stop.rs
@@ -1,0 +1,61 @@
+use crate::cli::StopCommand;
+use crate::utils::extract_manifest_fields;
+use reqwest::{
+    blocking::Client,
+    header::{HeaderMap, AUTHORIZATION},
+    StatusCode,
+};
+use serde_json::{to_string_pretty, value::Value, Map};
+use std::fs;
+use std::io::Read;
+use tracing::{error, info};
+
+pub fn init(command: StopCommand) -> anyhow::Result<()> {
+    let mut manifest_file = fs::File::open(&command.manifest).unwrap_or_else(|_| {
+        panic!(
+            "Index manifest file at '{}' does not exist",
+            command.manifest.display()
+        )
+    });
+
+    let mut manifest_contents = String::new();
+    manifest_file.read_to_string(&mut manifest_contents)?;
+    let manifest: serde_yaml::Value = serde_yaml::from_str(&manifest_contents)?;
+
+    let (namespace, identifier, _, _) = extract_manifest_fields(manifest)?;
+
+    let target = format!("{}/api/index/{}/{}", &command.url, &namespace, &identifier);
+
+    let mut headers = HeaderMap::new();
+    headers.insert(
+        AUTHORIZATION,
+        command.auth.unwrap_or_else(|| "fuel".into()).parse()?,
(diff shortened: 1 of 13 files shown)


plugins/forc-index/src/ops/forc_index_stop.rs
```diff
@@ -0,0 +1,61 @@
...
+    header::{HeaderMap, AUTHORIZATION},
+    StatusCode,
+};
+use serde_json::{to_string_pretty, value::Value, Map};
+use std::fs;
+use std::io::Read;
+use tracing::{error, info};
+
+pub fn init(command: StopCommand) -> anyhow::Result<()> {
+    let mut manifest_file = fs::File::open(&command.manifest).unwrap_or_else(|_| {
```
ra0x3, line 14: What about?

```rust
// With fuel_indexer_lib as a dependency
use fuel_indexer_lib::manifest::Manifest;

let manifest: Manifest = Manifest::from_file(command.manifest)?;
```

plugins/forc-index/src/ops/forc_index_stop.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    let target = format!("{}/api/index/{}/{}", &command.url, &namespace, &identifier);
+
+    let mut headers = HeaderMap::new();
+    headers.insert(
+        AUTHORIZATION,
+        command.auth.unwrap_or_else(|| "fuel".into()).parse()?,
+    );
+
+    info!("\n🛑 Stopping index at {}", &target);
```
ra0x3, line 35: Love it 👌 

Maybe `"Stopping index 'namespace.identifier' at {}"` (just being more explicit about what is being stopped)

plugins/forc-index/src/utils/mod.rs
```diff
@@ -3,3 +3,42 @@ pub mod defaults;
 pub(crate) fn dasherize_to_underscore(s: &str) -> String {
     str::replace(s, "-", "_")
 }
+
+pub(crate) fn extract_manifest_fields(
```
ra0x3, line 7: 👌

ra0x3 (changes requested):
Small pieces of feedback that aren't make or break
- Using `fuel_indexer_lib` for some utility stuff
- Additional info in the logs (so it's _extremely_ clear to the user what's going on)

Looks good 👌

ra0x3 (approved):

PR #421, opened 2022-12-14
chore: Upgrade dependencies

deekerno:
Closes #334, #386.

## Changelog
Upgrades:
| Dependency | Current Version | New Version | crates.io Link |
| --------------- | --------------- | --------- | ---------- |
| fuel-tx | 0.23 | 0.23 | https://crates.io/crates/fuel-tx |
| fuel-gql-client | 0.14 | 0.15 | https://crates.io/crates/fuel-gql-client |
| fuel_types | 0.5 | 0.5 | https://crates.io/crates/fuel-types |
| fuels | 0.30 | 0.32 | https://crates.io/crates/fuels |
| fuels-types | 0.30 | 0.32 | https://crates.io/crates/fuels-types |
| fuels-core | 0.30 | 0.32 | https://crates.io/crates/fuels-core |
| fuels-abigen-macro | 0.30 | removed | https://crates.io/crates/fuels-abigen-macro |

PR #426, opened 2022-12-15
Re-integrate block producer

### LINKED ISSUE #334: Re-integrate block producer into block-explorer example
The `producer` field has been removed from the `Block` struct that is returned from `fuel-gql-client`; we should find a way to re-integrate that information back into the block explorer example.

deekerno:
Closes #334.

## Changelog
- Re-integrates block producer into `BlockData`

## Notes
Since #421 is blocked by the proverbial dependency hell, I elected to do this in its own PR to move things forward.


deekerno:
It looks like this is failing due to an error inside of `fuel-gql-client` when it attempts to decode the response for the `blocks()` endpoint.

deekerno:
Closing this for now.

PR #427, opened 2022-12-16
fix (fuel-indexer): Make sure that handle for old executor is aborted on redeploy

deekerno:
## Changelog
- Ensures that currently running handle is aborted if index is redeployed

## Testing Plan
On master:

1. Start the indexer with an index.
2. Generate test data and verify that data is indexed.
3. `forc index deploy` with the same manifest.
4. `forc index stop` with the same manifest.
5. Generate test data again; one should see that the data is still being indexed.

On this branch:

1. Start the indexer with an index.
2. Generate test data and verify that data is indexed.
3. `forc index deploy` with the same manifest.
4. `forc index stop` with the same manifest.
5. Generate test data again; one should see that the data is **not** indexed.

### DIFF
--- a/packages/fuel-indexer/src/service.rs
+++ b/packages/fuel-indexer/src/service.rs
@@ -419,7 +419,11 @@ impl IndexerService {
                         .expect("Failed to spawn executor from index asset registry")
                         .0;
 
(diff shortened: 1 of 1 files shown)


ra0x3 (approved):
utACK

PR #430, opened 2022-12-19
fix (fuel-indexer,fuel-indexer-types): Re-integrate block producer

### LINKED ISSUE #334: Re-integrate block producer into block-explorer example

deekerno:
Closes #334.

## Changelog
- Re-integrates block producer into `BlockData`

## Notes
Since #421 is blocked by the proverbial dependency hell, I elected to do this in its own PR to move things forward.

## Testing Plan
1. Run test components.
2. Start indexer service with block explorer example.
3. Generate blocks for test data.
4. Confirm that `producer` is present when adding values to database.


examples/block-explorer/explorer-index/src/lib.rs
```diff
@@ -48,12 +48,15 @@ mod explorer_index {
         // a block entity `Block` that we can persist to the database. The `Block` type below is
         // defined in our schema/explorer.graphql and represents the type that we will
         // save to our database.
-        //
-        // Note: There is no miner/producer address for blocks in this example; the producer field
-        // was removed from the `Block` struct as part of fuel-core v0.12.
+        let producer = match block_data.producer {
```
ra0x3, line 51:
```suggestion
        let producer = block_data.producer.unwrap_or(Bytes32::zeroed());
```

packages/fuel-indexer/src/service.rs
```diff
@@ -151,6 +151,11 @@ fn run_executor<T: 'static + Executor + Send + Sync>(
 
             let mut block_info = Vec::new();
             for block in results.into_iter().rev() {
+                let producer = match block.block_producer() {
```
ra0x3, line 154:
```suggestion
                let producer = block.block_producer().map(|pk| pk.hash());
```

ra0x3 (changes requested):
2 small code nits but otherwise looks good 👌

ra0x3 (approved):

Issue #269, opened 2022-10-17
Move `fuel-indexer-test-web` components to `fuel_indexer_tests::fixtures`


ra0x3:
- We are currently using [a few components](https://github.com/FuelLabs/fuel-indexer/blob/master/scripts/start_test_components.bash) to aide our unit/integration testing
- These components are _external_ in the sense that they are required to be built outside of the context of `cargo test` (as prerequisites)
- Ideally, we should only use _internal_ testing components (i.e., within the context of `cargo test`)

Additional context
- This primarily means spinning up the fuel-node, wallet, etc within a testing fixture (i.e., `fuel_indexer_tests::fixtures`) as a replacement for the web components
- This also means using `IndexerService::run()` as opposed to building the indexer on CI 
  - This would be a huge help for https://github.com/FuelLabs/fuel-indexer/issues/252
- I previously did this in a PR but had some issues getting `IndexerService` to see the triggered/indexable events

ra0x3:
Update
- This will consist of a few parts
- [ ] Adding an Indexer service fixture to e2e tests
- [ ] Adding a web service fixture to e2e tests
- [ ] Add some type of Fuel node fixture to e2e tests OR booting up the Fuel node via a CLI command
- [ ] Adding a fuel-indexer-api-server/axum service fixture for the metrics e2e tests

Issue #432, opened 2022-12-20
Asset upload fails with `500 - InternalServerError` w/ Sqlite backend


ra0x3:
### Description
- Asset upload/deployment currently fails w/ Sqlite backend

```text
➜  fuel-indexer git:(rashad/391-breakout-tests) curl -v http://127.0.0.1:29987/api/index/fuel_indexer_test/index1 \ 
    -F "manifest=@packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml" \
    -F "wasm=@packages/fuel-indexer-tests/assets/fuel_indexer_test.wasm" \
    -F "schema=@packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql" \
    -H 'Content-type: multipart/form-data' -H "Authorization: foo" | json_pp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 127.0.0.1:29987...
* Connected to 127.0.0.1 (127.0.0.1) port 29987 (#0)
> POST /api/index/fuel_indexer_test/index1 HTTP/1.1
> Host: 127.0.0.1:29987
> User-Agent: curl/7.79.1
> Accept: */*
> Authorization: foo
> Content-Length: 456814
> Content-Type: multipart/form-data; boundary=------------------------dc7953782be89030
> 
} [755 bytes data]
* We are completely uploaded and fine
100  446k    0     0  100  446k      0  87248  0:00:05  0:00:05 --:--:-- 87395* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< content-type: application/json
< content-length: 53
< date: Tue, 20 Dec 2022 01:35:10 GMT
< 
{ [53 bytes data]
100  446k  100    53  100  446k     10  86623  0:00:05  0:00:05 --:--:--    13
* Connection #0 to host 127.0.0.1 left intact
{
   "details" : "Inernal server error.",
   "success" : "false"
}
```

### Reproduction steps
- Start the indexer with a Sqlite backend
- Deploy an index

### Expected result
- Index should be registered successfully

deekerno:
Just adding additional information here: it seems that it's completely broken in SQLite as it doesn't work in our integration tests nor through `curl`.

PR #437, opened 2022-12-21
enhancement (fuel-indexer-tests): Add axum fixture for metrics tests

deekerno:
Part of #269.

## Changelog
- Adds `axum` API server fixture for metrics tests
- Refactor `GraphQLApi::run` into `build`, `run`, and `build_and_run`
- Adjust tests
- Remove `fuel-indexer` build step from testing portion of CI

## Testing Plan
- Run each of the following _without using the `start_test_components.bash` script_:
-- `test_metrics_endpoint_returns_proper_count_of_metrics_postgres`
-- `test_database_postgres_metrics_properly_increments_counts_when_queries_are_made`
-- `test_metrics_endpoint_returns_proper_count_of_metrics_sqlite` (needs to be uncommented)
-- `test_database_sqlite_metrics_properly_increments_counts_when_queries_are_made` (needs to be uncommented)


deekerno:
To keep this small, I'll be doing a different PR for #432 and #440.

packages/fuel-indexer-tests/tests/e2e/metrics_postgres.rs
```diff
@@ -1,10 +1,30 @@
...
-use fuel_indexer_tests::fixtures::{http_client, postgres_connection_pool};
+use fuel_indexer_tests::fixtures::{
+    api_server_app_postgres, http_client, indexer_service_postgres,
+    postgres_connection_pool,
+};
+use tokio::task::spawn;
 
 #[tokio::test]
-#[cfg(all(feature = "e2e", feature = "postgres"))]
+#[cfg(all(feature = "e2e", feature = "postgres", feature = "metrics"))]
```
ra0x3, line 11: - Is there a reason for this additional `metrics` feature flag? Doesn't seem like something we'd want metrics testing feature flagged.
  - `e2e` is feature flagged because those tests run differently than all other tests
  - `postgres` is feature flagged cause that decides which suite of database tests to run

packages/fuel-indexer-api-server/src/api.rs
```diff
@@ -100,16 +100,15 @@ impl IntoResponse for ApiError {
 pub struct GraphQlApi;
 
 impl GraphQlApi {
-    pub async fn run(
+    pub async fn build_router(
```
ra0x3, line 103: opinion nit: `build` ?

ra0x3 (changes requested):
@deekerno  Thanks for tackling this, it looks like it's almost there. A few notes

- I pulled down the branch and tested the Postgres tests - they worked ✅ 
- The SQLite tests are commented out because we were testing the metrics via building the `fuel-indexer` binary and pinging the web API. We could only start that binary with one type of connection - so we used Postgres.
  - Now that we don't need to do that anymore - we want to include (uncomment) these SQLite tests as well 
- Now that we have this axum fixture, we don't need to [build the `fuel-indexer` crate in the `cargo-test-workspace-all-features` step anymore](https://github.com/FuelLabs/fuel-indexer/blob/master/.github/workflows/ci.yml#L256)
  - We still need `fuel-indexer-test-web` but I can remove that in a follow up PR (since we don't need the `web-api` binary. I can just include the `fuel-node` binary in our testing assets)
- Lastly, now that we don't have to build the `fuel-indexer` binary anymore to test this metrics stuff, I wouldn't consider these `e2e` tests anymore. Now they're just normal tests.
  - So we can move them to `packages/fuel-indexer-tests/tests/integration/web_api_postgres.rs` and `packages/fuel-indexer-tests/tests/integration/web_api_sqlite.rs`
    - I could also see the case for being more specific and putting them in `metrics_sqlite` and `metrics_postgres`
      - My thinking is that if  the`web_api_*` modules get too big we can always break them down/out later

Good stuff, almost there 👌

ra0x3:
> To keep this small, I'll be doing a different PR for #432 and #440.

Yep. Those should stay separate 👌

packages/fuel-indexer-tests/tests/e2e/metrics_postgres.rs
deekerno: Nah, I was testing a bunch of things and forgot to remove it. 😓

packages/fuel-indexer-api-server/src/api.rs
```diff
@@ -151,6 +150,17 @@ impl GraphQlApi {
 
         let app = Router::new().nest("/api", api_routes);
 
+        Ok(app)
+    }
+
+    pub async fn run(
```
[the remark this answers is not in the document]
deekerno: Sounds good to me!

ra0x3 (approved):
:shipit:

PR #442, opened 2022-12-22
chore: remove test web components

ra0x3:
### Description
- Removes `web-api` from CI
- Removes old testing modules from e2e.bash
- Breaks `fuel-indexer-test-web` into separate `web-api` and `fuel-node` crates
  - Because we still need `fuel-node` on CI (for now), but we don't need `web-api` on CI. However, we still use `web-api` for localhost testing/dev 
- Adds `postgres` and `sqlite` feature testing to `cargo-verifications`

### Testing steps 
- [ ] CI should pass

### Changelog 
- chore: remove test web components


deekerno (approved):
🚢

PR #465, opened 2023-01-10
enhancement: add support for blob type

ra0x3:
### Description
- PR adds support for a `Blob` type
- Implementation details
  - `Blob` was added as a `varchar(10485760)` (max `varchar` size in Postgres, open to lowering that size later) and _not_ a `bytea`
    - This was done because of the way we use our `INSERT` queries in `fuel_indexer::database::put_object`
      - The only bytes being used in `put_object` are the entire serialized `Vec<FtColumn>`. This is nice because it only requires 1 known `bind` in the query builder [here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-database/postgres/src/lib.rs#L20)
      - If we tried to make `Blob` a `bytea` then we run into a situation where we have any arbitrary amount of byte payloads being passed to `put_object` (which would be a mess 🔥 )
   - In this implementation, we just `hex::encode` a `Vec<u8>` and then `hex::decode` when loading the item from the DB
   - Since we now have `Blob` being used differently that it was previously, I created a new `Object` type to act as the default `FtColumn` for all `object` fields in schema (wanted to keep `Blob` and `Object` separate)

### Testing steps
- [ ] The `hello-world` was updated to include this new `Blob` type
- [ ] This `Blob` type was also added to the `fuel-indexer-test` test index, and included in e2e tests, so CI should pass

### Changelog
- enhancement: add support for blob type


packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql
```diff
@@ -16,6 +16,7 @@ type QueryRoot {
     log: Log
     scriptresult: ScriptResult
     messageout: MessageOut
+
```
deekerno, line 19: Nit: Unnecessary newline here?

deekerno (approved):
Nice!

PR #472, opened 2023-01-12
fix (CI): Remove `--target` flag from `cargo install cross`

deekerno:
It looks like the target flag during the `cargo install cross` step is causing some issues: https://github.com/FuelLabs/fuel-indexer/actions/runs/3898047114/jobs/6657141825


hrryndrsn (approved):
Looks good

PR #473, opened 2023-01-12
fix (CI): Re-add cargo-install action to install `cross`

deekerno:
## Changelog
- Re-add cargo-install action to install `cross`; there seems to be release issues when using the built-in `cargo`


hrryndrsn (approved):
:+1:

PR #485, opened 2023-01-16
enhancement: Support optional fields in GraphQL schema

### LINKED ISSUE #480: Support optional fields in GraphQL schema
- [GraphQL support optional/nullable fields in schema](https://hasura.io/blog/graphql-nulls-cheatsheet/)
- Currently in all of our examples and tests, all fields are required slash not nullable
- We should add support for nullable/optional fields in GraphQL schema so that users can save entities with null fields, and query for objects with null fields
- The following schema should be valid and work properly

```graphql
schema {
    query: QueryRoot
}

type QueryRoot {
    block: Block
    tx: Tx
}

type Block {
    id: ID!
    height: UInt8!
    hash: Bytes32! 
}

type Tx {
    id: ID!
    block: Block!
}
```

Additional context
- ⚠️ `id` fields (which every object is required to have) should _not_ be nullable, ever ⚠️ 
- [Some of this functionality actually exists](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-macros/src/schema.rs#L38) but is slightly half-baked and untested
- I should be able to save entities that have optional fields, and I should be able to query for objects that have nullable fields
- https://github.com/FuelLabs/fuel-indexer/issues/460 is _not_ a prerequisite to this work

deekerno:
Closes #480.

## Changelog
- Allows for optional fields in a schema
-- Special care was taken to ensure that ID fields cannot be made optional
- Adds test to ensure query fragments are correctly generated when types are not present
- Adds e2e test in which event has two optional fields: one with an assigned value, and another with a `None` value
- Re-generate `insta` snapshots for tests (this is where the bulk of the files changed value comes from)

## Testing Plan
New tests covering optional types were added; CI should pass.

## Notes
There is virtually no API change for the end user. A value for an optional field should be wrapped in `Some()`, if present; otherwise, the value should be `None`  (just as one would expect in Rust).


packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -59,13 +60,30 @@ fn process_field<'a>(
     let typ = process_type(types, field_type, true);
     let ident = format_ident! {"{}", name};
 
-    let extractor = quote! {
-        let item = vec.pop().expect("Missing item in row");
-        let #ident = match item {
-            FtColumn::#typ(t) => t,
-            _ => panic!("Invalid column type {:?}", item),
-        };
+    // Type may be nullable, so let's grab the actual column type
```
ra0x3, line 63: @deekerno We can remove these types of comments. They don't add any real information beyond how the code reads

packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -91,13 +109,23 @@ fn process_fk_field<'a>(
     let typ = process_type(types, &field_type, false);
     let ident = format_ident! {"{}", field_name.to_lowercase()};
 
+    let (_, column_type) = get_column_type(typ.clone());
+
+    // TODO: We should make a formal decision as to whether we'll allow
```
ra0x3, line 114: Since both Postgres and SQLite support `NULL` foreign keys, we should do the same (again, for everything except `ID` types). Apologies for not including this in the original issue.

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -453,6 +453,60 @@ mod tests {
         assert_eq!(statements[2], create_thing2_schema);
     }
 
+    #[test]
+    fn test_schema_builder_for_basic_postgres_schema_with_optional_types_returns_proper_create_sql(
```
ra0x3, line 457: Nice 👍

packages/fuel-indexer-schema/src/lib.rs
```diff
@@ -154,94 +154,178 @@ impl FtColumn {
     pub fn query_fragment(&self) -> String {
         match self {
             FtColumn::ID(value) => {
-                format!("{}", value)
+                if let Some(val) = value {
+                    format!("{}", val)
+                } else {
+                    panic!("ID schema field is required and cannot be None")
```
ra0x3, line 160: Maybe more clarity here?

> "Schema fields of type `ID` cannot be nullable"

packages/fuel-indexer-schema/src/lib.rs
```diff
@@ -154,94 +154,178 @@ impl FtColumn {
     pub fn query_fragment(&self) -> String {
         match self {
             FtColumn::ID(value) => {
-                format!("{}", value)
+                if let Some(val) = value {
```
ra0x3, line 157: Maybe we can cleanup all of this `if let Some(x) = y { a } else { b }`?

`String::from("NULL")` looks ripe for a default value `const NULL_VALUE: &str = "NULL" (maybe using `lazy_static for this)

ra0x3 (changes requested):
@deekerno  
- Finishing this review as a first pass to unblock you
- I see tests were added, and the test index was updated, good stuff 👏 
- This PR currently covers the "frontend" portion - there's also another part of this - the "backend" piece, which is querying this nullable types for GraphQL
  - With regard to this, I'd say if you don't add a test for this (querying nullable fields) functionality, add those testing steps to your PR **Testing steps**
    - So far in the PR you've proven that we can _build_ nullable fields, but the second part of that is can we _query_ for nullable fields and have them be nullable as expected
    
Good stuff

ra0x3 (approved):
This looks good!
- Minimal API changes
- Tests included (phew man, service-level integration testing in Rust is so terse)
- New GraphQL API tests 🙌 

> There is virtually no API change for the end user

Music to my ears 👂 

✅

packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql
```diff
@@ -92,3 +91,10 @@ type MessageOut {
     len: UInt8!
     digest: Bytes32!
 }
+
+type OptionEntity {
+    id: ID! @unique
+    int_required: UInt8!
+    int_optional_some: UInt8
```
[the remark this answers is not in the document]
deekerno: We can definitely rename! I just wanted to make it exceedingly clear what results should be expected in the tests.

PR #373, opened 2022-11-29
Allow for live index to be stopped

deekerno:
Closes #348.

## Changelog
- Adds `IndexStopRequest` struct for stopping a live index
- Adds new `ServiceRequest` enum, and adds `AssetReloadRequest` and `IndexStopRequest` to it
- Adds functionality to API to stop execution of an index
- Removes the `run_once` flag for the service; this will be added back in soon as part of #383 

## Testing Steps

1. Rebuild WASM, if necessary
2. Start web components `bash scripts/utils/refresh_test_db.bash` and `bash/scripts/utils/start_web_components.bash`
3. Start indexer `SQLX_OFFLINE=1 cargo run --bin fuel-indexer -- --manifest packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml`
4. Trigger event and see it's indexed `curl -v -X POST http://127.0.0.1:8000/ping`
5. Delete index request `curl -v -X DELETE http://127.0.0.1:29987/api/index/fuel_indexer_test/index1 -H "Authorization: fuel"` -- you should see something in the service logs stating that the index has been stopped
6. Make another request to index event `curl -v -X POST http://127.0.0.1:8000/ping` -- there should be no activity for that index in the logs and there should be no data added to the database as a result of this call

PR #487, opened 2023-01-17
fix: refactor service to retire futs

ra0x3:
### Description
- I found a bug in our `IndexerService` that involves executor tasks not being awaited
- If you go try to run the `hello-world-native` index example on master, it won't work - because the execution task (created by `fuel_indexer::executor::run_executor` (in this PR)) is never awaited.
- If we go to [an older version of `fuel_indexer::service::IndexerService::run`](https://github.com/FuelLabs/fuel-indexer/blob/f667dd9d02b01dc14edeae0c12fb6efa0c8b717b/indexer/src/service.rs#L306) we see that these execution tasks _used_ to be awaited via a stream of unordered futures
- However, in https://github.com/FuelLabs/fuel-indexer/pull/373 we removed this streaming functionality in order to be able to pass `ServiceRequest`s back to the `IndexerService` so that we could `abort` spawned tasks (`JoinHandle<()>`)
- I believe the service is currently running by...luck?
- This PR makes a few small changes:
  - Moves executor logic out of `fuel_indexer::service`, and into `fuel_indexer::executor`
    - **Just moving this code is responsible for the overwhelming majority of this PR's code diff**
  - Adds the old streaming logic back
  - Uses a separate tokio task to listen for `ServiceRequest`s, and add new indexers (created by deploying indices to the GraphQL API) to the futures stream -- instead of trying to add them to `handles: HashMap<String, JoinHandle<()>>`
    - There is no way in safe Rust (that I know of after my research) to at the same time have these `handles` owned as futures in `fuel_indexer::service::IndexerService::run` _and_ be able to push new execution/indexer tasks to these handles as well

### Testing steps
- [ ] CI should pass
- [ ] Run the `hello-world-native` example:
```bash
# start your fuel node
cargo run --bin fuel-node

# run the hello world native example
cargo run --bin hello-index-native -- --manifest examples/hello-world-native/hello_index_native.manifest.yaml

# generate some data
cargo run --bin hello-bin

# !! you should see the data being indexed !!
```
- [ ] Run the `block-explorer` example and deploy another indexer at the same time

```bash
# run the block explorer example
cargo run --bin fuel-indexer -- --fuel-node-host node-beta-2.fuel.network --fuel-node-port 80 --manifest examples/block-explorer/explorer_index.manifest.yaml

# deploy the fuel_indexer_test index 
curl -v http://127.0.0.1:29987/api/index/fuel_indexer_test/index1 \
    -F "manifest=@packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml" \
    -F "wasm=@packages/fuel-indexer-tests/assets/fuel_indexer_test.wasm" \
    -F "schema=@packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql" \
    -H 'Content-type: multipart/form-data' -H "Authorization: foo" | json_pp

# !! you should see the data being indexed for fuel_examples.block AND fuel_indexer_test.block !!
```
- [ ] Now with the `explorer-index` and the `fuel-indexer-test` index running at the same time, try to stop the `explorer-index`
  - Note that the index stops after about 2 seconds
```curl
curl -v -X DELETE http://127.0.0.1:29987/api/index/fuel_examples/explorer_index -H "Authorization: fuel"
```

### Changelog 
- fix: refactor service to retire futs


packages/fuel-indexer/src/executor.rs
```diff
@@ -17,10 +17,301 @@ use wasmer::{
...
+
+                empty_block_reqs += 1;
+
+                if empty_block_reqs == MAX_EMPTY_BLOCK_REQUESTS {
+                    error!("No blocks being produced, giving up. <('.')>");
+                    break;
+                }
+            } else {
+                next_cursor = cursor;
+            }
```
deekerno, line 304: I think that we need to reset `empty_block_reqs` in the else block here. Otherwise, we keep incrementing with the old value even if a new block has successfully been received.

deekerno (changes requested):
Testing steps work! Just one comment.

packages/fuel-indexer/src/executor.rs
ra0x3: @deekerno Doh! Good find (I originally thought you were talking about something else). Fix incoming

deekerno (approved):

PR #491, opened 2023-01-19
update readme

ra0x3:


docs/README.md
```diff
@@ -215,356 +153,385 @@ for these executables in your `$PATH`.
...
+
+##### 2.3.2 Deploying your index to your Fuel indexer service
+
+With our database and Fuel indexer indexer containers up and running, we'll deploy the index that we previously created. If all goes well, you should see the following:
 
-impl Greet for Contract {
-    fn new_greeting(id: u64, greeting: str[32], person_name: str[32]) {
-        log(Greeting{ id, greeting, person: Person{ name: person_name }});
+```text
+forc-index deploy --manifest hello_index.manifest.yaml --url http://0.0.0.0:29987
```
deekerno, line 259: Similar comment here about deploying `hello_index` but mentioning the block explorer above and then querying for the block explorer types in the next section.

docs/README.md
```diff
@@ -215,356 +153,385 @@ for these executables in your `$PATH`.
...
+]
+```
+
+#### Finished! 🥳
 
+Congrats, you just created, built, and deployed your first index on the world's fastest execution layer. For more detailed info on how the Fuel indexer service works, make sure you [**read the book**](https://fuellabs.github.io/fuel-indexer/master/).
+
+## Plugin
+
+The primary way of developing Fuel indexers for users that are not also contributors, is via the `forc-index` plugin. The `forc-index` plugin, is a CLI tool that is bundled with Fuel's primary CLI tooling interface, [`forc`](https://github.com/FuelLabs/sway/tree/master/forc) ("Fuel Orchestrator").
```
deekerno, line 348: Nit: Remove the commas after contributors and plugin.

Also, I think it's more clear if we change "users that are not also contributors" to "end users". In my estimation, an end user is usually considered to be different from a contributor on most projects, so I think it would be fine here.

docs/README.md
```diff
@@ -215,356 +153,385 @@ for these executables in your `$PATH`.
...
+
+As mentioned in the [dependencies](#dependencies) section, the `forc-index` plugin is made available once you download [`fuelup`](#fuelup).
+
+If you've gone through the [Quickstart](#quickstart), you should hopefully already have `forc-index` installed and available in your `PATH`.
```
deekerno, line 352: Consider:
> If you've successfully gone through the [Quickstart](#quickstart), you should already have `forc-index` installed and available in your `PATH`.

docs/README.md
```diff
@@ -215,356 +153,385 @@ for these executables in your `$PATH`.
...
-            },
-        };
-
-        // Both entity saves will occur in the same transaction
-        greeting.save();
-        greeter.save();
-    }
-}
-```
+Within the context of the Fuel indexer, WebAssembly (WASM) modules are binaries that are compiled to a `wasm32-unknown-unknown` target, that can be deployed to a running indexer service.
```
deekerno, line 423: Consider:
> ...`wasm32-unknown-unknown` target, which can then be deployed...

deekerno (changes requested):
Left a number of comments.

Overall, I think it's good and much less scary to read at first glance. One general comment I have for the README overall is to use paragraphs rather than bulleted sentences in some parts (mainly early in the PR) as it will give the README more of the flow of a document rather than that of lists under separate headings.

plugins/forc-index/README.md
```diff
@@ -36,18 +36,26 @@ Deploy a given index project to a particular endpoint
...
 
-### `forc index stop`
+### `forc index remove`
 
-Stop a running index
+Kill a running indexer
 
 ```bash
-forc index stop --url https://index.swayswap.io --manifest my_index.manifest.yaml
+forc index remove --url https://index.swayswap.io --manifest my_index.manifest.yaml
```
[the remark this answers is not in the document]
ra0x3: @deekerno  Let's keep this review to the top-level README (I know I updated the plugin docs here), and follow up with a PR for the plugin README

docs/README.md
```diff
@@ -9,189 +9,127 @@
...
+3. Query the indexer service for indexed data using GraphQL.
 
-For other installation options and information, visit the [install](https://www.rust-lang.org/tools/install) page of the Rust website.
+> IMPORTANT: Docker is a prerequisite for using this Quickstart. If Docker is not installed on your machine, please review the Docker installation instructions [here](https://docs.docker.com/engine/install/).
 
-Alternatively, you can [build Cargo from source.](https://github.com/rust-lang/cargo#compiling-from-source)
+### 1. Setting up your environment
 
-### `fuelup`
+In this Quickstart, we'll use Docker's Compose to spin up a Fuel indexer service with a Postgres database backend. We will also use Fuel's toolchain manager [`fuelup`](https://github.com/FuelLabs/fuelup) in order to install the `forc-index` binary that we'll use to develop our index.
```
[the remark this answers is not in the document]
ra0x3: So "Compose" is a component of "Docker" (even though everyone always says "Docker Compose"). You can see  [here](https://docs.docker.com/compose/) how Docker refers to "Compose" - thus me referencing "Docker's Compose". Let me know if it still doesn't make sense

docs/README.md
```diff
@@ -215,356 +153,385 @@ for these executables in your `$PATH`.
 +--------+------------------------+----------------------------------------------------------------------------+
 ```
 
+#### 2.2 Creating a new index
+
+Now that we have our development environment setup, the next step is to create an index.
+
+```bash
+forc index new hello-index --namespace my_project && cd hello-index
```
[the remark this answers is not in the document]
ra0x3: @deekerno 
- So the example is called "Hello Index", but if you check the default code that gets created with `forc index new`, it's just indexing blocks and transactions. 
- So that's why I mention "visit block explorer for more info".
- Let me know if it still doesn't make sense

docs/README.md
ra0x3: Same response as  [here](https://github.com/FuelLabs/fuel-indexer/pull/491#discussion_r1084091607)

deekerno (approved):

Issue #510, opened 2023-01-24
Add comments to index manifest generated by `forc-index new `


ra0x3:
- `forc index new` generates a manifest with some default parameters ([here](https://github.com/FuelLabs/fuel-indexer/blob/master/plugins/forc-index/src/utils/defaults.rs#L60))
- We should add some helpful comments for these default parameters which might be useful to new users

Additional context
- Maybe something like below, where the comments/descriptions are copied [from our book](https://fuellabs.github.io/fuel-indexer/master/components/assets/manifest.html)

```yaml
# ...

# The particular start block after which you'd like your indexer to start indexing events.
start_block: ~

# The contract_id specifies which particular contract you would like your index to subscribe to.
contract_id: ~

# ...

yash251:
please assign me this issue

ra0x3:
Hi @yash251 you don't have to wait to be assigned, feel free to put up a PR if you feel comfortable 😄  

(Being "assigned" is more so for internal Fuel contributors. External contributors don't have to follow those exact rules 👌 )

deekerno:
Just like @ra0x3 said, you don't have to be assigned through GitHub, but thanks for letting us know that you want to work on this issue! We'll make sure that no internal Fuel contributors work on this while you're working on it.

yash251:
@ra0x3 @deekerno I have added the comments along with the indentation. Let me know if I need to change anything

```rust
pub fn default_index_manifest(
    namespace: &str,
    index_name: &str,
    project_path: &str,
) -> String {
    format!(
        r#"
            # A namespace is a logical grouping of declared names. Think of the namespace as an organization identifier
            namespace: {namespace}
        
            # The identifier field is used to identify the given index.
            identifier: {index_name}
        
            # The abi option is used to provide a link to the Sway JSON ABI that is generated when you build your project.
            abi: ~
        
            # The particular start block after which you'd like your indexer to start indexing events.
            start_block: ~
        
            # The contract_id specifies which particular contract you would like your index to subscribe to.
            contract_id: ~
        
            # The graphql_schema field contains the file path that points to the GraphQL schema for the given index.
            graphql_schema: {project_path}/schema/{index_name}.schema.graphql
        
            # The module field contains a file path that points to code that will be run as an executor inside of the indexer.
            # Important: At this time, wasm is the preferred method of execution.
            module:
              wasm: ~
        
            # The report_metrics field contains boolean whether or not to report Prometheus metrics to the Fuel backend
[...]

PR #521, opened 2023-01-26
chore: add comments to example manifest generated by `forc index new`

yash251:
solves #510


ra0x3 (changes requested):
@yash251 
- Looks almost there!
- 2 things:

**1. Let's try to wrap comments at around a line length of maybe 70?**
So a line like this (1 long line):
```text
# The module field contains a file path that points to code that will be run as an executor inside of the indexer.
```
Would become (2 shorter lines):
```text
# The module field contains a file path that points to
# code that will be run as an executor inside of the indexer.
```
**2. Be mindful of indentation**

Your updated manifest content is indented, whereas the current manifest content is is not
   
Example:

What you have:
```rust
format!(
  r#"
      Your content is indented like this
  "#
);
```

What we want:
```rust
format!(
  r#"
It should not be indented, like this
  "#
);
```

Overall looks good. Changes should be easy to make 👌

yash251:
> @yash251
> 
> * Looks almost there!
> * 2 things:
> 
> **1. Let's try to wrap comments at around a line length of maybe 70?** So a line like this (1 long line):
> 
> ```
> # The module field contains a file path that points to code that will be run as an executor inside of the indexer.
> ```
> 
> Would become (2 shorter lines):
> 
> ```
> # The module field contains a file path that points to
> # code that will be run as an executor inside of the indexer.
> ```
> 
> **2. Be mindful of indentation**
> 
> Your updated manifest content is indented, whereas the current manifest content is is not
> 
> Example:
> 
> What you have:
> 
> ```rust
> format!(
>   r#"
>       Your content is indented like this
>   "#
> );
> ```
> 
> What we want:
> 
> ```rust
> format!(
>   r#"
> It should not be indented, like this
>   "#
> );
> ```
> 
> Overall looks good. Changes should be easy to make 👌

For the first thing should I keep the note thing 
```
# Important: At this time, wasm is the preferred method of execution.
```
then it becomes 3 liner

ra0x3:
@yash251

> then it becomes 3 liner

3 lines is fine :) we're more so concerned about it looking nice/neat without the long lines (rather than how many lines we have - if that makes sense).

yash251:
> @yash251
> 
> > then it becomes 3 liner
> 
> 3 lines is fine :) we're more so concerned about it looking nice/neat without the long lines (rather than how many lines we have - if that makes sense).

alrightyyyy. so please check the requested changes

yash251:
working ok

deekerno:
@yash251 Go ahead and merge the latest changes from master, and then we'll test and review one more time. 😄

yash251:
done

yash251:
done

deekerno (approved):
Thanks for this!

ra0x3 (approved):
Thanks @yash251 looks great! 🥳

PR #523, opened 2023-01-26
Update docs to include additional dependencies and clear up info regarding docker

bitzoic:
I've made the following changes:

1. Added a `wasm` section to dependencies with commands to download.
2. Updated a section on docker compose to use the file name of the file later in the documentation.
3. Moved a section on the `docker-compose.yaml` file out of a NOTE regarding local postgres instances for better visibility.
4. Added a `docker compose up` command after the creation of the `docker-compose.yaml` file.

**Please ensure that the dependencies section is correct.** I am not familiar with what is 100% required to run the quickstart but have included what is needed for `forc index check` to come back all green check marks and to match the example output in the quickstart. If any of the links, ways to install, what a dependency is used for, or description is incorrect, please let me know in the review and I can change it.


docs/README.md
```diff
@@ -57,6 +57,24 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
 - We use Docker to produce reproducible environments for users that may be concerned with installing components with large sets of dependencies (e.g. Postgres).
 - Docker can be downloaded [here](https://docs.docker.com/engine/install/).
 
+### `postgresSQL` and `SQLite`
+
+- We use PostgresSQL and SQLite to manage our database in the following quickstart.
```
ra0x3, line 62: @bitzoic 
- So Postgres and SQLite are actually not required for the Quickstart (because we use Docker), which is why we forgot to add them to this "Users" section of the README.
  - But they _should_ be in the "Users" section of the README (see comment above ☝️) - so good catch here.

docs/README.md
```diff
@@ -57,6 +57,24 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
...
+- Postgres can be downloaded [here](https://www.postgresql.org/download/)
+- SQLite can be downloaded [here](https://www.sqlite.org/download.html)
+
+### `wasm`
```
ra0x3, line 66: Great catch!

docs/README.md
```diff
@@ -246,6 +266,12 @@ services:
       - postgres
 ```
 
+Then run `docker compose up` to spin up Postgres and Fuel indexer containers:
```
ra0x3, line 269: 👍

docs/README.md
```diff
@@ -57,6 +57,24 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
 - We use Docker to produce reproducible environments for users that may be concerned with installing components with large sets of dependencies (e.g. Postgres).
 - Docker can be downloaded [here](https://docs.docker.com/engine/install/).
 
+### `postgresSQL` and `SQLite`
```
ra0x3, line 60: @bitzoic 
- I suggest we:
  - Add `PostgreSQL` and `SQlite` sub-headings under `For Users > Dependencies`.
  - Then link those sub-headings here, in `PostgreSQL` and `SQLite` sections
  - Then for now, we can just link to the download page (indexer team can follow up with more later)

Thoughts?

docs/README.md
```diff
@@ -57,6 +57,24 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
...
+
+- We use PostgresSQL and SQLite to manage our database in the following quickstart.
+- Postgres can be downloaded [here](https://www.postgresql.org/download/)
+- SQLite can be downloaded [here](https://www.sqlite.org/download.html)
+
+### `wasm`
+
+- Two additonal rust-based components required are wasm-snip and wasm32.
```
ra0x3, line 68:
```suggestion
Two additonal cargo components will be required to build your indexers: `wasm-snip` and the `wasm32-unknown-unknown` target.
```

docs/README.md
```diff
@@ -57,6 +57,24 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
...
+- Run the following command to install wasm-snip:
```
ra0x3, line 69:
```suggestion
To install `wasm-snip`:
```

docs/README.md
```diff
@@ -57,6 +57,24 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
...
+```bash
+cargo install wasm-snip
+```
+- Run the following command to install wasm32:
```
ra0x3, line 73:
```suggestion
To install the `wasm32-unknown-unknown` target via `rustup`:
```

docs/README.md
```diff
@@ -149,6 +167,8 @@ forc index check
 +--------+------------------------+------------------------------------------------------------------+
 ```
 
+If there are any missing components please install them in the [Dependencies](#dependencies) section.
```
ra0x3, line 170: So some of these components being missing is OK for the Quickstart (e.g., `psql`, `fuel-core`, etc), but some are required (e.g., `docker`, `fuelup`) - so we (indexer) should make this clearer in a follow up PR.

docs/README.md
```diff
@@ -211,11 +231,11 @@ By now we have a brand new index that will index some blocks and transactions, b
 
 #### 2.3.1 Starting an indexer service
 
-- To start an indexer service, we'll be spinning up Postgres and Fuel indexer containers via `docker compose`. Our indexer service will connect to Fuel's `beta-2` network so that we can index blocks and transactions from an _actual_ Fuel node. We'll use the `docker compose` file below, and spinning everything up with `docker compose up`.
+- To start an indexer service, we'll be spinning up Postgres and Fuel indexer containers via `docker compose`. Our indexer service will connect to Fuel's `beta-2` network so that we can index blocks and transactions from an _actual_ Fuel node. We'll use the `docker-compose.yaml` file below, and spinning everything up with `docker compose up`.
 
 > IMPORTANT: Ensure that any local Postgres instance that is running on port `5432` is stopped.
->
-> You can open up a `docker-compose.yaml` file in the same directory as your index manifest, and paste the YAML content below to this `docker-compose.yaml` file.
+
```
ra0x3, line 237: 👌

ra0x3 (changes requested):
@bitzoic 
- Left some feedback
- Some of of which we (indexer) can do in a follow up PR
- Apologies if it seems like "a lot" of feedback, we're learning that writing docs is not easy 🥲

docs/README.md
bitzoic: I will go ahead and remove this for now then but I think it would be valuable to show the output in the `2.1 Check for components` section only showing what is required rather than all components installed in that follow up PR👍

docs/README.md
bitzoic: Ah, I only now saw that this is a "For Users" section. As per my other comment, since it's specifically a quickstart and not required I'll remove this.

bitzoic:
> @bitzoic
> 
> * Left some feedback
> * Some of of which we (indexer) can do in a follow up PR
> * Apologies if it seems like "a lot" of feedback, we're learning that writing docs is not easy 🥲

Not at all too much feedback! 😆 Coming from the Sway-apps repo I've seen my fair share of feedback not to mention I'm unfamiliar with a lot of this. Please let me know if anything else needs to be changed 👍

ra0x3 (approved):
@bitzoic 

Major thanks for this 🙏 

Will defer to @deekerno .

deekerno (approved):
Looks great, thanks @bitzoic!

PR #524, opened 2023-01-26
(WIP) enhancement: Support foreign key relationships in GraphQL API queries

### LINKED ISSUE #460: Support foreign key relationships in GraphQL API queries

Additional context
- Should include tests

deekerno:
Closes #460.

## Testing Steps
New tests have been added to cover the added functionality.


deekerno:
Not ready for primetime yet as it needs a bit more work, closing for now to clear up the PR list.

PR #539, opened 2023-01-27
chore: add `build.rs` files for docs.rs

deekerno:
## Changelog
- Add `build.rs` files to database crates that set the `SQLX_OFFLINE` to true if built for [docs.rs](https://docs.rs)

## Notes
In the [logs for the failing docs.rs builds](https://docs.rs/crate/fuel-indexer/0.2.0/builds/728327), you can see that it fails due to SQLX's compile-time query checking. Since users are unable to set environment variables through `package.metadata.docs.rs`, a `build.rs` file is used to check for the [docs.rs environment](https://github.com/rust-lang/docs.rs/issues/191#issuecomment-654213450) and set the necessary environment variables, which should enable the crate documentation to build successfully.


ra0x3 (approved):
utACK :shipit:

PR #554, opened 2023-02-07
ci: add check-rustc job

### LINKED ISSUE #530: Create CI job to make sure we're always up-to-date with our `rustc` version
- As of now, we're on 1.65, but stable is at 1.67 👎 
- We need to make sure we stay up-to-date (as is the convention)

rvmelkonian:
Closes #530 
 
## Changelog 

• adds new job `check-rustc` to `ci.yml`

## Testing Plan
- The run command for the job executes as expected locally. The CI job has ran and successfully updated to `rustc` to 1.67.

## question 
- Is the rustc installation in the correct directory / environment for deployment?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ env:
   RUSTFLAGS: -D warnings
   REGISTRY: ghcr.io
   SQLX_OFFLINE: true
-  RUSTC_VERSION: 1.67
+  RUSTC_VERSION: 1.67.0
 
 jobs:
   cancel-previous-runs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,19 +50,16 @@ jobs:
     steps:
       - name: Check rustc version
         run: |
+          # Check rustc version, verify if updated 
           RUSTC_VERSION=$(rustc --version | awk '{print $2}')
           if [ "$RUSTC_VERSION" != "stable" ]; then
-            echo "Updating Rustc to the latest version: the-latest-version"
+            echo "Updating Rustc to the latest version: stable"
             curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain stable 
-          fi 
-
-      - name: Verify Rustc installation
-      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +43,20 @@ jobs:
...
 
+  check-rustc:
+    needs:
+      - cancel-previous-runs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check rustc version
+        run: |
+          # Check rustc version, install if not on latest stable
+          RUSTC_VERSION=$(rustc --version | awk '{print $2}')
```
ra0x3, line 54: @rvmelkonian 
- There's no one way to do this but here are some thoughts (I see the issue is skimpy on the details 😅 )
- Context
  - We had a situation not too long ago where we were 2 releases behind the latest stable `rustc`
  - We've only been learning about new `rustc` releases randomly (whether online or by checking the `#rust-sdk` team Slack channel)
  - Ideally we want this little CI job to just tell us whether we're on the latest `rustc` or not. A 👍  or 👎 
    - What happens when we're _not_ on the latest `rustc` is TBD - but I imagine we'd just make a PR then and there to upgrade to the latest
    - Since other jobs (e.g., `get-workspace-members`) would be dependent on this this `check-rustc` job, if `check-rustc` fails that's all we need to know/do 👌 
- What does that look like in practice?
  - There's obviously a lot of ways to do this - but maybe the simplest would be to hardcode ([here](https://github.com/FuelLabs/fuel-indexer/blob/master/.github/workflows/ci.yml#L11)) which version of Rust we (internal contributors) are on and then compare the output of what you have (`rustc --version | awk '{print $2}'`)

@deekerno Feel free to chime in here as there might be a more automated way to do this.

rvmelkonian: Ok, that makes sense! Now the job checks against the hardcoded env value. I added a regex to match the version as `rustc --version` returns a string with other info, like hash etc. 

The CI dependency flow now looks like : 
Screenshot 2023-02-08 at 12 38 18

ra0x3 (approved):
Nice. This will be super helpful, thanks for this.

PR #558, opened 2023-02-07
enhancement: forc postgres

ra0x3:
cc @SilentCicero 

### Description
- This PR adds support for a `forc postgres` plugin:
  - Updates docs
  - Adds new plugin
  - Adds new plugin as subcommand of `forc index`
  - Replaces `forc index start` command with `IndexerArgs` (they are the same thing)

### ⚠️ Notes
- **You will need to stop any local Postgres instance running**
- You might have to change the commands below from `forc index` to `forc-index`

### Testing steps

- [ ] Build the new binary `cargo build --release -p forc-index && cp target/release/forc-index ~/.fuelup/bin`
  - Note: You'll probably have to reinstall `fuelup` after testing this
- [ ] Go through the Quickstart - it should work

#### Creating a database
- `forc index postgres create postgres`
  - Will create a temp database - will not persist DB artifacts to disk -- no real point in ever using this command
- `forc index postgres create postgres --persistent`
  - Will create a temp database - will persist the DB artifacts to disk -- this is what we want to use in most use-cases

- `forc index postgres start postgres`
  - This will invoke pg_embed's pg_ctl and start the database.
    - You an `Ctrl+C` and the DB will still run
    - You can stop the DB with the stop details below 👇 

- `forc index postgres stop postgres`
  - This will kill the specified running
    - Works whether or not DB is in background

- `forc index postgres drop postgres`
  - This will drop the DB but it _will not_ remove the persisted files
- `forc index postgres drop postgres --remove-persisted`
  - This will drop the DB _and_ remove the persisted files

#### Additional UAT
- [ ] Try starting a DB that does not exist
- [ ] Try dropping a DB that is not started
- [ ] Try dropping a DB that is not stopped 
- [ ] Try dropping a DB, keeping it's persistent files, then recreating it

### Changelog
- enhancement: forc postgres


plugins/forc-postgres/src/ops/forc_postgres_createdb.rs
```diff
@@ -0,0 +1,164 @@
...
+    match pg.create_database(&name).await {
+        Ok(_) => {}
+        Err(e) => {
+            if name == DEFAULT_DATABASE {
+                info!("\nDefault database {DEFAULT_DATABASE} already exists.\n");
+            } else {
+                anyhow::bail!(e);
+            }
+        }
+    }
```
deekerno, line 142: Nit: Is it possible to use an `if let Err(e)...` here as we don't care about the `Ok` arm?

docs/README.md
```diff
@@ -156,26 +138,108 @@ forc index check
...
++--------+------------------------+---------------------------------------------------------+
+|   ✅   | forc-postgres          |  /Users/rashad/.fuelup/bin/fuelup                       |
++--------+------------------------+---------------------------------------------------------+
+```
+
+### 2.2 Database setup
+
+To quickly setup and bootstrap the PostgreSQL database that we'll need, we'll use the `forc-postgres` plugin that is included in `fuelup`.
+
+> IMPORTANT: Ensure that any local PostgreSQL instance that is running on port `5432` is stoppe
```
deekerno, line 166: s/stoppe/stopped

docs/README.md
```diff
@@ -156,26 +138,108 @@ forc index check
...
+```text
+Downloading, unpacking, and bootstrapping database.
+▸▹▹▹▹ ⏱  Setting up database...                                                                                          🎬 Starting database at 'postgres://postgres:postgres@localhost:5432/postgres'.
+💡 Creating database at 'postgres://postgres:postgres@localhost:5432/postgres'.
+waiting for server to start....2023-02-09 16:11:25.764 EST [86849] LOG:  starting PostgreSQL 14.6 on x86_64-apple-darwin20.6.0, compiled by Apple clang version 12.0.0 (clang-1200.0.32.29), 64-bit
+2023-02-09 16:11:25.766 EST [86849] LOG:  listening on IPv6 address "::1", port 5432
+2023-02-09 16:11:25.766 EST [86849] LOG:  listening on IPv4 address "127.0.0.1", port 5432
+2023-02-09 16:11:25.766 EST [86849] LOG:  listening on Unix socket "/tmp/.s.PGSQL.5432"
+2023-02-09 16:11:25.769 EST [86850] LOG:  database system was shut down at 2023-02-09 16:06:37 EST
+2023-02-09 16:11:25.772 EST [86849] LOG:  database system is ready to accept connections
```
deekerno, line 182: Can we remove some of the lower level logs from the `forc postgres` blurb? I think some of it is valuable but other parts (e.g. "received fast shutdown request") don't seem necessary and make the blocks longer.

docs/README.md
```diff
@@ -231,63 +295,28 @@ Take a quick tour.
...
-Then run `docker compose up` to spin up PostgreSQL and Fuel indexer containers:
+#### 2.4.1 Starting an indexer service
 
 ```bash
-docker compose up
+fuel-indexer \
+    --fuel-node-host node-beta-2.fuel.network \
+    --fuel-node-port 80 \
+    --postgres-password postgres
 ```
```
deekerno, line 309: 👏🏽

plugins/forc-postgres/src/ops/forc_postgres_startdb.rs
```diff
@@ -0,0 +1,51 @@
...
+    info!("\nStarting PostgreSQL.\n");
+    pg.start_db().await?;
+
+    let pg_db_uri = pg.full_db_uri(&name);
+
+    match pg.database_exists(&name).await {
+        Ok(exists) => {
+            if exists {
+                info!("✅ Successfully started database at '{pg_db_uri}'.");
+                run_forever();
```
deekerno, line 40: In the updated README, we say that a user can Ctrl-C and the database will still be running. If that's the case, do we need to run a loop and keep this particular process alive?

deekerno (changes requested):
Did a first pass, still need to do some in-depth testing.

Code looks good so far! Left a couple of corrections and questions.

docs/README.md
```diff
@@ -156,26 +138,108 @@ forc index check
...
+waiting for server to shut down.... done
+server stopped
+
+```
+
+Then we can start our database with
+
+```bash
+forc index postgres start postgres
+```
```
rvmelkonian, line 209: when testing I ran all these commands with `forc-index`

plugins/forc-postgres/src/ops/forc_postgres_startdb.rs
rvmelkonian: I would assume that `forc index postgres stop` would stop all processes, even if I didn't `Ctrl-C` `start postgres`. 

It would be great if `postgres stop` would stop all processes even if `start postgres` was still running.

Or alternatively, if `start postgres` could by itself `Ctrl-C` in the CLI, and continue running in the background. If I'm not mistaken this is similar to how `brew services start/stop postgresql` works. This parity to brew services would be neat.

rvmelkonian (commented):
Testing all works as expected. I think this is great, we probably don't want anyone to be dependant on Docker at the Quickstart stage.

docs/README.md
ra0x3: @deekerno Ah, looks like the output could be formatted not-great (thus the long lines). Will fix

docs/README.md
ra0x3: @rvmelkonian Yea when we build the plugin locally and copy it over the `fuelup`, we clobber fuleup and the commands are a bit different. But when this is released via fuelup - all commands are/will-be `forc index` (with space)

plugins/forc-postgres/src/ops/forc_postgres_startdb.rs
ra0x3: @rvmelkonian I agree. Let me play around with this a bit more.

packages/fuel-indexer-lib/src/utils.rs
```diff
@@ -203,3 +208,140 @@ pub mod index_utils {
...
+    format!("{s: ^n$}")
+}
+
+pub fn rightpad_whitespace(s: &str, n: usize) -> String {
+    format!("{s:0n$}")
+}
+
+
+// NOTE: We aren't using this now but will leave it here, as I think we'll revisit
+pub fn create_forc_postgres_database(config: &IndexerConfig) -> Result<()> {
```
ra0x3, line 260: As mentioned in the PR description, this is ultimately where we wanna go - automatically creating the database. However, there's a lot of effort involved in this - so making the tradeoff to just merge without this for now, then follow up in a separate issue.

rvmelkonian (approved):
The tests are passing and all the command are working. Did my best to try and break it, but seems solid. Only thing to note is a user has to run `brew services stop postgresql` otherwise, conflict on the same port.

rvmelkonian:
just one unused dependency on warning. Otherwise all is the same and working.

rvmelkonian (approved):

deekerno (approved):
Everything works on my end as well. Super job! 🚢

PR #563, opened 2023-02-08
enhancement: support foreign key relationships in GraphQL API queries

### LINKED ISSUE #460: Support foreign key relationships in GraphQL API queries

deekerno:
Closes #460.

## Changelog
- Add a hashmap to the `Schema` struct that allows for foreign keys to be checked at GraphQL query time
- Add method to get foreign keys inside of `load_from_db(..)` but by using already stored schema string
- Add functionality to format query results into nested JSON by using underlying database functions
- Add tests for:
-- nested queries with implicit foreign keys
-- nested queries with explicit foreign keys
-- deeply-nested queries
- Makes other GraphQL API tests more robust by ensuring that their respective handlers are not triggered by events meant for other tests

## Testing Plan
CI should pass with all the new tests that have been added. If you'd like to try your own example, you can create a new index, model the schema after the new entities added in this PR, and query for results. The returned JSON response should be properly formatted. In fact, I encourage that as we should make sure that it handles schemas other than those I thought up.

## Notes
- The PR looks big, but it's not that bad, I promise 😅 in fact, more than half of the added line count is testing and documentation
- The new `UserQuery` type (which the API queries are parsed into) included in this PR will also allow us to support other SQL as well as non-SQL storage backends, if we so choose.


packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -419,6 +424,81 @@ impl Schema {
...
+
+    let base_ast = match parse_schema::<String>(BASE_SCHEMA) {
+        Ok(ast) => ast,
+        Err(e) => {
+            panic!("Error parsing graphql schema {e:?}",)
+        }
+    };
+    let (primitives, _) = build_schema_objects_set(&base_ast);
+
+    let ast = match parse_schema::<String>(schema) {
```
rvmelkonian, line 438: maybe a comment above this line // abstract-syntax-tree

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -419,6 +424,81 @@ impl Schema {
     }
 }
 
+fn get_foreign_keys(schema: &str) -> HashMap<String, HashMap<String, String>> {
```
rvmelkonian, line 427: perhaps this should function should be a method on `Schema`?

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -419,6 +424,81 @@ impl Schema {
...
+    };
+    let (primitives, _) = build_schema_objects_set(&base_ast);
+
+    let ast = match parse_schema::<String>(schema) {
+        Ok(ast) => ast,
+        Err(e) => panic!("Error parsing graphql schema {e:?}",),
+    };
+    let types_map = build_schema_fields_and_types_map(&ast);
+
+    for def in ast.definitions.iter() {
```
rvmelkonian, line 444: feel like this logic related to definitions could also be put into it's own function. So that the broad functionality of `get_foreign_keys()` can be digested more or less at a glance. `process_defenitions()`?

packages/fuel-indexer-tests/contracts/fuel-indexer-test/src/main.sw
```diff
@@ -42,6 +42,14 @@ pub struct SimpleTupleStruct {
     data: (u32, u64, str[12]),
 }
 
+pub struct ExplicitQueryStruct {
+    id: u64
+}
```
rvmelkonian, line 47: would these ids need to be unique, or not?

rvmelkonian (commented):
Great job. This functionality is a big deal for the indexer. I've left just a few comments mostly related to readibility. 

I do wonder if some target unit tests on the `GraphQuery` methods and `get_foreign_keys`, `process_type` would be of benefit, to drive some weird edgecases into the functions, as this functionality is critical and users could be writing all kinds of bizare queries.

packages/fuel-indexer-tests/contracts/fuel-indexer-test/src/main.sw
deekerno: That probably don't need to be unique IDs between the two structs for the query tests, but I felt that it was good hygiene to make them unique so that we don't run into collisions later down the line.

packages/fuel-indexer-schema/src/db/tables.rs
deekerno: Yeah, I'm down for that, it is a bit unwieldy.

packages/fuel-indexer-database/database-types/src/lib.rs
```diff
@@ -446,3 +447,90 @@ impl IdCol {
...
+                            elements.push(")".to_string());
+
+                            if let Some(QueryElement::Field { .. }) =
+                                peekable_elements.peek()
+                            {
+                                elements.push(",".to_string());
+                            }
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
deekerno: Yeah, it's a bit of a doozy; I'll move that into a new function.

packages/fuel-indexer-database/database-types/src/lib.rs
```diff
@@ -446,3 +447,90 @@ impl IdCol {
...
+    pub filters: Vec<QueryFilter>,
+}
+
+impl UserQuery {
+    pub fn to_sql(&self, db_type: &DbType) -> String {
+        match db_type {
+            DbType::Postgres => {
+                let mut peekable_elements = self.elements.iter().peekable();
+
+                let mut elements = Vec::new();
```
ra0x3, line 481: Lots of business logic here and by reading the code it's not completely clear what's going on. Maybe some comments?

packages/fuel-indexer-schema/src/db/graphql.rs
```diff
@@ -261,43 +261,136 @@ impl Operation {
...
+                            // selection, join the foreign table on that primary key
+                            // and set the field as the innermost entity by pushing to the stack.
+                            if let Some(field_to_foreign_key) =
+                                schema.foreign_keys.get(&entity_name.to_lowercase())
+                            {
+                                if let Some(foreign_key) =
+                                    field_to_foreign_key.get(&field_name.to_lowercase())
+                                {
+                                    let reference_table =
+                                        format!("{namespace}_{identifier}.{field_name}");
```
ra0x3, line 336: Maybe we can make some of this stuff a bit more self-documenting?
```suggestion
                                        fully_qualitified_field_name(&namespace, &identifier, &field_name);
```
Where
```rust
fn fully_qualitified_field_name(namespace: &str, identifier: &str, field_name: &str) -> String {
    format!("{namespace}_{identifier}.{field_name}")
}

File: packages/fuel-indexer-schema/src/db/graphql.rs [hunk 8]
```diff
@@ -261,43 +261,136 @@ impl Operation {
...
+                                    let referencing_key =
+                                        format!("{namespace}_{identifier}.{entity_name}.{field_name}");
+                                    let primary_key =
+                                        format!("{namespace}_{identifier}.{field_name}.{foreign_key}");
+                                    let join = format!(
+                                        "INNER JOIN {reference_table} ON {referencing_key} = {primary_key}");
```
ra0x3, line 342: Similar comment here. We're doing a lot of raw-string processing - which is fine, but I think there's justification to abstract this a bit

packages/fuel-indexer-schema/src/db/tables.rs
ra0x3: I think `ast` is pretty self-explanatory.

packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml
```diff
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id: fuel1u47xjlucyjf2hkn874675fgdfue0vmyne72ucch74jcgu457rluq236j5e
+contract_id: fuel1rxsd5yxllcmfp6daw9l2l6y3fvxa083xfj9c0gw5akf670fnt63qr03w9v
```
ra0x3, line 5: 👌🏽

packages/fuel-indexer-tests/tests/e2e/graphql_api_postgres.rs
```diff
@@ -101,8 +101,165 @@ async fn test_can_return_query_response_with_nullable_fields_postgres() {
...
+    assert!(v[0]["timestamp"].as_i64().unwrap() > 0);
+    assert!(v[0]["block"]["id"].as_i64().is_some());
+    assert!(v[0]["block"]["id"].as_i64().unwrap() > 0);
+    assert!(v[0]["block"]["height"].as_i64().is_some());
+    assert!(v[0]["block"]["height"].as_i64().unwrap() > 0);
+}
+
+#[actix_web::test]
+#[cfg(all(feature = "e2e", feature = "postgres"))]
+async fn test_can_return_query_response_with_multilevel_nested_query_postgres() {
```
ra0x3, line 164: Love to see it <3

packages/fuel-indexer-schema/src/db/tables.rs
ra0x3: Agree with @rvmelkonian here

ra0x3: Also `get_foreign_keys` and `process_definitions` seem like perfect candidates for simple unit tests.

ra0x3 (changes requested):
@deekerno 
- Took a first pass
- On the whole, I think this is mostly there
- I agree with some of the feedback @rvmelkonian left
- Overall, the part that I think we should update/revisit is all the raw string logic. 
  - We don't really have comments in any of our code because, well it's self-documenting (hooray!)
  - If we could extend that logic to this PR that would be great
  - Basically asking if we can update some of this business logic to be a bit more idiomatic

docs/src/reference-guide/components/graphql/queries.md
```diff
@@ -0,0 +1,166 @@
+# Queries
```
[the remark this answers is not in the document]
deekerno: Heard that. Less lines overall. 👍🏽

deekerno:
@ra0x3 @rvmelkonian

Cleared up some of the nastier bits of the code by moving things into functions where applicable as well as adding some tests and comments; for the `definitions.iter()`, I decided to break some things out and self-document rather than turn the iterator into a function as I felt that doing so would make things a bit muddier. Let me know what you think (after the tests pass, of course).

ra0x3 (changes requested):
- So I tested this over at https://github.com/FuelLabs/fuel-indexer/pull/588
- You can use that branch to repro
- But the SQL generated seems to be invalid? Thus I get an error
- If you get that PR ☝🏽 to work on this branch I'll review again
- Also of course lmk if I'm doing something incorrect in my testing steps

deekerno:
> * So I tested this over at #588
> 
>     * You can use that branch to repro
> 
>     * But the SQL generated seems to be invalid? Thus I get an error
> 
>     * If you get that PR ☝🏽 to work on this branch I'll review again
> 
>     * Also of course lmk if I'm doing something incorrect in my testing steps

Yeah, looks like there's a missing comma somehow? Not sure how I didn't catch that in all of my manual testing, but it's a simple fix. This was on the latest commit, right @ra0x3?

deekerno:
Pulling this down while I pivot to other milestone features.

PR #569, opened 2023-02-10
chore, testing: build test WASM blobs in CI

### LINKED ISSUE #565: Create CI job to ensure `fuel_indexer_test.wasm` is up-to-date
- We use the `fuel-indexer-test` index on CI to make sure the project runs as expected
- As of now, we manually re-compile this `fuel_indexer_test.wasm` binary after making any changes
- This has the unfortunate side effect of the binary being out of date sometimes
- To ensure this doesn't happen, we should add an early CI job to ensure that the binary built on CI matches the binary in the commit

Additional context
- Could maybe just hash both and compare

deekerno:
Closes #565.

## Changelog
- Adds a step to build the test WASM blobs using the `build_test_wasm_module` script
- Removes existing fuel-indexer-test WASM blob from the repo
- Fixes asset paths in `update_test_manifest_asset_paths`
- Removes some problematic asserts in `e2e::indexing_postgres`

## Testing Plan
This should solve our CI E2E test woes. CI should pass.

## Notes
This _will_ require contributors to build test WASM blobs if they want to run the test suite locally as they won't be included in the repository; this shouldn't be too much of an ask as we typically do this already anyways.

Regarding the problematic asserts, I had done some extra work in my node fixture PR that solved these and other test weirdness. Once we merge our bigger PRs from the last few days, I'll update it and the others will be fixed.


.gitignore
```diff
@@ -10,3 +10,6 @@ docs/book/
 
 # Add this to your global .gitignore but whatever
 .DS_Store
+
+# Don't add test WASM blobs, CI generates them on their own
+*.wasm
```
ra0x3, line 15: - I think we can remove this. 
- Contributors should know to not add these files to their commit(s)
- The `.DS_Store` technically shouldn't be there either (I added that when I was having a bad day I believe).

packages/fuel-indexer-tests/src/lib.rs
```diff
@@ -47,17 +47,21 @@ pub mod utils {
     pub fn update_test_manifest_asset_paths(manifest: &mut Manifest) {
         let manifest_dir = Path::new(WORKSPACE_ROOT);
         manifest.graphql_schema = manifest_dir
-            .join("..")
-            .join("..")
+            .parent()
```
ra0x3, line 50: - Is there a reason for this change? 
- I believe they accomplish the same thing but this requires an additional `.unwrap()`?

.github/workflows/ci.yml
```diff
@@ -270,6 +270,11 @@ jobs:
       - run: cargo install sqlx-cli
       - run: bash scripts/run_migrations.bash
 
+      - name: Install wasm-snip
+        run: cargo install wasm-snip
+      - name: Build test WASM blobs
+        run: bash scripts/utils/build_test_wasm_module.bash
```
ra0x3, line 276: `build_test_wasm_module` builds 4 indexers but we only need 2 for testing (`fuel-indexer-test` and `simple-wasm`) so maybe we can just build those?

.github/workflows/ci.yml
```diff
@@ -145,7 +145,7 @@ jobs:
           command: udeps
           args: --locked --all-targets --all-features
 
-  cargo-build-wasm-example:
+  build-test-wasm-blobs:
```
ra0x3, line 148: - This job just seems to build binaries and nothing else? 
- Now that we're building them on CI for e2e tests I think we can remove this `build-test-wasm-blobs` as it will just add CI time for no additional benefit.

.github/workflows/ci.yml
```diff
@@ -189,7 +189,7 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     needs:
       - cancel-previous-runs
-      - cargo-build-wasm-example
+      - build-test-wasm-blobs
```
ra0x3, line 192: See comment above, I don't think we need `build-test-wasm-blobs` anymore

.github/workflows/ci.yml
```diff
@@ -246,8 +246,8 @@ jobs:
     if: github.event_name != 'release' && github.event.action != 'published'
     needs:
       - cancel-previous-runs
-      - cargo-build-wasm-example
       - get-workspace-members
+      - build-test-wasm-blobs
```
ra0x3, line 250: Same here. Not needed I don't believe

.github/workflows/ci.yml
```diff
@@ -295,6 +300,7 @@ jobs:
     needs:
       - cancel-previous-runs
       - cargo-toml-fmt-check
+      - check-rustc
```
ra0x3, line 303: 👍

ra0x3 (changes requested):

packages/fuel-indexer-tests/src/lib.rs
deekerno: So the `.join()`s weren't actually referring to the parent in practice; it was just building a string that wasn't resolved correctly. See the error message here -- https://github.com/FuelLabs/fuel-indexer/actions/runs/4146742221/jobs/7172823151:

```
---- e2e::indexing_postgres::test_index_metadata_is_saved_when_indexer_macro_is_called_postgres stdout ----
thread 'e2e::indexing_postgres::test_index_metadata_is_saved_when_indexer_macro_is_called_postgres' panicked at 'Failed to initialize indexer.: IoError(Custom { kind: NotFound, error: VerboseError { source: Os { code: 2, kind: NotFound, message: "No such file or directory" }, message: "could not open `/home/runner/work/fuel-indexer/fuel-indexer/packages/fuel-indexer-tests/../../packages/fuel-indexer-tests/assets/fuel_indexer_test.wasm`" } })', packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs:514:10
```

.github/workflows/ci.yml
deekerno: Oops, that's fair. This was left in when I assumed that files were persisted across each check.

packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml
```diff
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id: fuel1mm5dud7hsw2u0nj0n60yr20ag4gewezpg7au9twgdgpupcup6ajsge77vg
+contract_id: ~
```
ra0x3, line 5: @deekerno Let's either add this back or pull in changes from https://github.com/FuelLabs/fuel-indexer/pull/570 and take the incoming change from the merge conflict?

Other than that PR is fine and will ✅ when CI passes

ra0x3 (approved):

PR #574, opened 2023-02-13
enhancement: add default body limit to post routes

### LINKED ISSUE #448: Add `DefaultBodyLimit` to web API endpoint(s)
- So as to not entice some user to do malicious things, we should set some max size for web requests
- Doesn't have to be anything crazy, maybe just the size of a large WASM blob
- https://dev.materialize.com/api/rust/axum/extract/struct.DefaultBodyLimit.html

rvmelkonian:
Closes #448 

## Changelog 
Adds `DefaultBodyLimit::max(10000)` to post routes.

## Testing 
We could add an intergration test for this, but that would mean adding a large Wasm file to the repo. I could also smoke test manually using Postman. If postman (or something similar) is used to sanity test the api, is there a collection that already exists?

### DIFF
--- a/packages/fuel-indexer-api-server/src/api.rs
+++ b/packages/fuel-indexer-api-server/src/api.rs
@@ -123,6 +123,7 @@ impl GraphQlApi {
 
         let graph_route = Router::new()
             .route("/:namespace/:identifier", post(query_graph))
+            .layer(DefaultBodyLimit::max(10000))
             .layer(Extension(schema_manager.clone()))
             .layer(Extension(pool.clone()));
 
--- a/packages/fuel-indexer-api-server/src/api.rs
+++ b/packages/fuel-indexer-api-server/src/api.rs
@@ -4,7 +4,7 @@ use crate::uses::{
 };
 use async_std::sync::{Arc, RwLock};
 use axum::{
-    extract::{Extension, Json},
+    extract::{DefaultBodyLimit, Extension, Json},
     http::StatusCode,
     middleware,
     response::{IntoResponse, Response},
@@ -23,6 +23,7 @@ use thiserror::Error;
 use tokio::sync::mpsc::{error::SendError, Sender};
 use tower_http::{
     cors::{Any, CorsLayer},
+    limit::RequestBodyLimitLayer,
     trace::{DefaultMakeSpan, DefaultOnRequest, DefaultOnResponse, TraceLayer},
     LatencyUnit,
 };
@@ -127,6 +128,7 @@ impl GraphQlApi {
 
         let index_routes = Router::new()
             .route("/:namespace/:identifier", post(register_index_assets))
+            .layer(DefaultBodyLimit::max(10000))
             .route_layer(middleware::from_fn(authorize_middleware))
             .layer(Extension(tx.clone()))
             .layer(Extension(schema_manager))


rvmelkonian:
@deekerno thanks for this. I'll test with the same curl command, perhaps I'm missing something.

ra0x3 (changes requested):
Looks ok. 2 things:
1. Can we make the max size configurable via a CLI option / config file and use a sensible default?
  - This would be similar to an option like `stop_idle_indexers`
  - Might require docs updates as well
2. We'll definitely want a larger default body. Maybe something like 5MB to start? @deekerno Thoughts on the size?

deekerno:
> @deekerno thanks for this. I'll test with the same curl command, perhaps I'm missing something.

@rvmelkonian, it looks like you may also need to use [`RequestBodyLimit::disable()`](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html#method.disable) to allow for bigger sizes as @ra0x3 suggested; as far as size goes, I think 5MB is a good size to start with.

rvmelkonian:
When I hit the api with this command
```
curl -v http://0.0.0.0:29987/api/index/fuel_examples/fuel_examples \
    -F "manifest=@packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml" \
    -F "wasm=@packages/fuel-indexer-tests/assets/bad_simple_wasm.wasm" \
    -F "schema=@packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql" \
    -H 'Content-type: multipart/form-data' -H "Authorization: foo" | json_pp
```
I see this again
```
thread 'tokio-runtime-worker' panicked at 'Failed to spawn executor from index asset registry: FFIError(Export(Missing("get_identifier_ptr")))', /Users/rvm/fuel/fuel-indexer/packages/fuel-indexer/src/service.rs:254:34
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(15), ...)', /Users/rvm/fuel/fuel-indexer/packages/fuel-indexer/src/service.rs:203:10
```
I've ran the scripts:
```
bash scripts/utils/build_test_wasm_module.bash
```
and 
```
bash scripts/utils/refresh_test_db.bash
```
If your able to reproduce this @deekerno, then I know its the code, (although I doubt it) otherwise it must be environment related

rvmelkonian:
Switching to use tower's `RequestBodyLimitLayer` did the trick. I believe `DefaultBodyLimit` was not working because 
DefaultBodyLimit is local in that it only applies to `FromRequest` https://docs.rs/axum/0.6.2/axum/extract/trait.FromRequest.html) implementations that explicitly apply it.

Also `RequestBodyLimitLayer` is recommended in the axum docs if accepting data from untrusted remotes.

packages/fuel-indexer-lib/src/config/mod.rs
```diff
@@ -101,6 +101,10 @@ pub struct IndexerArgs {
     #[clap(long, help = "Database type.", default_value = defaults::DATABASE, value_parser(["postgres"]))]
     pub database: String,
 
+    /// Max body size for the GraphQL API.
```
ra0x3, line 104:
```suggestion
    /// Max body size for WASM binary uploads.
```

packages/fuel-indexer-lib/src/defaults.rs
```diff
@@ -14,6 +14,7 @@ pub const POSTGRES_PASSWORD: &str = "postgres";
 pub const INDEX_FAILED_CALLS: usize = 10;
 
 pub const GRAPHQL_API_RUN_MIGRATIONS: bool = false;
+pub const MAX_BODY_LIMIT: &str = "10"; // 5MB in bytes
```
ra0x3, line 17: - So I see [`new`](https://docs.rs/tower-http/latest/tower_http/limit/struct.RequestBodyLimitLayer.html#method.new) takes a `usize` but here we're making `MAX_BODY_LIMIT` a `String`. 
- I'm assuming `clap` doesn't allow `usize` values in command fields?
- I also see we're assigning it to a value of `"10"` ? I would expect the value (even if a string) to be what you added in `fuel-indexer-test/lib/defaults` i.e., `pub const MAX_BODY_LIMIT: usize = 5242880;`

ra0x3 (commented):
Left some questions.

packages/fuel-indexer-lib/src/defaults.rs
rvmelkonian: My bad, left this in for testing. Will fix. I can attempt to coeerce clap with `usize`, I'm sure its possible, that will be cleaner yes.

rvmelkonian: So clap does allow `usize` type in command fields, but it requires the type `&str` to set the `default_value` in the `#[clap()]` attribute. For that reason I've left it as a string, with the conversion to `usize` happening on 302. Let me know if you prefer a different way.

deekerno (dismissed):
```
❯ curl -v http://0.0.0.0:29987/api/index/fuel_examples/fuel_examples \
    -F "manifest=@packages/fuel-indexer-tests/assets/fuel_indexer_test.yaml" \
    -F "wasm=@packages/fuel-indexer-tests/assets/bad_simple_wasm.wasm" \
    -F "schema=@packages/fuel-indexer-tests/assets/fuel_indexer_test.graphql" \
    -H 'Content-type: multipart/form-data' -H "Authorization: foo" | json_pp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 0.0.0.0:29987...
* Connected to 0.0.0.0 (127.0.0.1) port 29987 (#0)
> POST /api/index/fuel_examples/fuel_examples HTTP/1.1
> Host: 0.0.0.0:29987
> User-Agent: curl/7.79.1
> Accept: */*
> Authorization: foo
> Content-Length: 24937313
> Content-Type: multipart/form-data; boundary=------------------------d5a97407e77e2d34
> Expect: 100-continue
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 413 Payload Too Large
```
Nice!

packages/fuel-indexer-api-server/src/api.rs
```diff
@@ -113,6 +114,7 @@ impl GraphQlApi {
         let sm = SchemaManager::new(pool.clone());
         let schema_manager = Arc::new(RwLock::new(sm));
         let config = config.clone();
+        let max_body_limit = config.max_body_limit;
```
ra0x3, line 117: Nit: I do think the naming is interesting. It effectively has 2 nouns that imply the same thing "max" and "limit". Not a blocker, can always update later.

rvmelkonian: true, I agree. Let me update all occurences to `max_body`

ra0x3 (changes requested):
@rvmelkonian  Last tiny change
- With this new flag we should update the docs to include this new flag.
- The flag should be added to `starting-the-fuel-indexer.md` in `docs/` 
- Would be great to do something like `forc` and have that stuff auto-generated, but not there yet 🥲 
- Then ✅

ra0x3 (approved):

PR #577, opened 2023-02-13
enhancement: forc index relative paths

ra0x3:
### Description
- This PR defaults `forc index` to using relative paths for assets
- Also adds an `--absolute-paths` flag to `init` and `new` so the user can use absolute paths if they'd like

### Testing steps
- [ ] Rebuild the plugin  `cargo build --release -p forc-index && cp target/release/forc-index ~/.fuelup/bin`
- [ ] Test default behavior `forc-index new foo --namespace bar && cd foo && cat/foo.manifest.yaml`
- [ ] Test absolute path behavior `forc-index new foo --namespace bar --absolute-paths && cd foo && cat foo.manifest.yaml`

### Changelog 
- enhancement: forc index relative paths

### DIFF
--- a/plugins/forc-index/src/commands/init.rs
+++ b/plugins/forc-index/src/commands/init.rs
@@ -19,16 +19,17 @@ pub struct Command {
     )]
     pub path: Option<PathBuf>,
 
-    /// Name of the index namespace
+    /// Namespace in which index belongs.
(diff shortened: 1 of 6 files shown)


rvmelkonian (approved):
This all works. In the example test commands above there is a typo `cat/foo.manifest.yaml` should be `cat foo.manifest.yaml`

PR #581, opened 2023-02-15
changes to fix wasm support

segfault-magnet:
Currently points to the feature branch of the rust SDK. 

Also I didn't run the whole test suite due to Postgress, but the command given when the issue was raised is now passing. 

The fuel-* deps needed bumping because the SDK is reexporting parts of the fuel-* crates (especially fuel-types) and if you have a mismatch in versions between the indexer and the SDK with regards to fuel-* crates then types coming from those dependencies, although the same, will be considered different because the ones from the SDK are coming from another version of the same crate.


deekerno:
The WASM stuff builds with a couple of changes, so I'm merging this into my feature branch where I'll finish up the upgrade work. Thank you so much, @segfault-magnet! ❤️

PR #586, opened 2023-02-16
chore: add test for querying and registering multiple indicies 

### LINKED ISSUE #455: Add test for multiple indices
- In pretty much all tests and local dev work we only use a single index
- We should add tests to ensure that all functionality works, even with several indices
- This could be as simple as just writing a single test with several registered indices, triggering some events, and then expecting some outputs
  - We have plenty of tests that display this functionality already
  - The indices wouldn't need to be anything complex, just receiving and indexing simple events

rvmelkonian:
Closes #455

This PR adds a new test which creates three instances of the test manifest file, updating the namespace and identifier, registering those indexes in the db and asserting the blockheights of each respectively. 

## Changelog
- adds `test_can_trigger_and_index_events_with_multiple_fuel_indexes`


rvmelkonian:
Update: the fuel SDK upgrade changes @deekerno is working on will fix this error on the tests.

rvmelkonian:
@deekerno 

I've merged in master and made use of the new `setup_example_test_fuel_node()` (thanks for this!), however now I run the test I see these errors

```
Feb 24 20:50:43.730 ERROR fuel_indexer::database: TypeId(1406149282443221513) not found in tables: {}
Feb 24 20:50:43.734 ERROR fuel_indexer::database: TypeId(-6012155513071029430) not found in tables: {}
Feb 24 20:50:43.738 ERROR fuel_indexer::database: TypeId(-6012155513071029430) not found in tables: {}
Feb 24 20:50:43.740 ERROR fuel_indexer::database: TypeId(1406149282443221513) not found in tables: {}
Feb 24 20:50:43.741 ERROR fuel_indexer::database: TypeId(-6012155513071029430) not found in tables: {}
Feb 24 20:50:43.743 ERROR fuel_indexer::database: TypeId(-6012155513071029430) not found in tables: {}
Feb 24 20:50:43.744 ERROR fuel_indexer::database: TypeId(-1599046217856387010) not found in tables: {}
```
I think this is stopping the indexer from beginning to index these indicies, when we do the SELECT, there are no rows being returns; the schemas are being successfully created in the DB.  

I looked at the other tests but can't seem to see any obvious things I'm missing. Do I need to create some new util function to create multiple contracts perhaps?

deekerno:
@rvmelkonian: Yeah, the example test node function only creates one contract, which currently uses a hardcoded `CONTRACT_ID` constant as we're currently unable to spawn the contract deployment in a way that gives us the contract ID _and_ a `JoinHandle` that we can use to kill the test node at test completion (or at least, I haven't figured it out). So yeah, it'd probably be best to make a separate function that spawns multiple contracts; you could probably do it in the test itself as we'd only be writing one test for multiple contracts.

You can find information for using the SDK for multiple contract deployment [here](https://fuellabs.github.io/fuels-rs/v0.37.1/abigen/the-abigen-macro.html).

rvmelkonian:
currently blocked when trying to deploy multiple instances of the same contract. I am doing what's recommended from the docs: 
```rust
       for contract in contracts {
        let rng = &mut StdRng::seed_from_u64(2322u64);
        let salt: [u8; 32] = rng.gen();

        let contract_id = Contract::deploy_with_parameters(
            contract.bin_path.as_os_str().to_str().unwrap(),
            &wallet,
            tx_params(),
            StorageConfiguration::default(),
            Configurables::default(),
            Salt::from(salt),
        )
        .await
        .expect("Failed to deploy contract");

        let contract_id = contract_id.to_string();
        println!("Contract deployed at: {}", &contract_id);
    }
```
the compiler complains:
```
error[E0308]: mismatched types
   --> packages/fuel-indexer-tests/src/fixtures.rs:153:13
    |
153 |         let salt: [u8; 32] = rng.gen();
    |             ^^^^  -------- expected due to this
    |             |
    |             expected array `[u8; 32]`, found struct `fuels::client::schema::schema::__fields::Contract::salt`
    ```
    
  I've tried to get around this several ways, for now will wait to hear back from fuels team.

rvmelkonian:
@deekerno you'll see I've updated `setup_test_fuel_node()` to take multiple contracts, added a `generate_contracts` fn, this gives us ids before deployment, and also created `connect_to_deployed_contracts()` 

the code is successfully spawning multiple contracts, but it seems like when `Contract::deploy_with_parameters` is called this triggers an event which throws an error in the indexer. 

I think I have all the pieces there, but some logic is still wrong and I've been trying to debug it for a while. If you could point me in the right direction, would be appreciated!

PR #591, opened 2023-02-16
v0.3.0

ra0x3:


ra0x3:
@deekerno 
- I believe this is our first minor version bump since we've had the plugin (thus all the default files)
- Will need a keen eye here

rvmelkonian (approved):
Did a grep for `0.2` seems like all is updated as we need.

PR #592, opened 2023-02-16
ci: add check-is-release-branch step

ra0x3:
### Description
- Add a check on CI to no build the plugin if it's a release branch
- You can see that this step works in the positive case, on this test branch's action https://github.com/FuelLabs/fuel-indexer/actions/runs/4204895463/jobs/7296222119
- You can see that this step works in the negative case, on this PR


deekerno (approved):
🚢

PR #595, opened 2023-02-17
enhancement: add resume index functionality

### LINKED ISSUE #568: Add "resume indexer" functionality
- As an indexer executor runs, it keeps track of block height during every call to `handle_events`
  - [Here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-macros/src/indexer.rs#L637)
- This allows the user to stop a running indexer, and sort've keep track of where their indexer last stopped
- We also allow users to set a `start_block` in their indexer manifest - where this `start_block` determines at which block height the user's indexer will start indexing 
  - [Here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer/src/executor.rs#L81)
- However, there's currently no way for an indexer to "pick up where it left off" automatically
- We should add functionality that allows an indexer to "pick up where it left off" by using the `IndexMetadata`
- When the indexer is created, it would either start at `start_block` or `pick_up_where_you_left_off_block`, depending on what is specified in the indexer manifest
  - Maybe `pick_up_where_you_left_off_block` should take precedence over `start_block`?

Additional context
- This might require some new parameter/flag in the indexer manifest

rvmelkonian:
Closes #568 

## Changelog
• adds `resumable` to manifest
• creates a new queries:: method `last_block_height_for_indexer`

## Testing 
- Run 
`cargo run --bin fuel-node`
`cargo run --bin web-api`
Update `fuel_indexer_test.yaml` to be resumable by adding the field:
```
resumable: true
```
- Run the test indexer
`cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml`

-Fire off some events (more than 2)
`curl -v -X POST http://127.0.0.1:8000/ping`

- Quit the indexer process. 
- Start it again with the same command above. Search the logs for `Last block`. It should be more than 1. 

I opted not to write an e2e test here as I wasn't sure yet how to actually quit the indexer and then start it again whilst the api and fuel node were still up. I could figure it out if we still want this.


packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -554,6 +554,29 @@ pub async fn latest_assets_for_index(
...
+    METRICS
+        .db
+        .postgres
+        .last_block_height_for_indexer_calls
+        .inc();
+
+    let query = format!(
+        "SELECT MAX(height) FROM {}.block",
+        format_args!("fuel_indexer_{}_{}", namespace, identifier)
+    );
```
deekerno, line 572: It's definitely not easy to find, but we have an `IndexMetadata` entity and backing database table that is created for each index; in the codegen, a final call is made to save the block height to said table. You can see that call [here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-macros/src/indexer.rs#L639-L640) and an example using the database table [here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs#L591-L594).

packages/fuel-indexer/src/executor.rs
```diff
@@ -74,18 +75,16 @@ impl ExecutorSource {
 pub fn run_executor<T: 'static + Executor + Send + Sync>(
     fuel_node_addr: &str,
     mut executor: T,
-    start_block: Option<u64>,
+    start_block: &u64,
     kill_switch: Arc<AtomicBool>,
     stop_idle_indexers: bool,
 ) -> impl Future<Output = ()> {
-    let start_block_value = start_block.unwrap_or(1);
```
deekerno, line 81: Removing this will cause a panic if a number is not used for the corresponding field in the manifest due to the `.unwrap()` call in `get_start_block()`.

packages/fuel-indexer/src/service.rs
```diff
@@ -286,3 +298,19 @@ async fn create_service_task(
...
+        let start_block = queries::last_block_height_for_indexer(
+            conn,
+            &manifest.namespace,
+            &manifest.identifier,
+        )
+        .await
+        .unwrap_or(0) as u64;
+        start_block
+    } else {
+        manifest.start_block.unwrap()
```
deekerno, line 314: This will cause the service to panic and crash if the manifest does not have a number for the `start_block` field. Up to now, we've allowed the manifest have the field undefined or assigned a dummy value of `~` by convention.

deekerno (changes requested):
I'm not sure that this is working correctly. I tested it with the block explorer example and the index still seems to be starting at the first block. Were you able to get it working with the explorer example?

Also, is it possible to add an end-to-end test for this?

packages/fuel-indexer-types/src/graphql.rs
```diff
@@ -14,10 +14,10 @@ impl GraphqlObject for IndexMetadata {
     fn schema_fragment() -> &'static str {
         r#"
 
-type IndexMetadataEntity {
-    id: ID!
-    time: Int8!
-}
-"#
+    type IndexMetadataEntity {
```
ra0x3, line 17: Looks like this was indented unintentionally?

packages/fuel-indexer-database/src/queries.rs
```diff
@@ -248,6 +248,18 @@ pub async fn latest_assets_for_index(
     }
 }
 
+pub async fn last_block_height_for_indexer(
```
ra0x3, line 251: 👌🏽

ra0x3 (changes requested):
@rvmelkonian 
- PR is definitely on the right track 🎖️ 
- Some minor comments from me in addition to @deekerno's feedback
- Since we don't include `packages/fuel-indexer-tests/assets/fuel_indexer_test.wasm` in the repo anymore we can remove that

rvmelkonian:
Thanks for the feedback all, yes I'll add an end-to-end test in this PR and a spend a bit more time on it. Learned lots about the codebase by working on this!

packages/fuel-indexer/src/service.rs
```diff
@@ -286,3 +299,23 @@ async fn create_service_task(
         }
     }
 }
+
+async fn get_start_block(
+    conn: &mut IndexerConnection,
+    manifest: &Manifest,
+) -> Option<u64> {
+    let resumable = manifest.resumable.unwrap_or(false);
```
ra0x3, line 307: Maybe 


```rust
match manfiest.resumable {
  Some(s) => {
     let last = queries::last_block_height_for_indexer().await?;  
     last
  }
  None => manifest.start_block.unwrap_or(1)
```

packages/fuel-indexer/src/service.rs
```diff
@@ -286,3 +299,23 @@ async fn create_service_task(
         }
     }
 }
+
+async fn get_start_block(
+    conn: &mut IndexerConnection,
+    manifest: &Manifest,
+) -> Option<u64> {
```
ra0x3, line 306: In all cases, `get_start_block` seems to return a `u64`? So maybe we can just make `u64` the return type as opposed to `Option<u64>`?

packages/fuel-indexer/src/service.rs
```diff
@@ -286,3 +299,23 @@ async fn create_service_task(
...
+    let resumable = manifest.resumable.unwrap_or(false);
+    if resumable {
+        let last_block = queries::last_block_height_for_indexer(
+            conn,
+            &manifest.namespace,
+            &manifest.identifier,
+        )
+        .await
+        .unwrap_or(1) as u64;
+        println!("Last block: {}", last_block);
```
ra0x3, line 316: - If there's info we want the user aware of we can use `logger::info`
  - But in this particular case I think we can remove this `println!` as it isn't too informative

packages/fuel-indexer/src/service.rs
```diff
@@ -243,12 +252,16 @@ async fn create_service_task(
                                     serde_yaml::from_slice(&assets.manifest.bytes)
                                         .expect("Failed to deserialize manifest");
 
+                                let start_block = get_start_block(&mut conn, &manifest)
```
ra0x3, line 255: The default case is handled in `get_start_block` already so we can just 

```rust
let start_block = get_start_block(&mut conn, &manifest).await?;
```

packages/fuel-indexer/src/service.rs
```diff
@@ -286,3 +299,23 @@ async fn create_service_task(
...
+) -> Option<u64> {
+    let resumable = manifest.resumable.unwrap_or(false);
+    if resumable {
+        let last_block = queries::last_block_height_for_indexer(
+            conn,
+            &manifest.namespace,
+            &manifest.identifier,
+        )
+        .await
+        .unwrap_or(1) as u64;
```
ra0x3, line 315: If `last_block_height_for_indexer` fails (for whatever reason) that's a legit error, and we should unwrap it as opposed to swallowing it with `.unwrap_or(1)` so maybe we can just 

```rust
        let last_block = queries::last_block_height_for_indexer(
            conn,
            &manifest.namespace,
            &manifest.identifier,
        )
        .await?;
```

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -554,6 +554,29 @@ pub async fn latest_assets_for_index(
...
+    identifier: &str,
+) -> sqlx::Result<i64> {
+    #[cfg(feature = "metrics")]
+    METRICS
+        .db
+        .postgres
+        .last_block_height_for_indexer_calls
+        .inc();
+
+    let query = format!(
```
ra0x3, line 569:
```suggestion
    let query = format!("SELECT MAX(id) FROM {namespace}_{identifier}.indexmetadataentity LIMIT 1");
```

packages/fuel-indexer/src/executor.rs
```diff
@@ -74,18 +75,16 @@ impl ExecutorSource {
...
-    let mut next_cursor = if start_block_value > 1 {
-        let decremented = start_block_value - 1;
+    let mut next_cursor = if *start_block > 1 {
```
ra0x3, line 82: Hm not too sure what this logic is doing, but I think `get_start_block` should return a value that doesn't have to be changed/altered by `run_executor` anymore 🤔

ra0x3 (commented):
@rvmelkonian 
- Left some style/Rust-y comments
- I'm sure those errors you're seeing might be related to your `start_block` changes

rvmelkonian:
@ra0x3 turns out the reason why tests were failing was because I had `resumable : true` in the test manifest, this caused `test_index_respects_start_block_postgres` and other tests to fail as resumable takes precedence.

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -554,6 +554,31 @@ pub async fn latest_assets_for_index(
...
+        "SELECT MAX(id) FROM {namespace}_{identifier}.indexmetadataentity LIMIT 1"
+    );
+
+    match sqlx::query(&query).fetch_one(conn).await {
+        Ok(row) => {
+            let id: i64 = row.get(0);
+            Ok(id as u64)
+        }
+        Err(_) => Ok(0),
```
ra0x3, line 578: Block heights will never be `0`. If we're keeping this logic we should make this `1`

ra0x3 (changes requested):

ra0x3 (changes requested):
@rvmelkonian 
- There is still a .DS_Store file in this PR
- I don't see any log in this PR that says "Resuming indexer from ___". I could be looking in the wrong place. Could we add one?
  - I imagine this might go in `get_start_block`
- You also need to:
  - Add this new flag and its description to `manifest.md` in the book
  - Update all manifests (examples, tests) to include this new resumable flag (probably with an empty `~` value)
  - Add this new manifest flag to the default forc-index manifest file in forc-index/defaults

ra0x3 (approved):

deekerno (approved):
Works as per the testing steps.

Great job! 🎉

PR #576, opened 2023-02-13
chore: create Fuel node fixture

deekerno:
## Changelog
- Add node fixture to `fuel-indexer-tests`
- Adjust tests to spin up and tear down node fixture for each test
- Remove `fuel-node` package from workspace
- Remove `fuel-node` build step from CI
- Add binary using test node fixture to deploy contract for `hello-world` example

## Testing Steps
CI should pass with new Fuel node fixture. Additionally, the "Hello World" example has been adjusted; run through both the example README and the example in the documentation, and verify that everything works.

## Notes
- No big improvement in CI runtime; however, it does allow for all of our testing to be done through `cargo` actions and for us not to use extra scripts.
- A number of tests needed to be adjusted to either remove asserts that weren't properly testing values or to delete records from tables at test invocation to ensure a clean slate.

Issue #386, opened 2022-12-02
Swap `fuel_indexer_types::abi::Identity` for `fuels_core::Identity`


ra0x3:
- With the completion of https://github.com/FuelLabs/fuel-indexer/issues/296 we will have support for `Identity` types
- However, the SDK (at the time) did not readily have an `Identity` type that would work for the Indexer -- and we can't wait for them to support it
- So we created our own `Identity` that does everything the SDK `Identity` does, plus a little more
- This works fine but requires the user to convert `Identity` types from Sway contracts to `Identity` types in indexer modules using `Identity::from(my_event.identity_field)`
- This if fine, but a hack. Ideally we should be using `fuel_core::Identity`

ra0x3:
Blocked by https://github.com/FuelLabs/fuels-rs/issues/717

deekerno:
Blocked by #423.

PR #598, opened 2023-02-17
chore: upgrade `fuels-rs` to v0.36

### LINKED ISSUE #423: Upgrade `fuels`/`fuels-types` to version without `mio` in dependency graph
Blocks #421, #386, #300.

Upon trying to upgrade dependencies in #421, we started to get build errors when trying to build any of the WASM examples. Upon further investigation, it seems that `fuels-types` has included `fuel-chain-config` in its dependencies which in turn depends on `fuel-core-interfaces`. `fuel-core-interfaces` lists `tokio` with the `process` and `signal` features as a dependency, which brings in `mio` in both cases. The inclusion of `mio` is the actual cause of this issue as it does not support being built for WASM.

As such, we need to discuss potential solutions with other teams to mitigate the inclusion of `mio` since writing an index requires `fuels-core`/`fuels-types` in order to be able to work with Fuel types.

deekerno:
Closes #423, #386.

## Changelog
Upgrades:
| Dependency | Current Version | New Version | crates.io Link |
| --------------- | --------------- | --------- | ---------- |
| fuel-tx | 0.23 | 0.26 | https://crates.io/crates/fuel-tx |
| fuel-core-client | 0.15 | 0.17.2 | https://crates.io/crates/fuel-core-client |
| fuel-types | 0.5 | 0.26 | https://crates.io/crates/fuel-types |
| fuels | 0.30 | 0.36 | https://crates.io/crates/fuels |
| fuels-types | 0.30 | 0.36 | https://crates.io/crates/fuels-types |
| fuels-core | 0.30 | 0.36 | https://crates.io/crates/fuels-core |
| fuels-macros | 0.30 | 0.36 | https://crates.io/crates/fuels-macros |
| forc-util | 0.31 | 0.35 | https://crates.io/crates/forc-util |

New dependencies:
| Dependency | Current Version | New Version | crates.io Link |
| --------------- | --------------- | --------- | ---------- |
| fuels-code-gen | - | 0.36 | https://crates.io/crates/fuel-tx |
| fuel-abi-types | - | 0.2.1 | https://crates.io/crates/fuel-core-client |

Notes:
- `fuel-gql-client` is now known as `fuel-core-client`
- `fuels-abigen-macro` has been superseded by `fuels-macros`

## Testing Plan
CI should pass. Run through each example and ensure that they work as described.

Note that the "hello world" examples have been adjusted due to changes in the SDK. To run a fuel node with the requisite contract deployed, use `cargo run --bin fuel-node -- --contract-bin examples/hello-world/contracts/greeting/out/debug/greeting.bin` and then `cargo run --bin hello-world-data -- --host 0.0.0.0:4000` to generate data for the example.

## Notes
- This PR changes the `fuels` dependencies to use @segfault-magnet's WASM hotfix branch. These dependencies will need to be changed once the branch has been merged into the main branch of `fuels-rs`; I'll create and link a tracking issue for said event.
- This PR uses the test fixture from #576 along with other adjustments in order to properly work with deployed contracts.


examples/block-explorer/explorer-index/Cargo.toml
```diff
@@ -11,9 +11,10 @@ crate-type = ['cdylib']
 fuel-indexer-macros = { version = "0.3", path = "../../../packages/fuel-indexer-macros", default-features = false }
 fuel-indexer-plugin = { version = "0.3", path = "../../../packages/fuel-indexer-plugin" }
 fuel-indexer-schema = { version = "0.3", path = "../../../packages/fuel-indexer-schema", default-features = false }
-fuel-tx = "0.23"
-fuels-core = "0.30"
-fuels-types = "0.30"
+fuel-tx = "0.26"
+fuels-core = { git = "https://github.com/FuelLabs/fuels-rs", branch = "segfault_magnet/wasm_friendly_abigen", default-features = false }
```
ra0x3, line 15: @deekerno  
- Do we have any idea how long we'll have to use this branch? 
- Is there some target date by which we can expect to be back on master?

cc @segfault-magnet

ra0x3 (changes requested):
@deekerno 
- As mentioned on slack. A lot of this functionality doesn't work
- [x] Block explorer ✅ 
- [x] Hello-world ❌ 
- [ ] Hello-world-native ❌ 
- [x] Hello-world-data ❌  (I imagine this is the cause of the breakage with hello-world, hello-world-native)
- [x] web-api fixture ❌ 

I'll test the dockerized versions after these ☝🏽 work


Other than that PR looks surprisingly simple.

examples/block-explorer/explorer-index/Cargo.toml
deekerno: I haven't received an answer about how long we'll need to use the branch. But I'll be monitoring it so that we can switch as soon as it goes into master, and then as soon as it's released.

deekerno:
> @deekerno
> 
>     * As mentioned on slack. A lot of this functionality doesn't work

I know what the problem is; at the end of the day, I had `git checkout`'d some of the changes from #576 as the SDK made some changes with the underlying `fuel-core` contract deployment bits, but I didn't pull in all of the dependent changes. Should be an easy fix.

examples/block-explorer/explorer-index/Cargo.toml
ra0x3: @deekerno Could we create a tracking issue for that?

packages/fuel-indexer-tests/src/lib.rs
```diff
@@ -29,6 +29,12 @@ pub mod defaults {
     pub const WALLET_PASSWORD: &str = "password";
     pub const INDEXED_EVENT_WAIT: u64 = 2;
     pub const COIN_AMOUNT: u64 = 11;
+
+    // Note: We need the contract ID in order to connect an instance to
+    // the node provider for E2E tests, but spawning a task that is supposed
+    // to return the contract ID won't give us the ID until the task is completed.
+    pub const CURRENT_TEST_CONTRACT_ID_STR: &str =
```
ra0x3, line 36: @deekerno 
- Is there any reason we're hardcoding this in a second place vs just connecting to the contract to get the contract ID? 
- I'm assuming there was some change in the dependencies now that prevents us from connecting to the contract like we were?

examples/block-explorer/explorer-index/Cargo.toml
deekerno: Done here: https://github.com/FuelLabs/fuel-indexer/issues/605

packages/fuel-indexer-tests/src/lib.rs
deekerno: Yes, you're correct in your assumption. Previously, we were using `get_contract_id` which was actually redeploying the contract under-the-hood. However, the underlying `fuel-core` dependency does not allow that anymore; instead, it returns a `ContractAlreadyDeployed` error. I attempted to work around this by spawning the fixture and then returning the ID for the deployed contract and the join handle for the node fixture; however, the only way that I would be able to get the ID from `Contract::deploy(..)` from a JoinHandle is if the task completed fully, which means that it wouldn't be running for the tests.

.github/workflows/ci.yml
```diff
@@ -293,37 +293,37 @@ jobs:
...
-
-      - name: Create new index
-        run: |
-          mkdir ../forc_index_test && cd ../forc_index_test && ../fuel-indexer/target/debug/forc-index init --namespace fuel
-
-      - name: Build new index
-        working-directory: ../forc_index_test
-        run: ../fuel-indexer/target/debug/forc-index build
+  # check-forc-index:
+  #   if: needs.check-is-release-branch.outputs.is_release_branch != 'true'
```
ra0x3, line 297: @deekerno You can just add `if: false` here vs commenting out the entire job

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/Cargo.toml
```diff
@@ -11,8 +11,9 @@ crate-type = ['cdylib']
 fuel-indexer-macros = { version = "0.3", path = "../../../../fuel-indexer-macros", default-features = false }
 fuel-indexer-plugin = { version = "0.3", path = "../../../../fuel-indexer-plugin" }
 fuel-indexer-schema = { version = "0.3", path = "../../../../fuel-indexer-schema", default-features = false }
-fuel-tx = "0.23"
-fuels-core = "0.30"
-fuels-types = "0.30"
+fuel-tx = "0.26.0"
+fuels-core = { git = "https://github.com/FuelLabs/fuels-rs", branch = "segfault_magnet/wasm_friendly_abigen", default-features = false }
+fuels-macros = {   git = "https://github.com/FuelLabs/fuels-rs", branch = "segfault_magnet/wasm_friendly_abigen"}
+fuels-types = {   git = "https://github.com/FuelLabs/fuels-rs", branch = "segfault_magnet/wasm_friendly_abigen" , default-features = false }
```
ra0x3, line 17: @deekerno  I don't think the formatting here matches the formatting of any of our other TOML files (spacing)

examples/hello-world/hello-world-data/src/main.rs
```diff
@@ -61,27 +62,24 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    let mut wallet =
+        WalletUnlocked::load_keystore(wallet_path_str, defaults::WALLET_PASSWORD, None)
+            .unwrap();
+
+    let provider = Provider::connect(defaults::FUEL_NODE_ADDR).await.unwrap();
+
+    wallet.set_provider(provider.clone());
+
+    let contract_id: Bech32ContractId =
+        "fuel18hchrf7f4hnpkl84sqf8k0sk8gcauzeemzwgweea8dgr7eachv4s86r9t9"
```
ra0x3, line 80: @deekerno Where is this magic string coming from?

deekerno: Deploying the contract and getting the generated contract ID from the node logs. I definitely don't think that these strings should stay long term, but I do think that finding an elegant way of spawning a contract deployment while simultaneously getting the contract ID may delay this a bit longer.

ra0x3:
@deekerno 
- Whereas `hello-world-native` previously worked, it seems some recent update has changed that?
- If I'm testing things wrong could you add new testing steps for the hello-world-native example?
  - I'm using `cargo run --bin fuel-node` for the node, and `cargo run --bin hello-world-data` to generate the test data

Getting errors such as
```text
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IOError(Custom { kind: Other, error: "Response errors; Transaction(0xe34303987e2c688c4fb74586468c90897cd7af043f5e2f58367a0f22d6998c01) execution error: Panic(ContractNotFound)" })', examples/hello-world/hello-world-data/src/main.rs:104:10
```
- Also the `hello-world/docker-compose.yaml` needs to be updated to use `fuel-indexer run` instead of `fuel-indexer`

examples/hello-world/hello-world-data/src/main.rs
deekerno: It seems like we may be able to use something like the first example from the SDK: https://fuellabs.github.io/fuels-rs/v0.36.1/providers/short-lived.html

ra0x3 (approved):

PR #600, opened 2023-02-19
chore: ensure proper testing and analysis of packages on CI

deekerno:
## Changelog
- Adjust CI so that tests and analyses are properly run
-- `cargo fmt --all --check` and `cargo clippy --all-targets` are run across the entire workspace
-- `cargo test --locked --workspace --all-targets --exclude fuel-indexer-tests`: this runs tests through the entire workspace excluding `fuel-indexer-tests`; this is because this command also tests doc-tests, and the path resolution in the use of the `abigen!` macro in `fuel-indexer-tests` causes the tests to fail. Workarounds described in the `cargo test` documentation did not work; this may be due to the structure of the package.
-- Because of the above issue, `cargo test --locked --all-targets --features e2e,postgres -p fuel-indexer-tests` is run in order to test E2E and integration functionality while avoiding triggering the doc-tests 
- Remove QEMU setup step as it's currently unused (and unlikely to be used at the moment due to its slowness in the build process)
- Small clippy and format fixes done to satisfy the new CI run
- Adjust a few tests to be more robust against faulty data

## Motivation
In our current CI workflow, we have a step in which we run some [tests and analyses](https://github.com/FuelLabs/fuel-indexer/blob/master/.github/workflows/ci.yml#L165) of the individual packages. However, if you look at any of the actions, it seems that we're only actually running one command (`test --locked --all-targets --features postgres --workspace`); also, it's being run as many times as we have packages. You can look at the [workflow run from the SDK upgrade PR](https://github.com/FuelLabs/fuel-indexer/actions/runs/4212175047/jobs/7311025204) as an example; click each package and you'll see that the same command is being run for each package. I ran some tests over on the ci-dummy-test repo, and [early results were promising](https://github.com/FuelLabs/ci-dummy-test/actions/runs/4217849061). Thus, I modeled our new process after both `fuel-core` and `fuels-rs`.

## Notes
- We should eventually have a different section for just the examples and run `cargo build --target wasm32-unknown-unknown` on each.
- The new workflow seems to take around the same time as our current one. We have other places that we can optimize the testing, but this is a good start.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,21 +165,13 @@ jobs:
     strategy:
       matrix:
         package: ${{ fromJSON(needs.get-workspace-members.outputs.members) }}
+        command: [clippy]
+        args: [--all-targets]
         include:
           - command: fmt
             args: --all --verbose -- --check
-          - command: clippy
-            args: --all-features --all-targets
-          - command: check
-            args: --all-features --locked --workspace --all-targets
-          - command: build
-            args: --locked --workspace --all-features --all-targets
           - command: test
-            args: --locked --workspace
-          - command: test
-            args: --locked --all-targets --no-default-features --workspace
-          - command: test
-            args: --locked --all-targets --features postgres --workspace
+            args: --locked --all-targets --all-features --workspace
 
     # disallow any job that takes longer than 45 minutes
     timeout-minutes: 45
@@ -192,11 +184,15 @@ jobs:
           toolchain: stable
       - run: cargo install sqlx-cli
       - run: bash scripts/run_migrations.bash
-      - name: ${{ matrix.command }} ${{ matrix.args }}
-        uses: actions-rs/cargo@v1
-        with:
-          command: ${{ matrix.command }}
-          args: ${{ matrix.args }}
+
+      - name: Cargo Workspace Checks
+        if: ${{ !matrix.package }}
+        run: cargo ${{ matrix.command }} ${{ matrix.args }}
+
+      - name: Cargo Package Checks
+        if: ${{ matrix.package }}
+        run: cargo ${{ matrix.command }} -p ${{ matrix.package }} ${{ matrix.args }}
+
       - name: Notify if Job Fails
         uses: ravsamhq/notify-slack-action@v2
         if: always() && github.ref == 'refs/heads/master'
@@ -209,10 +205,9 @@ jobs:
           notify_when: 'failure'
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_NOTIFY_BUILD }}
-          RUSTFLAGS: -D warnings
 
   # TODO: https://github.com/FuelLabs/fuel-indexer/issues/269
-  cargo-test-workspace-all-features:
+  cargo-test-e2e:
     if: github.event_name != 'release' && github.event.action != 'published'
     needs:
       - cancel-previous-runs
@@ -337,11 +332,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
-        with:
-          platforms: 'arm64'
 
       - name: Docker meta
         id: meta
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,24 +204,14 @@ jobs:
       - cargo-toml-fmt-check
       - check-rustc
     runs-on: ubuntu-latest
-    services:
-      postgres:
-        image: postgres:latest
-        env:
-          POSTGRES_PASSWORD: my-secret
-          POSTGRES_DB: postgres
-          POSTGRES_USER: postgres
-        ports:
-          - 5432:5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
           toolchain: stable
-      - name: cargo clippy --all-targets
-        run: cargo clippy --all-targets
+      - name: cargo clippy --all-targets --all-features
+        run: cargo clippy --all-targets --all-features
 
   cargo-fmt-check:
     if: github.event_name != 'release' && github.event.action != 'published'
@@ -230,16 +220,6 @@ jobs:
       - cargo-toml-fmt-check
       - check-rustc
     runs-on: ubuntu-latest
-    services:
-      postgres:
-        image: postgres:latest
-        env:
-          POSTGRES_PASSWORD: my-secret
-          POSTGRES_DB: postgres
-          POSTGRES_USER: postgres
-        ports:
-          - 5432:5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,14 +164,15 @@ jobs:
     if: github.event_name != 'release' && github.event.action != 'published'
     strategy:
       matrix:
-        # package: ${{ fromJSON(needs.get-workspace-members.outputs.members) }}
         include:
           - command: clippy
             args: --all-targets
           - command: fmt
             args: --all --verbose -- --check
+          # Excluding fuel-indexer-tests package as the SDK macros cause
+          # failing tests; the package is properly tested in the next job.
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -164,22 +164,15 @@ jobs:
     if: github.event_name != 'release' && github.event.action != 'published'
     strategy:
       matrix:
-        package: ${{ fromJSON(needs.get-workspace-members.outputs.members) }}
         include:
+          - command: clippy
```
ra0x3, line 168: If we're not running this on a per-package basis, should we be including it in the matrix vs just using a separate job?

.github/workflows/ci.yml
```diff
@@ -164,22 +164,15 @@ jobs:
...
             args: --all --verbose -- --check
-          - command: clippy
-            args: --all-features --all-targets
-          - command: check
-            args: --all-features --locked --workspace --all-targets
-          - command: build
-            args: --locked --workspace --all-features --all-targets
-          - command: test
-            args: --locked --workspace
+          # Excluding fuel-indexer-tests package as the SDK macros cause
```
ra0x3, line 172: Same comment here regarding putting this in a matrix but not running it on a per-package basis

.github/workflows/ci.yml
```diff
@@ -192,11 +185,15 @@ jobs:
...
       - run: cargo install sqlx-cli
       - run: bash scripts/run_migrations.bash
-      - name: ${{ matrix.command }} ${{ matrix.args }}
-        uses: actions-rs/cargo@v1
-        with:
-          command: ${{ matrix.command }}
-          args: ${{ matrix.args }}
+
+      - name: Cargo Workspace Checks
+        if: ${{ !matrix.package }}
```
ra0x3, line 190: There are no packages so this is always true?

.github/workflows/ci.yml
```diff
@@ -192,11 +185,15 @@ jobs:
...
+        run: cargo ${{ matrix.command }} ${{ matrix.args }}
+
+      - name: Cargo Package Checks
+        if: ${{ matrix.package }}
```
ra0x3, line 194: Again here, the packages were removed so this will never run?

.github/workflows/ci.yml
```diff
@@ -254,25 +249,8 @@ jobs:
...
-          CARGO_MANIFEST_DIR=$PWD/packages/fuel-indexer-tests/ target/debug/fuel-node \
-            --chain-config packages/fuel-indexer-tests/assets/test-chain-config.json \
-            --contract-bin packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test.bin &
-
-      - uses: actions-rs/cargo@v1
-      - name: Cargo e2e tests
-        run: bash packages/fuel-indexer-tests/scripts/e2e.bash
-      - name: Stop testing components
-        run: kill -9 $(lsof -ti:4000)
+      - name: Run E2E and integration tests
```
ra0x3, line 252: 👌🏽

.github/workflows/ci.yml
```diff
@@ -164,22 +164,15 @@ jobs:
     if: github.event_name != 'release' && github.event.action != 'published'
     strategy:
       matrix:
-        package: ${{ fromJSON(needs.get-workspace-members.outputs.members) }}
```
ra0x3, line 167: You removed the packages here

ra0x3 (changes requested):
@deekerno 
- Left some feedback
  - Mainly around the fact that the "Cargo Checks" step isn't running (screenshot from your PR below) because we removed the packages
    - Let me know if I'm misreading here
Screen Shot 2023-02-22 at 4 11 00 PM

- I do think we should keep `build` in the `cargo-verifications` matrix as we've been bitten in the past during a release, where an individual crate will pass CI but break the release because the publish step builds crates independently

deekerno:
Both good points!

As far as I can tell, clippy does cargo check if you don't pass in the --fix argument: https://github.com/rust-lang/rust-clippy/blob/master/src/main.rs. So with clippy running across the workspace, it covers each of the packages. Let me know if you'd rather that it be separated by package; for context, I believe fuel-core does it by workspace and I think fuels-rs just fixed their CI to do the same. 

Regarding the builds, I can definitely add it back. 👌🏽

deekerno:
@ra0x3 With this new workflow, we do the following across the workspace --
- `cargo clippy`
- `cargo fmt --check`
- unit tests

-- along with separate builds for each package and a separate step for E2E and integration tests.

.github/workflows/ci.yml
```diff
@@ -209,14 +196,91 @@ jobs:
...
+  cargo-clippy:
     if: github.event_name != 'release' && github.event.action != 'published'
     needs:
       - cancel-previous-runs
-      - get-workspace-members
+      - cargo-toml-fmt-check
+      - check-rustc
+    runs-on: ubuntu-latest
+    services:
+      postgres:
```
ra0x3, line 208: I don't think we need postgres for `clippy`?

.github/workflows/ci.yml
```diff
@@ -209,14 +196,91 @@ jobs:
...
+
+  cargo-fmt-check:
+    if: github.event_name != 'release' && github.event.action != 'published'
+    needs:
+      - cancel-previous-runs
+      - cargo-toml-fmt-check
+      - check-rustc
+    runs-on: ubuntu-latest
+    services:
+      postgres:
```
ra0x3, line 234: Same here, I don't think we need postgres for `rustfmt`?

.github/workflows/ci.yml
```diff
@@ -209,14 +196,91 @@ jobs:
...
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+      - name: cargo clippy --all-targets
+        run: cargo clippy --all-targets
```
ra0x3, line 224: Could we add `--all-features` as well here?

packages/fuel-indexer-tests/tests/integration/web_api_postgres.rs
```diff
@@ -129,20 +133,25 @@ async fn test_asset_upload_endpoint_properly_adds_assets_to_database_postgres()
 
     let client = http_client();
     let resp = client
-        .post("http://127.0.0.1:29987/api/index/fuel_indexer_test/index1")
+        .post("http://127.0.0.1:29987/api/index/test_namespace/simple_wasm_executor")
```
ra0x3, line 136: Why this change?

packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
```diff
@@ -165,24 +165,32 @@ async fn test_can_trigger_and_index_blocks_and_transactions_postgres() {
 
     let contract = connect_to_deployed_contract().await.unwrap();
     let app = test::init_service(app(contract)).await;
+    let req = test::TestRequest::get().uri("/block_height").to_request();
```
ra0x3, line 168: Why the call here?

ra0x3 (changes requested):

.github/workflows/ci.yml
deekerno: Oh, you know, I included it for `sqlx` but I forgot that we had `SQLX_OFFLINE=1` set at the very beginning of the file. I'll change.

packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
deekerno: I found that this this test may fail if it's run after the other tests directly after other tests as we don't enforce the block heights to be unique upon insertion and the fuel node fixture's chain resets on every run, which leads to multiple records where `block_height = 1`. So it checks for a block that has the height of 1, but if you run a couple of other tests before this and then run the query, it may grab the wrong block and the TX length assertion will fail. Thus, I added a check for the current block height plus one to ensure that the correct block is selected.

If you want to verify the failure, run the `e2e::graphql_api_postgres` tests and then run the `e2e::indexing_postgres` tests directly after with no database refresh; you should see a failure in this test.

packages/fuel-indexer-tests/tests/integration/web_api_postgres.rs
deekerno: If you run this test in sequence with the other tests (with no database refresh), then the test will fail as the assets are already present in the database. Since we're already using the `simple_wasm` assets as test uploads, I changed the route to match the manifest details in order to prevent any collisions.

ra0x3 (approved):
CI cut by 33%? I hope things average out that way 🙏🏽

Issue #590, opened 2023-02-16
Replace all `beta-2` doc references with `beta-3`


ra0x3:
- Where possible, we recommend users connect to Fuel's `beta-2` test network
- Now that `beta-3` is out, we should update our docs to reference `beta-3`

PR #609, opened 2023-02-22
enhancement: add `forc index revert` command

### LINKED ISSUE #464: Add `revert` command to `forc index`
In the event that a new asset deployment breaks a previously working index, there should be a way to quickly revert back to the previous version of the index so that indexing is interrupted as little as possible. This would need a new command to be added to `forc index` as well as a new endpoint in the API server.

rvmelkonian:
Closes #464

## Changelog 
• creates new `forc index revert command` and associated files for the new command to work.  
• updates docs for the new command
• adds new route for the revert op
• adds new queries method `penultimate_index_id_for`
• adds new match arm for `create_service_task()` inside `service.rs`.

## Testing
1. Start the service with a fresh database and using the block explorer example: `cargo run --bin fuel-indexer -- run --fuel-node-host node-beta-2.fuel.network --fuel-node-port 80`
2. Deploy the block explorer example with a log line.
3. Deploy the block explorer example a second time with a different log line.
4. Stop the service.
5. Immediately start it again.
6. Revert the index WASM using the explorer manifest.

 ```
 ⬅️  Revertingindex 'my_project.test_index' at http://127.0.0.1:29987/api/index/my_project/test_index

{
  "success": "true"
}

✅ Successfully reverted index 'my_project.test_index' at http://127.0.0.1:29987/api/index/my_project/test_index
```
If you don't see this make sure to rebuild the binary in step 1.

7. Trigger some events manually 
8. Check in the DB that the penultimate id of `hello-index` (id 1) is indexing the events, not id 2

## Notes
I noticed that in all the other plugin commands, the service is interacted with via the api, which I think is a good and clean approach. As there was no api route for reverting an index, I created a new one. This then triggers the `create_servce_task()` and the new `ServiceRequest::IndexRevert` is matched and executed.

### CODE REVIEW

File: docs/README.md [hunk 1]
```diff
@@ -76,7 +76,7 @@ rustup target add wasm32-unknown-unknown
 In this tutorial you will:
 
 1. Bootstrap your development environment.
-2. Create, build, and deploy an index to an indexer service hooked up to Fuel's `beta-2` testnet.
+2. Create, build, and deploy an index to an indexer service hooked up to Fuel's `beta-3` testnet.
```
ra0x3, line 79: Per the discussion in https://github.com/FuelLabs/fuel-indexer/issues/590, I don't think `beta-3` is ready yet?

ra0x3, line 79: Per the discussion in https://github.com/FuelLabs/fuel-indexer/issues/590, I don't think `beta-3` is ready yet?

docs/src/reference-guide/plugins/forc-index/index.md
```diff
@@ -22,5 +22,6 @@ SUBCOMMANDS:
     init      Create a new indexer project in the current directory
     new       Create a new indexer project in a new directory
     remove    Stop and remove a running index
+    revert    Revert a running index to the last deployed version
```
ra0x3, line 25: I should probably clarify some terminology at our next team sync as I think I started this trend of calling indexers indices 😅 
```suggestion
    revert    Revert a running indexer to its last previous version
```

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -173,6 +173,38 @@ pub(crate) async fn stop_index(
...
+    Path((namespace, identifier)): Path<(String, String)>,
+    Extension(tx): Extension<Option<Sender<ServiceRequest>>>,
+    Extension(pool): Extension<IndexerConnectionPool>,
+) -> ApiResult<axum::Json<Value>> {
+    let mut conn = pool.acquire().await?;
+
+    let _ = queries::start_transaction(&mut conn).await?;
+
+    if let Err(_e) =
+        queries::penultimate_index_id_for(&mut conn, &namespace, &identifier).await
```
ra0x3, line 186: TIL [penultimate](https://www.merriam-webster.com/dictionary/penultimate). Love it.

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -173,6 +173,38 @@ pub(crate) async fn stop_index(
...
 }
 
+pub(crate) async fn revert_index(
+    Path((namespace, identifier)): Path<(String, String)>,
+    Extension(tx): Extension<Option<Sender<ServiceRequest>>>,
+    Extension(pool): Extension<IndexerConnectionPool>,
+) -> ApiResult<axum::Json<Value>> {
+    let mut conn = pool.acquire().await?;
+
+    let _ = queries::start_transaction(&mut conn).await?;
```
ra0x3, line 183: Since `penultimate_index_id_for` is just a `SELECT` I don't think we need a transaction?

ra0x3 (changes requested):
@rvmelkonian 
- Awesome work on this so far, looks great! 🥳 
- Left a few small pieces of feedback
- Will test once those ☝🏽 are addressed

packages/fuel-indexer-api-server/src/uses.rs
rvmelkonian: Maybe I'm misunderstanding the code, but don't we need to fire a `tx.send()` so that `rx.try_recv()` in `service.rs` will receive the value?

ra0x3: @rvmelkonian 
- You're correct that we're going to send the results of `penultimate_index_id_for` to the indexer service via `tx.send`
- However, since `penultimate_index_id_for` isn't a mutation query (e.g., an `INSERT` or `DELETE`) then we don't really need a transaction
  - The other use cases in `uses.rs` that use transactions are queries that mutate the state of the database

rvmelkonian: got it, yes we are not mutating here. 
So we need `tx.send` but not this:
```
    let _ = queries::start_transaction(&mut conn).await?;

    if let Err(_e) =
        queries::penultimate_index_id_for(&mut conn, &namespace, &identifier).await
    {
        queries::revert_transaction(&mut conn).await?;
    } else {
        queries::commit_transaction(&mut conn).await?;
    }
    ```

File: packages/fuel-indexer-database/postgres/src/lib.rs [hunk 5]
```diff
@@ -612,6 +612,25 @@ pub async fn index_id_for(
...
+    conn: &mut PoolConnection<Postgres>,
+    namespace: &str,
+    identifier: &str,
+) -> sqlx::Result<i64> {
+    #[cfg(feature = "metrics")]
+    METRICS.db.postgres.penultimate_index_id_for_calls.inc();
+
+    let query = format!(
+    "SELECT id FROM index_registry WHERE namespace = '{namespace}' AND identifier = '{identifier}' ORDER BY id DESC LIMIT 1, 1",
+    );
```
deekerno, line 625: Interestingly, I get this error message whenever I do the revert operation:
```
ERROR fuel_indexer::service: 328: Failed to find Index(fuel_examples.explorer_index): error returned from database: LIMIT #,# syntax is not supported
```

plugins/forc-index/src/ops/forc_index_revert.rs
```diff
@@ -0,0 +1,69 @@
...
+        url,
+        path,
+        manifest,
+        ..
+    } = command;
+
+    let (_root_dir, manifest_path, _index_name) =
+        project_dir_info(path.as_ref(), manifest.as_ref())?;
+
+    println!("manifest_path: {:?}", manifest_path);
```
deekerno, line 22: Can we remove this as the user has to pass the manifest path in for the command to work?

deekerno (changes requested):
Looking great, I think it's almost there! I left a couple of comments.

Additionally, I think we should consider removing the reverted index from the database; if you restart the service after reverting an index, the previously reverted index will be the one executed.

Reproduction steps:
- Deploy an index with a log message included, e.g. "This is the first index." You should see this in the service logs.
- Deploy a second index with a different log message, e.g. "This is the second index." Verify that it's in the logs.
- Revert the second index.
- Stop the index and start it again.
- You should see that the second index's message is the one that appears.

packages/fuel-indexer-database/postgres/src/lib.rs
rvmelkonian: hmm interesting, I haven't seen that error on my side. Did you run migrations before testing? What postgres version are you on?

rvmelkonian:
@deekerno Yes I had also thought about removing the reverted index from the DB. I will update this with that logic, I think it makes sense and of course removes the bug you've mentioned.

rvmelkonian:
@deekerno I've added the `queries::remove_index` method *after* the reversion happens, inside service.rs. I did this so that the delete mutation will only happen if `queries::penultimate_index_id_for` does not return an error. The only downside I can think of this is that we don't have the safety of doing the DELETE within a transaction. Lmk what you think about this. 

As it stands now, after the revert command, the reverted index would be the latest index in the DB.

packages/fuel-indexer/src/service.rs
```diff
@@ -277,6 +277,80 @@ async fn create_service_task(
...
+                                {
+                                    Ok(_) => info!(
+                                        "Removed index: {}.{}",
+                                        &request.namespace, &request.identifier
+                                    ),
+                                    Err(e) => error!(
+                                        "Failed to remove index: {}.{}: {}",
+                                        &request.namespace, &request.identifier, e
+                                    ),
+                                }
```
deekerno, line 342: So `remove_index` removes the _entire_ index, as it deletes any rows that match to the `index_id` for the namespace and identifier. In contrast, I'd expect a removal of the reverted index to remove the most recent `version` of the WASM asset for said `index_id`. My apologies if I caused a misunderstanding when suggesting it earlier.

This would require a separate function to be created or for `remove_index` to take at least another argument to instruct what should be deleted.

@ra0x3: Thoughts?

packages/fuel-indexer-database/postgres/src/lib.rs
deekerno: PostgreSQL 14, the one that's installed through `brew install postgresql`. And yep, I've run the migrations fresh on each test.

It looks like you can use `LIMIT X OFFSET Y` to skip an arbitrary number of rows: https://www.postgresql.org/docs/current/queries-limit.html; testing it locally worked for me.

deekerno (changes requested):
@rvmelkonian: Left another comment concerning the removal of the reverted index. Once that's addressed, I'll re-test and approve.

Regarding not having a transaction, is there any reason that we couldn't start one in the `Ok(id)` match arm of `queries::penultimate_index_id_for(..)`? I think we should be able to start one with the same connection by doing ```let _ = queries::start_transaction(&mut conn).await?;```

rvmelkonian:
@deekerno I've tested with you steps above https://github.com/FuelLabs/fuel-indexer/pull/609#pullrequestreview-1322503833 and it's working as expected. However, note that after reverting, the indexer has to be restarted for the penultimate wasm asset to be the active index. 

After doing this revert I'm able to successfully continue deploying other indexes.

deekerno (dismissed):
LGTM! Thanks for adding this!

Eventually, we should work on our API errors across the product to ensure that users have a better idea of what's going on when things go wrong. But this is definitely good for now.

packages/fuel-indexer-api-server/src/api.rs
```diff
@@ -1,14 +1,14 @@
...
+    revert_index, stop_index,
 };
 use async_std::sync::{Arc, RwLock};
 use axum::{
     extract::{Extension, Json},
     http::StatusCode,
     middleware,
     response::{IntoResponse, Response},
-    routing::{delete, get, post},
+    routing::{delete, get, post, put},
```
ra0x3, line 11: 👌🏽

ra0x3 (changes requested):
@rvmelkonian 
- The logic looks fine
- Left some minor comments and one comment about moving some of the indexer swap logic out of the service module and into the use case module

ra0x3 (commented):
- [ ] Deployed the block explorer index
- [ ] Made changes to the example then redeployed
- [ ] Confirmed that the 2nd index replaced the 1st
- [ ] Ran `forc-index revert --path /path/to/indexer` and saw that the latest version was removed while 
- [ ] Confirmed that the 1st indexer came back while the 2nd was removed

@rvmelkonian This works, will approve after the last small feedback is addressed

packages/fuel-indexer/src/service.rs
```diff
@@ -285,6 +285,65 @@ async fn create_service_task(
...
+                            queries::revert_transaction(&mut conn)
+                                .await
+                                .expect("Failed to revert transaction");
+                        } else {
+                            queries::commit_transaction(&mut conn)
+                                .await
+                                .expect("Failed to commit transaction");
+                        }
+
+                        let manifest: Manifest =
```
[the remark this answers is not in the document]
rvmelkonian: @ra0x3 this has been added in now.

ra0x3 (approved):
:shipit:

Issue #544, opened 2023-01-30
Add `forc pg` plugin


ra0x3:
- If you checkout our [Quickstart](https://fuellabs.github.io/fuel-indexer/master/quickstart/index.html) you'll see we use Docker in order to get users up and running as fast as possible  - we optimized for speed
- This works fine - but it has been mentioned that we should opt-out of Docker as a primary point of tooling
  - With the mention of Docker in our examples and in our README/docs, it gives the vibe that Docker is super important - but it isn't
  - Docker is super useful - but it's very enterprise-y, and hugs too many resources
  - Using Docker in our examples/docs was fine because we were optimizing for speed of getting started 
- We should add support for a better way to bootstrap a Postgres backend, without having to use Docker
- This functionality would live as a plugin

Additional context
- Thinking this would be a separate plugin "forc postgres" or `forc pg`
  - I _don't_ think we should bake this into `forc index` directly - but if someone has a case for it, I'm open to hearing
- `forc index` would be able to invoke `forc pg` functionality
- This `forc pg` would basically allow users to use Postgres (in a Quickstart-like manner) without using Docker
  - We add a lot of other config later if neccesary 
- Someone did us the favor of creating a little PoC using https://docs.rs/pg-embed/latest/pg_embed/
- I'll paste the PoC below (don't have to use this exactly, just inspiration)

```rust
use pg_embed::postgres::{PgEmbed, PgSettings};
use pg_embed::pg_fetch;
use pg_embed::pg_enums::{PgAuthMethod, PgServerStatus};
use pg_embed::pg_fetch::{PgFetchSettings, PG_V13};
use pg_embed::pg_errors::PgEmbedError;
use std::time::Duration;
use std::path::PathBuf;

// async block only to show that these methods need to be executed in an async context
#[tokio::main]
async fn main() -> Result<(), PgEmbedError> {
    // Postgresql settings
    let pg_settings:PgSettings = PgSettings {
        // Where to store the postgresql database
        database_dir: PathBuf::from("data/db"),
        port: 5432,
        user: "postgres".to_string(),
        password: "password".to_string(),
        // authentication method
        auth_method: PgAuthMethod::Plain,
        // If persistent is false clean up files and directories on drop, otherwise keep them
        persistent: false,
        // duration to wait before terminating process execution
        // pg_ctl start/stop and initdb timeout
        // if set to None the process will not be terminated
        timeout: Some(Duration::from_secs(15)),
[...]

PR #610, opened 2023-02-22
docs: de-emphasize docker in docs

ra0x3:
### Description
- As previously mentioned, with https://github.com/FuelLabs/fuel-indexer/issues/544 done, we need to de-emphasize Docker for users in our docs

### Changelog
- docs: de-emphasize docker in docs

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -58,6 +58,7 @@ Users of the Fuel indexer project include dApp developers looking to write flexi
 ### `wasm`
 
 Two additonal cargo components will be required to build your indexers: `wasm-snip` and the `wasm32-unknown-unknown` target.
+
 - To install `wasm-snip`:
 
 ```bash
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -512,6 +504,13 @@ Contributors of the Fuel indexer project are devs looking to help  backends for
 
 ## Dev Dependencies
 
+### `docker`
```
ra0x3, line 507: Moving from the users section down to the devs section

deekerno (changes requested):
Looks good! The only other thing is that we have to adjust the `include` for the quickstart page; hopefully one day, we won't have to update it so much. The line numbers should now be from lines 81 to 405.

ra0x3:
@deekerno 
- Good catch, looks like that piece could use a bit of automation as well 🤔 
- Fixed in latest commit

deekerno (approved):

PR #617, opened 2023-02-23
chore: change `fuel-indexer`'s `fuel-core` dep feature flag from `local-node` to `fuel-core-lib`

ra0x3:
### Description
- Removes .DS_Store files
- Changes `fuel-core` feature flag from `local-node` to `fuel-core-lib` ([for parity with the sdk](https://github.com/FuelLabs/fuels-rs/blob/master/packages/fuels/Cargo.toml#L40))
- Update: I don't think this works on master either, so tagging as `bugfix`

### Testing steps
- [ ] Start service `cargo run --bin fuel-indexer --features fuel-core-lib -- run`
- [ ] Make sure node is up `curl http://127.0.0.1:4000/health`
  - Should produce `{"up":true}` 

### Changelog 
- chore: change `fuel-indexer`'s `fuel-core` dep feature flag from `local-node` to `fuel-core-lib`


deekerno (approved):
🚢

PR #620, opened 2023-02-24
chore: create bash script `clean_and_test`

rvmelkonian:
No issue for this. 

## Changelog
• creates new `script/utils` file `clean_and_test.bash`. 

## Notes
I found myself getting tripped up a few times when wanting to run the tests and not having ran other commands first. This is for the lazier devs among us. This script just calls the others in order and runs all the e2e tests, incase a dev is running into issues, they can run this. 

## Testing
Run `bash scripts/utils/clean_and_test.bash`


ra0x3:
- So we actually don't even use e2e.bash anymore (thanks @deekerno  🥳 )
  - This file should be removed from the project
- refresh_test_db.bash exists because it saves you from having to cd into a directory, run a command, cd back
  - However, I could see the case that we need to remove this file too and just use your shell history
- And build_test_wasm.bash exists because it would be super painful to run those commands separately and build about 4 different indexers

- So one of the files referenced in this PR isn't used anymore, another one can be easily removed (and probably should), while there's just one that should probably stay (for now)
- In general, I think we're trying to move away from putting some of this local-dev stuff in the repo (since it is a public repo)
  - We still have some work to do here
    - Just need time to go through and catalogue the files that need to be removed 
  - But I don't think we want to get in the habit of adding too many convenience scripts (they are just 3 easy commands after all)
   - You could always keep the file locally and still use them

@deekerno thoughts?

rvmelkonian:
@ra0x3 that's fair maybe it makes the repo dirty a bit. I'll use it locally. The main thought here was when running tests, if you see an error you immedietly assume its your changes causing the error, then after running this or that command, the tests pass, so it was just to remove that second guessing.

deekerno:
> thanks @deekerno 🥳 

❤️ 

So the only one of the mentioned files that I use `build_test_wasm.bash` and even that's a stretch as I've been moving to using aliases on my own system for dropping the DB, generating macro output, etc; the new CI workflow doesn't use it either. Additionally, I don't even know that we need `run_migrations.bash` as the service fixture can be configured to run them for us. 

_Personally_, I'd like to see each of the files removed as I don't like having to run scripts (but that may be a bit extreme); however, I don't see too much of an issue with them existing if they help other devs. In any case, I'm not sure that adding another script to the repository would be beneficial.

PR #622, opened 2023-02-25
enhancement: support foreign keys and nested queries in graphql api

### LINKED ISSUE #460: Support foreign key relationships in GraphQL API queries

deekerno:
Closes #460.

This PR adds the ability to support nested queries in the indexer's GraphQL API.

## Changelog
- Adds a `foreign_keys` field to `Schema` so that FKs can be retrieved at query time
- Adjusts `SchemaManager.load_from_db(..)` to parse schema string already present in DB and get FKs
- Changes `Operation.as_sql(..)` to `Operation.parse(..)`, and parses user queries into a `UserQuery` struct
- Adds functionality to parse `UserQuery`s into the corresponding storage backend queries
- Cleans up old, unused references to SQLite

## Testing Plan
Extensive tests were implemented; as such, most of the diff is actually testing code.

Unit tests were added in the following places:
- `fuel-indexer-schema/src/tables.rs`: Covers foreign key retrieval
- `fuel-indexer-schema/src/graphql.rs`: Covers `UserQuery` parsing
- `fuel-indexer-database/database-types/src/lib.rs`: Covers parsing of `UserQuery` into Postgres-specific query

New end-to-end tests were added in order to:
- Ensure implicit foreign keys are properly supported
- Ensure explicit foreign keys are properly supported
- Ensure deeply nested queries are properly supported
-- Inspiration for this test was taken from #588 and is deliberately constructed to be obtuse and confusing. It covers a number of edge cases, each of which are commented inside the test itself.

### Manual Testing
0. Pull down and checkout the https://github.com/FuelLabs/fuel-indexer/tree/deekerno/manual-test-for-622 branch. Ensure the database is cleared.
1. Build the test index for this example: `cargo build -p manualtest_index --release --target wasm32-unknown-unknown`.
2. Run `wasm-snip` on the binary and copy it to the `target` folder: `bash scripts/stripper.bash manualtest_index.wasm && cp manualtest_index.wasm target/wasm32-unknown-unknown/release/manualtest_index.wasm`.
3. Start a test Fuel node: `cargo run --bin fuel-node`.
4. Start the test web API: `cargo run --bin web-api`.
5. Start the indexer service with the example manifest: `cargo run --bin fuel-indexer -- run --manifest examples/manualtest/manualtest-index/manualtest_index.manifest.yaml`.
6. Generate a test block: `curl -X POST http://127.0.0.1:8000/block`. At this point, you should see a lot of inserts in the indexer service logs.
7. Run the following query and ensure that the returned data matches what's expected from the query:
```
curl -X POST http://127.0.0.1:29987/api/graph/fuel_examples/graphql_manualtest \
   -H 'content-type: application/json' \
   -d '{"query": "query { bookclub { id book { id name author { name genre { id name } } library { id name city { id name region { id name country { id name } } } } genre { id name }} member { name id } corporate_sponsor { id name amount } } }", "params": "b"}'
```
8. Check the indexer service logs for the SQL query that was generated and executed to return the data.

## Notes
A good chunk of this was previously reviewed in #563. The most significant difference between that PR and this one is the way through which database joins are created and collected has changed. Previously, joins were simply collected into a vector and then processed by iteration. However, due to query constraints imposed by Postgres (namely, the need for joins on different tables to be topologically sorted as well as prohibiting statements that ultimately join the same table), they are now placed into a map structure in order to efficiently satisfy those requirements. As a result, the transformation into SQL statements had to be adjusted as well. The insertion into the maps can be found in `Operation::parse()` while the SQL transformation can be found in `UserQuery::to_sql()`.


ra0x3 (approved):
@deekerno 
- Testing steps work for me.
- There's obviously a thousand different ways to test and edge cases to include and such, but given the manual testing and the e2e test that was added, I think we've got 80% of what we need with 50% of the work.
- If it's determined that this implementation is missing something glaring, or users can't do something obvious then we can come back and revisit

🚢

PR #623, opened 2023-02-27
ci: streamline ci

ra0x3:
### Description
- PR streamlines our CI a bit by:
  - Removing unnecessary packages from the build matrix
    - We can always add them back later if we want
  - Updating job dependencies
  - Using flags like `IS_RELEASE_OR_MASTER`

- Looks like PR CI runtime is at 16 - 20 mins?
- TODO: publish crates in parallel - this PR does not do that (don't want to bloat this PR too much)

### Testing steps
- [ ] CI should pass
  - Via the eye-ball test, looks like master/release will work as expected

### Changelog 
- ci: streamline ci


packages/fuel-indexer/Cargo.toml
```diff
@@ -17,55 +17,31 @@ path = "src/bin/main.rs"
...
-fuel-indexer-postgres = { workspace = true }
-fuel-indexer-schema = { workspace = true, features = ["db-models"] }
-fuel-indexer-types = { workspace = true }
-fuel-tx = "0.26.0"
-fuel-types = "0.26.0"
+fuel-indexer-api-server = { version = "0.4", path = "../fuel-indexer-api-server", optional = true }
+fuel-indexer-database = { version = "0.4", path = "../fuel-indexer-database" }
+fuel-indexer-lib = { version = "0.4", path = "../fuel-indexer-lib" }
+fuel-indexer-schema = { version = "0.4", path = "../fuel-indexer-schema", features = ["db-models"] }
+fuel-indexer-types = { version = "0.4", path = "../fuel-indexer-types" }
```
deekerno, line 29: What's the reasoning for changing this from `{ workspace = true }`?

deekerno (approved):
Looks good, thanks for doing this! Left just a single non-blocking question about dependencies.

packages/fuel-indexer/Cargo.toml
ra0x3: @deekerno Nope. Bad rebase, will fix.

deekerno (approved):
Loving that new lint usage too. 👌🏽

PR #630, opened 2023-03-05
enhancement: indexer authentication

ra0x3:
### Description
- This PR adds support for user authentication against an indexer operator
- Authentication is used to protect certain endpoints that mostly support write operations

#### Authentication changes
- Adds an `authentication` section to the config
- Authentication is turned off by default
- This implementation is setup to support multiple types of auth. 
  - JWT is the default/1st authentication scheme supported because its lightweight, stateless, and fairly ubiquitous (thus understandable by a lot of devs)
- Adds docs for this new functionality
  - ⚠️ Please review the docs as well 
- Adds a `forc index auth` command
- Updates `forc index check` to include the new `forc-wallet` dependency
- Adds new migrations

#### Other non-authentication changes
- Also makes a few changes here and there for other things including:
  - Changing the `max_body` node config opt to `max_body_size`
  - Using the `default_value_t` for non-str types (instead of using `String` then converting the types)
  - Updates some of the default config options (so that we can just pass the flag `--flag` instead of a flag and value `--flag value`)
  - Updates `forc index start`
- Updates other migration names (some of them naming was not standard)

### How does this authentication work?
- As mentioned, the 1st/default scheme we'll support is JWT because its lightweight, stateless, and popular
- A user will authenticate as follows:
  - Create a Fuel wallet with `forc-wallet`
  - Have some indexer service already running somewhere
  - Use `forc index auth`
    - Account index `0` in the wallet will be used by default
  - User is prompted for their wallet password
  - If password is successful, a request is made to `http://indexer.operator.com:29987/auth/nonce` to get a nonce
    - These nonces are short-lived-state managed in  a `nonce` table. Why?
      - So the indexer operator can be in charge of who's allowed to auth against their service
        - Else anyone could just send a random signed string and as long as the signature is valid, they would be authenticated
      - Nonces have an expiry so that they can't be used forever.
  - That nonce is then signed using the user's wallet private key
  - This signature, and the nonce are sent back to the indexer operator
  - The indexer operator uses the signature and the original message (nonce) to derive the users's public key (using  the `fuel-crypto` crate)
  - When the public key is derived, we assert that the user did indeed sign the message
  - When we see the derived public key asserts the message was signed by the corresponding private key, when create `Claims`
    - These `claims` include the issuer (set in the node config), the public key (of the authenticated user), and an expiry (when the token expires)
      - Tokens have 30-day expiry by default
   - These `claims` are encoded and then sent back to the user as a JWT
   - The user then includes this token in their `Authorization` header for all requests that require authentication
     - When using the plugin, the token is passed via the `--auth` flag
   - In our middleware, we set `Claims` in our Request extensions.
     - This means for every route/handler that requires authentication, we will use a `Claims` extension on that route.
       - In those `claims` we can look at the `sub` (subscriber) field to see the public key of the user who's taking the action.
     - One public key can own one or many indexers

### Testing steps 
- ⚠️ There are no integration tests for this at the moment, need to do some more thinking about the best way to do that and don't want it to be a blocker

#### Prerequisites
- [ ] Referesh to test DB
- [ ] Build your WASM modules
- [ ] Build a new `forc-index` binary and reference it via `./target/release/forc-index` (don't copy it over to fuelup)
- [ ] You should be able to use the `forc-wallet` that comes with fuelup

#### Denied auth
- [ ] Start fuel node `cargo run --bin fuel-node`
- [ ] Start the indexer with JWT auth enabled
```
cargo run --bin fuel-indexer -- run --auth-enabled --auth-strategy JWT --jwt-secret 6906573247652854078288872150120717701634680141358560585446649749925714230966 --jwt-issuer FuelLabs
```
- [ ] Try to deploy to an indexer without authenticating
```
forc-index deploy \ 
   --path examples/block-explorer/explorer-index \
   --output-dir-root /Users/rashad/development/repos/fuel-indexer \
   --url http://localhost:29987 \
   --target wasm32-unknown-unknown \
   --release
```
- [ ] Should see output similar to
```
▹▹▹▹▸ ⏰ Building...                                                                                                                       Finished release [optimized] target(s) in 1.19s
▪▪▪▪▪ ✅ Build succeeded.                                                                                                              Deploying indexer at /Users/rashad/development/repos/fuel-indexer/examples/block-explorer/explorer-index/explorer_index.manifest.yaml to http://localhost:29987/api/index/fuel_examples/explorer_index
▸▹▹▹▹ 🚀 Deploying...
❌ http://localhost:29987/api/index/fuel_examples/explorer_index returned a non-200 response code: 401

{
  "details": "Unauthorized.",
  "success": "false"
}
```
- [ ] This same thing should happen when try to `forc-index revert` or `forc-index remove` an indexer

#### Successful auth
- [ ] Create a wallet `forc wallet new` 
- [ ] Start an indexer service with JWT auth enabled
```text
cargo run --bin fuel-indexer -- run --auth-enabled --auth-strategy JWT --jwt-secret 6906573247652854078288872150120717701634680141358560585446649749925714230966 --jwt-issuer FuelLabs
```
- [ ] Authenticate at the running indexer service `forc-index auth`
  - You should get a JWT back
- [ ] Now deploy an indexer 
```
forc-index deploy \
  --path examples/block-explorer/explorer-index \
  --output-dir-root /Users/rashad/development/repos/fuel-indexer \
  --url http://localhost:29987 \
  --target wasm32-unknown-unknown \
  --release \
  --auth $FOOBAR
```
- [ ] The deployment should work

### Changelog
- [ ] enhancement: indexer authentication


packages/fuel-indexer-lib/src/config/auth.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Issuer.
+    pub iss: String,
+
+    /// Issued at (as UTC timestamp).
+    pub iat: usize,
+
+    // Expiration time (as UTC timestamp).
+    pub exp: usize,
+}
```
deekerno, line 101: Could we expand these names to make this a bit more self-documenting? Also, line 99 should be a doc comment for consistency.

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -293,6 +293,78 @@ pub(crate) async fn register_index_assets(
...
+) -> ApiResult<axum::Json<Value>> {
+    if config.authentication.enabled {
+        let mut conn = pool.acquire().await?;
+        match config.authentication.strategy {
+            Some(AuthenticationStrategy::JWT) => {
+                let nonce = queries::get_nonce(&mut conn, &payload.message).await?;
+                queries::delete_nonce(&mut conn, &nonce).await?;
+
+                let mut buff: [u8; 64] = [0u8; 64];
+                buff.copy_from_slice(&payload.signature.as_bytes()[..64]);
```
deekerno, line 318: This ends up copying the whole signature anyways; see [this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=336b9b9274fb64f1759fd889651250c7). You could get away with just doing `buff.copy_from_slice(&payload.signature.as_bytes()[..])` unless we plan on adding something to the end of the signature in the near future.

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -293,6 +293,78 @@ pub(crate) async fn register_index_assets(
...
+    Extension(config): Extension<IndexerConfig>,
+    Extension(pool): Extension<IndexerConnectionPool>,
+    Json(payload): Json<VerifySignatureRequest>,
+) -> ApiResult<axum::Json<Value>> {
+    if config.authentication.enabled {
+        let mut conn = pool.acquire().await?;
+        match config.authentication.strategy {
+            Some(AuthenticationStrategy::JWT) => {
+                let nonce = queries::get_nonce(&mut conn, &payload.message).await?;
+                queries::delete_nonce(&mut conn, &nonce).await?;
```
deekerno, line 315: Should we wait to delete the nonce until we know that the signature is at least recovered as it can possibly fail? Just a thought; I suppose if the signature doesn't get verified, then a new nonce should be used anyways?

deekerno (changes requested):
Every step in the testing plan works except for the deployment.

```
SELECT
  *
FROM
  index_registry
WHERE
  namespace = 'fuel_examples'
  AND identifier = 'explorer_index'

2023-03-14T22:40:39.820394Z  INFO sqlx::query: SELECT * FROM index_registry …; rows affected: 0, rows returned: 0, elapsed: 173.292µs

SELECT
  *
FROM
  index_registry
WHERE
  namespace = 'fuel_examples'
  AND identifier = 'explorer_index'

2023-03-14T22:40:39.821598Z  INFO sqlx::query: INSERT INTO index_registry (namespace, …; rows affected: 0, rows returned: 0, elapsed: 433.375µs

INSERT INTO
  index_registry (namespace, identifier, pubkey)
VALUES
  (
    'fuel_examples',
    'explorer_index',
    '0d2777576c50e5b1fe690875aaa8995d9832ed18817dd9005b1c9bb588a1a1aae9a1600edb3d4d9c07dc4112c452fb1fcc3fc589909a50789e290693b73c1538'
  ) RETURNING *

2023-03-14T22:40:39.821639Z ERROR fuel_indexer_api_server::api: 106: 500 - Internal server error.
2023-03-14T22:40:39.821709Z  INFO tower_http::trace::on_response: 254: finished processing request latency=4980 μs status=500
2023-03-14T22:40:39.821722Z ERROR tower_http::trace::on_failure: 137: response failed classification=Status code: 500 Internal Server Error latency=4 ms
```

Additionally, can we add a helpful error message when a signature can't be captured from the output?
```
❯ ./target/release/forc-index auth
Please enter your password:
thread 'main' panicked at 'Failed to capture signature output.', plugins/forc-index/src/ops/forc_index_auth.rs:66:18
zsh: abort      ./target/release/forc-index auth
```

docs/src/reference-guide/authentication/index.md
```diff
@@ -0,0 +1,30 @@
+# Authentication
+
+The Fuel Indexer service functionality offers users a range of options for verifying their identity. The system supports any arbitrary authentication scheme (in theory), although it in practice the service defaults to JWT authentication due to its stateless nature and popularity. To authenticate using JWT, users ask an index operator for a nonce, sign that nonce with their wallet, then send the signature and the nonce to the indexer operator for verification. Once the signature is confirmed as valid, a valid JWT is produced and returned to the user, and the user is authenticated.
+
+It is important to note that authentication is disabled by default. However, if authentication is enabled, users will need to authenticate before performing operations that involve modifying the state of the service, such as uploading or stopping indexers, etc. The new authentication functionality offers a flexible and secure way for users to authenticate and perform operations that affect the service's state.
```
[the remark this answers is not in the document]
ra0x3: Saw a really funny tweet once about using "etc" after only including one or two examples as a smooth brain slash lazy move 😆

packages/fuel-indexer-lib/src/config/auth.rs
ra0x3: @deekerno 
- You're right that outside the scope of JWT these variable names definitely make no sense
- However, within the scope of JWT - it's convention to name things like this https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-token-structure (will make the doc comment change)

plugins/forc-index/src/ops/forc_index_start.rs
```diff
@@ -109,10 +120,13 @@ pub async fn init(command: StartCommand) -> anyhow::Result<()> {
         }
     }
 
+    println!("{cmd:?}");
```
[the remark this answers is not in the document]
ra0x3: @deekerno 
- It should definitely be behind a verbose flag (as it helps to see what the literal command being run is). I planned to tackle that when I get to https://github.com/FuelLabs/fuel-indexer/issues/650
- Happy to remove it for now if we think it's too noisy

packages/fuel-indexer-api-server/src/uses.rs
ra0x3: @deekerno 
- So I tried your suggestion and turns out the signature (or at least the digest/string that is passed in the auth request)  is larger than the 64 bytes required to build a `Signature`
- @kayagokalp 
  - Am I resolving the `Signature` here properly using the output from `forc-wallet`? That `payload.signature` is the signature output from `forc-wallet` signing
  - This (the implementation that's in the PR) ☝🏽 actually works - though it is flaky

rvmelkonian:
I'm unable to test this as when I `forc-wallet new` I get 
```bash
error: The following required arguments were not provided:
    --namespace <NAMESPACE>
    <PATH>

USAGE:
    forc-wallet new [OPTIONS] --namespace <NAMESPACE> <PATH>
```
but these args are required by the `forc-index` plugin and not mentioned in the wallet docs. Is there some step I'm missing here?

plugins/forc-index/src/ops/forc_index_start.rs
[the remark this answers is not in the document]
deekerno: Nah, we can keep it for now. 👍🏽 it's not a blocker at all.

deekerno (approved):
Everything worked as expected on the first attempt, removal and reversion included.

Super huge value add, great job. 🥇

rvmelkonian:
went through the testing steps and all works as expected. This is a huge feature! We should write some tests soon.

packages/fuel-indexer-api-server/src/api.rs
```diff
@@ -83,20 +91,36 @@ impl From<StatusCode> for ApiError {
...
+                format!("Database error: {e}."),
+            ),
+            ApiError::FuelCrypto(e) => {
+                (StatusCode::BAD_REQUEST, format!("Crypto error: {e}."))
+            }
+            _ => (StatusCode::INTERNAL_SERVER_ERROR, generic_details),
         };
 
-        error!("{:?} - {}", status, err_msg);
+        error!("{status:?} - {details}");
```
rvmelkonian, line 117: 👌

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -116,58 +125,42 @@ pub(crate) async fn health_check(
...
-        tx.send(ServiceRequest::IndexStop(IndexStopRequest {
-            namespace,
-            identifier,
-        }))
-        .await?;
+        error!("Failed to remove Indexer({namespace}.{identifier}): {e}");
 
         return Ok(Json(json!({
-            "success": "true"
+            "success": "false"
```
rvmelkonian, line 148: will this return a `401` error code?

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -3,11 +3,14 @@
 use fuel_indexer_database_types::*;
 use fuel_indexer_lib::utils::sha256_digest;
 use sqlx::{pool::PoolConnection, types::JsonValue, Postgres, Row};
+use std::time::{SystemTime, UNIX_EPOCH};
 use tracing::info;
 
 #[cfg(feature = "metrics")]
 use fuel_indexer_metrics::METRICS;
 
+const NONCE_EXPIRY: u64 = 3600; // 1 hour
```
rvmelkonian, line 12: does this mean that a user requesting the service will have to sign again every hour?

packages/fuel-indexer-lib/src/config/auth.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Issuer.
+    pub iss: String,
+
+    /// Issued at (as UTC timestamp).
+    pub iat: usize,
+
+    /// Expiration time (as UTC timestamp).
+    pub exp: usize,
+}
```
rvmelkonian, line 101: it's a nit, but it would be nicer to work with if these fields were just `subject` `issuer` `issued_at` `expiration` one less double take.

rvmelkonian (approved):

packages/fuel-indexer-database/postgres/src/lib.rs
ra0x3: @rvmelkonian 
- No the nonce is used to authenticate, but is not the authentication token.
- After a user requests a nonce they have an hour to use it or lose it.

rvmelkonian (approved):

PR #631, opened 2023-03-06
enhancement: welcome plugin

### LINKED ISSUE #603: enhancement: create `forc index welcome`
This has come out of this idea [https://github.com/FuelLabs/fuel-indexer/issues/502#issuecomment-1428905571]() but rather than doing it in bash, we've decided to turn this into a plugin. 

Greet a new user with an interactive CLI wizard. Where a user can choose some default settings and configurations so that they can quickly get to grips with the indexer and what it can do. 

The wizard will take them from through creating an indexer and deploying it on `beta-3`.

The wizard should have fun and colourful ASCII art along with friendly and helpful text. 

I went with the command `welcome` as `start` is already taken. Does welcome sound a bit too formal? @ra0x3 @deekerno

rvmelkonian:
closes #603

## Chanelog
• creates `forc_index_welcome.rs` and updates associated files
• adds `owo-colors = "1.3.0"` and `rand = "0.8"` to forc-index .toml
• creates wizard experience in the CLI

## Todo
- [ ] add tree output to show user what has been created

## Testing
Please test and break! There is probably a lot of things we could add here, but I thought I'd keep with essential for the PoC. 
• make sure to have fuel-node and fuel-indexer running
• build the binary `cargo build --release -p forc-index && cp target/release/forc-index ~/.fuelup/bin`
• create a new directory cd into that directory 
• run `forc-index welcome`
• go through the wizard

## Notes
• For some reason when the build command is called, it takes quite a bit a longer to build that if you were doing `forc-index build`, 
• when the build command is run an error is thrown saying `error: Found argument 'run' which wasn't expected, or isn't valid in this context` 
• working on resolving the above but they are not blocking review. 
• currently the welcome program will not work without a fuel-node and the indexer running in separate processes. I did think about adding this in, but it would increase the size of this PR quite a bit. Let me know thoughts on this. Perhaps I could just show a message to the user telling them they need to run those services to proceed.


ra0x3:
@deekerno 

Just discussed this PR w/ @rvmelkonian  and he's interested in possibly merging and iterating on this (so merge and hide it, then iterate). I mentioned that I wanted to get your $0.02 on this. 
- On one hand, we _are_ trying to limit the "merge now iterate later" stuff
- But at the same time, we could always hide this and make some relatively quick iterations
- Thoughts?

rvmelkonian:
Ok great, I will fix the conflicts then re-request review.

rvmelkonian:
@ra0x3 unable to re-request, but its ready for review now.

rvmelkonian:
@deekerno @ra0x3 Just need an approve for this and then it's ready to merge in 🍶

deekerno (approved):
Looks fine to me. I wonder if we should actually hide this though. Won't this be shown as part of the help message? In any case, we can get this in and then go from there.

rvmelkonian:
@deekerno hide behind a feature flag? I can put that in before I merge

deekerno:
That may be a good idea, but we'd have to change the flag on our CI process for the release while we get this to a good point. @ra0x3, do you have any thoughts?

ra0x3:
@deekerno @rvmelkonian 
- So since we're "hiding" this, we can remove the actual command for `ForcIndex` since if we leave it there, it will show in `--help` as @deekerno mentions.
- It's just removing one line of code, so if we want to hook it back up to iterate on it, should be pretty easy

plugins/forc-index/src/cli.rs
```diff
@@ -32,6 +32,7 @@ pub enum ForcIndex {
     Build(BuildCommand),
     Auth(AuthCommand),
     Postgres(ForcPostgresOpt),
+    Welcome(WelcomeCommand),
```
ra0x3, line 35: @rvmelkonian (per my most recent comment responding to @deekerno with regard to hiding the command from `--help` I'm merely suggesting we just comment this (line 35), and line 54 - and possibly add an `#[allow(unused)]` above so that clippy doesn't complain about the import

rvmelkonian: Ah yes, I missed this comment. Will do.

ra0x3 (approved):

PR #633, opened 2023-03-07
chore: add fuel-core to indexer release builds

ra0x3:
### Description
- We should be including `fuel-core-lib` in our release builds

### Testing steps
- This was tested in a previous (now squashed) commit
- If you want to test for yourself, just remove the `if:` condition from the `publish-docker-image` step in this PR

### Changelog
- chore: add fuel-core to indexer release builds


deekerno (approved):
🚢

PR #649, opened 2023-03-14
enhancement: add support for tracking function calls in `Call` receipt

deekerno:
_Note: This PR is pointed at [`rashad/503-add-vec-support`](https://github.com/FuelLabs/fuel-indexer/tree/rashad/503-add-vec-support)._
## Changelog
- Add `abi::Call` struct
-- This struct substitutes the `param1` field in the `Call` receipt for a `fn_name: String` field so that users can see what function is being called
- Add a match structure for mapping function selectors to function names
- Adjust `Receipt::Call` match arm to generate an `abi::Call` struct
- Add test to trigger and index pure Sway function

## Testing Plan
Added new test to E2E test suite. CI should pass.

### Manual Testing Steps
0. Ensure your DB is cleared and that your WASM modules are up-to-date.
1. Run each of the following commands in a different terminal session:
- `cargo run --bin fuel-node`
- `cargo run --bin web-api`
- `cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml`
- `curl -X POST http://127.0.0.1:8000/pure_function`
2. You should see the data for `callentity` being indexed into the database.


packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -244,14 +244,25 @@ fn process_fn_items(
...
+                quote! {
+                    #selector => #ty_id,
+                },
+                quote! {
+                   #selector => #fn_name.to_string(),
+                },
+            )
         })
-        .collect::<Vec<proc_macro2::TokenStream>>();
+        .collect::<Vec<(proc_macro2::TokenStream, proc_macro2::TokenStream)>>();
```
ra0x3, line 260: - With the codgen refactor, we now see that all of this codegen stuff just loops ABIgen TypeDefinitions
- Thankfully we can loop as many times as we want since this is all compile time
- With that said, can we split this out?
  - We're combining two distinct `TokenStream`s under a single `abi_selector_tuple_vec`  (btw I don't think you need to add "vec" suffix here since the code shows that it's clearly a `Vec` 😄 )
  - Maybe leave `abi_selectors` as is, and split out the `selector => fn_name` to `fn_name_selectors` (or something similar?) as its own separate loop?

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -244,14 +244,25 @@ fn process_fn_items(
...
+
+    let (abi_selectors, abi_selectors_to_fn_names): (
```
ra0x3, line 262: See comment above

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -449,10 +460,25 @@ fn process_fn_items(
 
                     for receipt in tx.receipts {
                         match receipt {
-                            Receipt::Call { param1, to: id, ..} => {
+                            Receipt::Call { id: contract_id, amount, asset_id, gas, param1, to: id, .. } => {
                                 #contract
+                                fn selector_to_fn_name(sel: u64) -> String {
```
ra0x3, line 465: Maybe we could move this to the decoder? Since that's where all the other match logic is?

ra0x3 (commented):
@deekerno 
- Functionality looks OK
- Left some comments about making things a bit cleaner

packages/fuel-indexer-macros/src/indexer.rs
deekerno: Sure, I can do that. I was going for _elegance and speed_, but you're right that it's at compile time anyways and the operation is hardly going to add that much time to the process.

ra0x3 (approved):

PR #652, opened 2023-03-15
chore: remove sqlx query_as 

### LINKED ISSUE #645: Remove all sqlx `query_as!` logic
- This is an example of using `query_as!` [here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-database/postgres/src/lib.rs#L320)
- We want to swap all this `query_as!` with the boring/manual way to resolve the query into a type
- That boring/manaul method is shown [here](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-database/postgres/src/lib.rs#L580)
- Why? The `query_as!` method is flaky, and is not worth the trouble for the tiny bit of time it saves
  - Further `query_as!` requires keeping track of this "sqlx-data.json" file that is super confusing and inconsistent?
  - There's really no excuse to be using this unecessary abstraction when we can just use basic getters (as demonstrated in the boring/manual approach) and avoid all of this rigmarole

rvmelkonian:
closes #645 

## Changelog 
• Removes all occurrences of sqlx macros in `postgres/src/lib.rs`
• Pass the col name as a string arg into `.get()`
• Uses the `.bind()` method instead of creating a query string variable and interpolating with `format!`. 

## Testing
All tests are passing locally

## Notes
Regarding the second point in the changelog, I went ahead with this as it seemed more self documenting and less error prone than the old way with `.get(1)`. 

If we prefer this syntax, i'll open another PR to refactor the other places where `.get(1)` is still being used.


ra0x3:
@rvmelkonian 
- So for this work, we want to remove _all_ macro references. 
- No `query_as!` no `query_scalar!`. 
- We want just plain/simple/boring getters (e.g., `.get(0)`) where possible

rvmelkonian:
@ra0x3 yep, I just realised I misunderstood the original issue, I couldn't understand why we would not want these macros. But after working with them, I can certainly see why. Making these changes now!

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -92,16 +92,23 @@ pub async fn root_columns_list_by_id(
...
-        RootColumns,
-        r#"SELECT
-               id AS "id: i64", root_id AS "root_id: i64", column_name, graphql_type
-           FROM graph_registry_root_columns
-           WHERE root_id = $1"#,
-        root_id
-    )
-    .fetch_all(conn)
-    .await
+    let rows =
```
ra0x3, line 95: Just a style nit. This could be chained into one single `Ok`
```suggestion
    Ok(sqlx::query("SELECT * FROM graph_registry_root_columns WHERE root_id = $1")
            .bind(root_id)
            .fetch_all(conn)
            .await?
            .into_inter()
            .map(|row| RootColumns {
              id: row.get("id"),
              root_id: row.get("root_id"),
              column_name: row.get("column_name"),
              graphql_type: row.get("graphql_type"),
        })
        .collect::<Vec<Rootcolumns>>())
            
```

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -162,30 +169,30 @@ pub async fn graph_root_latest(
...
-    let schema_identifier: String = row.get(5);
-
     Ok(GraphRoot {
-        id,
-        version,
-        schema_name,
-        query,
-        schema,
-        schema_identifier,
+        id: row.get("id"),
```
ra0x3, line 190: Meh, I did prefer the previous version here as it made the types explicit and looked cleaner. @deekerno thoughts?

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -198,29 +205,36 @@ pub async fn type_id_list_by_name(
...
+    let rows = sqlx::query(
+        "SELECT 
+            id, 
+            schema_version, 
+            schema_name, 
+            table_name, 
+            graphql_name, 
+            schema_identifier
+        FROM graph_registry_type_ids
+        WHERE schema_name = $1 AND schema_version = $2 AND schema_identifier = $3",
```
ra0x3, line 217: @rvmelkonian 
- I see we switched a lot of the string formatting to `.bind()`
- That's cool if these values are more complex types (e.g. a byte array) but for a simple string, is there any reason we're using `.bind()` over the string formatting?

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -198,29 +205,36 @@ pub async fn type_id_list_by_name(
...
-            schema_identifier,
-            schema_version,
-            schema_name,
-            table_name,
-            graphql_name,
-        }
-    }).collect::<Vec<TypeId>>())
+    .await?;
+
+    let type_ids: Vec<TypeId> = rows
```
ra0x3, line 225: Another style comment here, no real reason to not chain all of this


```rust
    Ok(sqlx::query(
        "SELECT 
            id, 
            schema_version, 
            schema_name, 
            table_name, 
            graphql_name, 
            schema_identifier
        FROM graph_registry_type_ids
        WHERE schema_name = $1 AND schema_version = $2 AND schema_identifier = $3",
    )
    .bind(namespace)
    .bind(version)
    .bind(identifier)
    .fetch_all(conn)
    .await?
    .into_iter()
    .map(|row| TypeId {
            id: row.get("id"),
            schema_version: row.get("schema_version"),
            schema_name: row.get("schema_name"),
            table_name: row.get("table_name"),
            graphql_name: row.get("graphql_name"),
            schema_identifier: row.get("schema_identifier"),
      })
      .collect::<Vec<TypeId>>())
```

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -319,7 +334,27 @@ pub async fn list_column_by_id(
...
+        "SELECT id, type_id, column_position, column_name, column_type, nullable, graphql_type FROM graph_registry_columns WHERE type_id = $1"
+        )
+        .bind(col_id)
+        .fetch_all(conn)
+        .await?;
+
+    let columns = rows
+        .into_iter()
+        .map(|row| Columns {
+            id: row.get("id"),
```
ra0x3, line 347: Again here, I think we should do the boring thing and just extract the values (while making the types explicit) _then_ create the struct.


```rust
let value: String = row.get(0);
let MyThing { value };
```

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -331,27 +366,34 @@ pub async fn columns_get_schema(
...
+            c.column_position as column_position,
+            c.column_name as column_name,
+            c.column_type as column_type
+        FROM graph_registry_type_ids as t
+        INNER JOIN graph_registry_columns as c ON t.id = c.type_id
+        WHERE t.schema_name = $1 AND t.schema_identifier = $2 AND t.schema_version = $3
+        ORDER BY c.type_id, c.column_position
+    ";
+
+    let rows = sqlx::query(query)
```
ra0x3, line 382: Same comment on chaining here

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -633,13 +687,15 @@ pub async fn index_id_for(
...
+    let row = sqlx::query(
+        r#"select id from index_registry where namespace = $1 and identifier = $2"#,
+    )
+    .bind(namespace)
+    .bind(identifier)
+    .fetch_one(conn)
+    .await?;
 
-    let id: i64 = row.get(0);
+    let id: i64 = row.get("id");
```
ra0x3, line 698: This is the only place we're referencing by field name instead of field position. Any reason this is done differently?

ra0x3 (changes requested):

packages/fuel-indexer-database/postgres/src/lib.rs
rvmelkonian: Ok

packages/fuel-indexer-database/postgres/src/lib.rs
rvmelkonian: the main reason is to prevent SQL injection, `format!` doesn't guarantee that kind of safety, also .bind() allows sqlx to do type checking on the values being passed to the query.

packages/fuel-indexer-database/postgres/src/lib.rs
deekerno: I agree. I'm surprised that it actually worked because `row.get(..)` has never worked any time that I've excluded the type annotation. 🤔

packages/fuel-indexer-database/postgres/src/lib.rs
deekerno: For consistency, you can do the individual `row.get(..)`s with the type annotations inside of a map closure and have the closure return a `RootColumns` struct.

packages/fuel-indexer-database/postgres/src/lib.rs
rvmelkonian: I was trying to save memory by not initialising variables where we can just pass the value in directly, but the performance difference here is probably minimal. To my eyes the latter looks cleaner, as there are less lines to read and it expresses the same logic, but I don't mind at all, I'll switch to the former way!

deekerno: Yeah, I know we don't _always_ annotate types but I think it's good practice to do so, especially in these queries where we're pulling things out of the raw result.

ra0x3: @rvmelkonian 

Memory-wise these 2 things are the same

```
# 1
let bar = Bar{ foo: row.get(0) };

# 2
let foo: String = row.get(0);
let bar = Bar{ foo };
```

In the second example `Bar` owns `foo` so no memory is copied. Number 2 is just more explicit

rvmelkonian: you're absolutely right, It's my cursed golang brain kicking in!

rvmelkonian:
Ok @ra0x3 thanks for the review, all your feedback has been integrated. The only thing I wasn't sure of is if we prefer `.get()` by position or name. I opted for by name as it makes the query less brittle, if for example future migrations change the ordering of column positions.

packages/fuel-indexer-database/postgres/src/lib.rs
rvmelkonian: I'm not sure exactly what you mean. But now I'm passing in the name arg and declare a variable with explicit types. Maybe you were (rightly) pointing my lack of consistency in this old commit!

rvmelkonian: using positional args always in latest commit.

ra0x3 (dismissed):
✅ But let's hold off on merge until #630 is in 🙏🏽

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -162,30 +172,30 @@ pub async fn graph_root_latest(
...
 
     let id: i64 = row.get(0);
     let version: String = row.get(1);
     let schema_name: String = row.get(2);
-    let query: String = row.get(3);
-    let schema: String = row.get(4);
-    let schema_identifier: String = row.get(5);
+    let schema_identifier: String = row.get(3);
+    let query: String = row.get(4);
+    let schema: String = row.get(5);
```
deekerno, line 190: Is there a reason that the order was changed here? This is in the incorrect order and will break querying in the GraphQL API.

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -198,29 +208,36 @@ pub async fn type_id_list_by_name(
...
+    .into_iter()
+    .map(|row| {
         let id: i64 = row.get(0);
         let schema_version: String = row.get(1);
         let schema_name: String = row.get(2);
-        let table_name: String = row.get(3);
+        let schema_identifier: String = row.get(3);
         let graphql_name: String = row.get(4);
-        let schema_identifier = row.get(5);
+        let table_name: String = row.get(5);
```
deekerno, line 229: Same comment regarding order and GraphQL breakage.

deekerno (changes requested):
Looks like the GraphQL API E2E tests broke. Left a couple comments as to why.

packages/fuel-indexer-database/postgres/src/lib.rs
rvmelkonian: I copied out the order in the struct 
```
#[derive(Debug)]
pub struct GraphRoot {
    pub id: i64,
    pub version: String,
    pub schema_name: String,
    pub schema_identifier: String,
    pub query: String,
    pub schema: String,
}
```
apart from referring to master, is there another way I should know what the correct order would be? Just wondering if there is some piece I'm not seeing here.

rvmelkonian: I know these fields are not ordered, but it was an attempt. Albeit a fallible one!

deekerno: Unfortunately, you'd probably have to look at the migrations and determine the order as new columns would be added to the end. Alternatively, you could also do a `SELECT` against the table to see the structure. These `query_as!` macros did a lot of magic under the hood to match the columns and such, so we just have to make sure that we're checking the order.

rvmelkonian: migrations as source of truth sounds good. just updating these now. Thanks!

deekerno (dismissed):
👏🏽

ra0x3 (dismissed):

ra0x3 (approved):
:shipit:

PR #653, opened 2023-03-15
ci: use buildjet runners

ra0x3:
### Description
- This PR updates our CI to use the new buildjet runners
- fuel-core has already updated to this https://github.com/FuelLabs/fuel-core/pull/1017
- Updates the `cargo-toml-fmt-check` step to use `cargo sort`

### Testing steps
- [ ] CI should pass (and be faster! 🚗💨)

### Changelog
- ci: use buildjet runners


ra0x3:
@deekerno I don't think they have. No reason to believe it won't work for release though (famous last words)

PR #655, opened 2023-03-15
chore: enable sparse protocol in CI

### LINKED ISSUE #654: chore: enable sparse protocol
Sparse protocol has been enabled in the new version of rust, more info on this feature [here](https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html). 

This should speed up our CI significantly, also it's just been enabled in fuels-rs [here](https://github.com/FuelLabs/fuels-rs/pull/888). Let's give it a whirl?

rvmelkonian:
closes #654 

According to [this](https://github.com/FuelLabs/fuels-rs/pull/888#issuecomment-1469491450), all we need to do is bump the version to enable sparse protocol in the CI. See the linked issue about this new rust feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ env:
   RUSTFLAGS: -D warnings
   REGISTRY: ghcr.io
   SQLX_OFFLINE: true
-  RUSTC_VERSION: 1.67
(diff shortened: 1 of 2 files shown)


rvmelkonian:
closing this for now as github CI runner is still on `1.67` will try these when they update

PR #625, opened 2023-02-27
enhancement: add vec support

ra0x3:

PR #658, opened 2023-03-15
chore: add function to get test contract ID from binary file

deekerno:
## Changelog
- Add function to get the contract ID of test contract from compiled binary
- Set `contract_id` field in manifest to be empty and adjust tests to fill it at test time
-- I'm okay with doing this as we've added the "make sure you change your contract ID" blurb into the docs refresh
- Remove unused testing functions
- Remove `CURRENT_TEST_CONTRACT_ID_STR` constant
- Adjust `hello-world-data` to not use a hard-coded contract ID

## Testing Plan
Adjustments were made to the E2E tests; CI should pass.

The `hello-world-data` package was adjusted to remove a hard-coded contract ID string. To ensure that it works, run the following steps:

0. Ensure your DB is cleared.
1. `cargo run --bin fuel-node -- --contract-bin examples/hello-world/contracts/greeting/out/debug/greeting.bin`
2. `cargo run --bin fuel-indexer -- run --manifest examples/hello-world/hello-index/hello_index.manifest.yaml`
3. `cargo run --bin hello-world-data -- --host 0.0.0.0:4000`
4. Ensure that the test data was successfully indexed.

## Notes
With this PR, we no longer have to adjust the contract ID fields in multiple places when making adjustments to the test contract.

### DIFF
--- a/packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
+++ b/packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/components/indices/fuel-indexer-test/schema/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id: fuel1dthcclayuwhce7t7jt8dlpvv2c0q4wsyj4xsudcek5gwc6nd5cnqj4qkua
+contract_id:
 identifier: index1
 module:
   wasm: target/wasm32-unknown-unknown/release/fuel_indexer_test.wasm
--- a/packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
+++ b/packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/components/indices/fuel-indexer-test/schema/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id:
+contract_id: ~
 identifier: index1
 module:
   wasm: target/wasm32-unknown-unknown/release/fuel_indexer_test.wasm
--- a/packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
+++ b/packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/components/indices/fuel-indexer-test/schema/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id: ~
+contract_id: fuel1dthcclayuwhce7t7jt8dlpvv2c0q4wsyj4xsudcek5gwc6nd5cnqj4qkua
 identifier: index1
 module:
   wasm: target/wasm32-unknown-unknown/release/fuel_indexer_test.wasm
--- a/packages/fuel-indexer-tests/src/fixtures.rs
+++ b/packages/fuel-indexer-tests/src/fixtures.rs
@@ -1,4 +1,3 @@
-use crate::defaults::CURRENT_TEST_CONTRACT_ID_STR;
 use crate::{defaults, WORKSPACE_ROOT};
 use axum::routing::Router;
 use fuel_indexer::IndexerService;
@@ -17,10 +16,7 @@ use fuels::{
     },
     signers::Signer,
 };
-use sqlx::{
-    pool::{Pool, PoolConnection},
-    Postgres,
-};
+use sqlx::{pool::Pool, Postgres};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use tracing_subscriber::filter::EnvFilter;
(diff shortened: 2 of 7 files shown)


packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
```diff
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/components/indices/fuel-indexer-test/schema/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id: fuel1dthcclayuwhce7t7jt8dlpvv2c0q4wsyj4xsudcek5gwc6nd5cnqj4qkua
+contract_id:
```
rvmelkonian, line 5: in other places where we have a null field in yaml I've seen `~`, do we want that here?

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -163,57 +143,20 @@ pub async fn setup_example_test_fuel_node() -> Result<(), ()> {
...
+        .join("out")
+        .join("debug")
+        .join("fuel-indexer-test.bin");
 
-    let mut wallet =
-        WalletUnlocked::load_keystore(wallet_path, defaults::WALLET_PASSWORD, None)
+    let compiled =
+        Contract::load_contract(contract_bin_path.as_os_str().to_str().unwrap(), &None)
             .unwrap();
+    let (id, _) = Contract::compute_contract_id_and_state_root(&compiled);
```
rvmelkonian, line 157: 👌 our new favourite SDK method

rvmelkonian (dismissed):
This is much cleaner. All unit tests and e2e tests passing.

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
deekerno: I can try it, but I think it may need to be empty in order for `#contract_conditional` in `indexer.rs` to not try to parse it into a contract ID. With it being empty, the WASM module will listen to all contracts anyways.

@ra0x3, thoughts?

ra0x3: @deekerno 
- `~` is `null` in YAML so the `if let Some(contract_id)` part of the codegen should never be hit.
- If we can let's try to use explicit null `~` instead of implied null (blank space)

ra0x3 (commented):
@deekerno 
- Does any of this prevent us from having to hard-code contract IDs?
- Also let's put this in after #625

deekerno:
@ra0x3 Yep, with this change, we no longer have to hard-code any contract ID for the tests. Actually, I can probably go one step further and do the same thing for the `hello-world` example as I see we have a hard-coded string there too.

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
deekerno: @ra0x3 Using the explicit null worked.

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
```diff
@@ -2,7 +2,7 @@ namespace: fuel_indexer_test
 graphql_schema: packages/fuel-indexer-tests/components/indices/fuel-indexer-test/schema/fuel_indexer_test.graphql
 abi: packages/fuel-indexer-tests/contracts/fuel-indexer-test/out/debug/fuel-indexer-test-abi.json
 start_block: 1
-contract_id: fuel1dthcclayuwhce7t7jt8dlpvv2c0q4wsyj4xsudcek5gwc6nd5cnqj4qkua
+contract_id: ~
```
ra0x3, line 5: - Remind me why we're removing this? 
- Keeping this tests the `#contract_conditional` in the codegen
- If we remove this and (for whatever reason) the `#contract_conditional` breaks, we won't see it

ra0x3 (commented):
@deekerno
- This change affects our examples too I believe? But I don't see testing examples in the testing steps (or any mention of the examples at all)?
- Can you confirm that all examples work as they did previously ?

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
deekerno: > @deekerno
> 
>     * This change affects our examples too I believe? But I don't see testing examples in the testing steps (or any mention of the examples at all)?
> 
>     * Can you confirm that all examples work as they did previously ?

Ah, my apologies. Yes, the `hello-world` example works as expected. I'll add some testing steps.

As far as removing it, it was to get rid of the hard-coded contract strings everywhere. I suppose leaving it there is fine, as we don't (yet) have a way to programmatically compile a WASM blob to test out the contract conditional.

ra0x3: @deekerno 
- Thanks, keep in mind that while we do want to use as few hard-coded strings as possible, keeping the contract_id in the manifest is actually how the indexer works 😄
- So I wouldn't consider putting the contract ID in the manifest as using magic strings
- (We will have to update it when the contract changes, but that's only annoying cause we're making changes (as devs) often. Users won't really experience that pain like we do)

deekerno: I guess too much optimization _can_ be a bad thing. 😅

packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
```diff
@@ -108,6 +109,7 @@ async fn test_can_trigger_and_index_callreturn_postgres() {
     let mut srvc = indexer_service_postgres().await;
     let mut manifest: Manifest =
         serde_yaml::from_str(assets::FUEL_INDEXER_TEST_MANIFEST).expect("Bad yaml file.");
+    manifest.contract_id = Some(get_test_contract_id().to_string());
```
ra0x3, line 112: @deekerno 
- Maybe we can tuck this into `update_test_manifest_asset_paths` ?
- We technically don't need to set the contract_id but for parity (given the last comment I left about including the contract_id) we can keep it
- `update_test_manifest_asset_paths` is a hack to update the manifest, so might as well tuck this there (as one line) rather than spreading it out. Thoughts? 🧠

ra0x3 (approved):

PR #659, opened 2023-03-16
chore: use verbose in plugins

### LINKED ISSUE #650: Hide plugin logging/print behind `-v --verbose` flag(s)
- We have several commands that print output to stdout
- We should hide those print/log statements behind the `-v --verbose` flag(s)
- Most commands have `-v --verbose` flags already but I don't think they're used?
- Unless the message is _critical_ don't send it to stdout unless one of these flags is passed

rvmelkonian:
closes #650 

## Changelog
• adds `verbose` flags to command structs
• adds if blocks to print more detailed output if verbose

## Testing 
• Rebuild the plugins binary. 
• Run through all the commands, using verbose and then not using verbose.


plugins/forc-index/src/ops/forc_index_auth.rs
```diff
@@ -95,7 +95,7 @@ pub fn init(command: AuthCommand) -> anyhow::Result<()> {
     let response: SignatureResponse = res.json().unwrap();
 
     if let Some(token) = response.token {
-        info!(
+        println!(
```
ra0x3, line 98: So this issue is 2 parts:
1. We want to hide (most) output behind a `-v --verbose` flag
2. The existing info that we're logging (e.g., this `print`) we want to make less verbose
  - This `print` is a _really_ long informative string. That's ok if `-v --verbose` is passed, but if not, it should just say something like `"✅ Authenticated successfully"`
  - Also this should remain `info` and not `println`

plugins/forc-postgres/src/ops/forc_postgres_createdb.rs
```diff
@@ -159,7 +159,7 @@ pub async fn init(command: CreateDbCommand) -> anyhow::Result<()> {
 
     pb.finish();
 
-    info!("\n✅ Successfully created database at '{pg_db_uri}'.");
+    println!("\n✅ Successfully created database at '{pg_db_uri}'.");
```
ra0x3, line 162: Same comment here regarding shortening existing commands if `- --verbose` is not set. Something like `✅ Database created` is fine if no verbosity is passed

ra0x3 (changes requested):
@rvmelkonian 
- I think we might have had different ideas on this. 
- When creating this issue I think my idea was for something a bit more simple. Like quite literally


```rust
if command.verbose {
  info!("Some long verbose/informative statement saying what exactly happened");
} else {
  info!("Short really statement");
  // Keeping in mind that these short statements would only apply to 
  // operations where I need to know if something failed/succeed
}
```
- In general we want to try to see what `forc` uses and use that (e.g., their TracingSubsriberSetup)
- As an example, when I'm running `forc index build` (which does  a lot under the hood), if I pass `-v --verbose` I want to see _all_ the output, all the technical details in the console
  - However, if I run `forc index build` w/o passing `-v --verbose`, I just wanna see (1) that it's `"Building..."`, then just whether the build succeeded (not a ton of info, but I still get some messages)
  - This applies to other commands like revert, remove, deploy as well
- I wouldn't conflate verbosity `-v --verbose` with log level 
- Lmk if any of this doesn't make sense, happy to sync on this as well cc @deekerno

rvmelkonian:
@ra0x3 ok understood. I actually did it the simple way first, then went on with this Logger idea. I will implement the requested changes.

plugins/forc-index/src/ops/forc_index_init.rs
```diff
@@ -185,9 +188,12 @@ pub fn init(command: InitCommand) -> anyhow::Result<()> {
         );
     }
 
-    debug!("\n✅ nSuccessfully created index {project_name}");
+    if verbose {
+        info!("\n✅ Successfully created index {project_name}");
+    } else {
+        info!("\n✅ Successfully created index");
+    }
```
deekerno, line 195: s/index/indexer to comply with our recent terminology changes.

deekerno (changes requested):
Looks good for the most part, just left some nits on logic and terminology.

plugins/forc-index/src/ops/forc_index_remove.rs
```diff
@@ -55,12 +60,16 @@ pub fn init(command: RemoveCommand) -> anyhow::Result<()> {
         .json::<Map<String, Value>>()
         .expect("Failed to read JSON response.");
 
-    println!("\n{}", to_string_pretty(&res_json)?);
+    info!("\n{}", to_string_pretty(&res_json)?);
```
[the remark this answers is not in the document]
rvmelkonian: we do indeed!

plugins/forc-index/src/ops/forc_index_remove.rs
```diff
@@ -32,16 +33,20 @@ pub fn init(command: RemoveCommand) -> anyhow::Result<()> {
         headers.insert(AUTHORIZATION, auth.parse()?);
     }
 
-    info!(
-        "\n🛑 Removing index '{}.{}' at {target}",
-        &manifest.namespace, &manifest.identifier
-    );
+    if verbose {
+        info!(
+            "\n🛑 Removing indexer'{}.{}' at {target}",
```
ra0x3, line 38: Literally, perfect 🏅

plugins/forc-index/src/ops/forc_index_start.rs
```diff
@@ -120,7 +120,7 @@ pub async fn init(command: StartCommand) -> anyhow::Result<()> {
         }
     }
 
-    println!("{cmd:?}");
+    info!("{cmd:?}");
```
ra0x3, line 123: So for this, and _all_ `Command`s, could we only `info!` log them if `-v --verbose` is set?

plugins/forc-index/src/cli.rs
```diff
@@ -21,7 +21,7 @@ struct Opt {
 }
 
 #[derive(Subcommand, Debug)]
-enum ForcIndex {
+pub enum ForcIndex {
```
ra0x3, line 24: Does this still need to be `pub` ?

ra0x3 (changes requested):
@rvmelkonian Looks really good thanks so much for this, left two or 3 minor comments

plugins/forc-index/src/ops/forc_index_start.rs
rvmelkonian: Sure

rvmelkonian: by the looks of it we only `info!("{cmd:?}")` in `forc_index_start` and in no other ops file.

ra0x3 (approved):
⚓

deekerno (dismissed):
🚢

ra0x3 (changes requested):
@rvmelkonian 
- There are errant .wasm files in this PR now
- This has actually been happening to me too and we should figure out why these files are not being removed by the test script (not in this PR obviously)
- If remove those I'll approve

rvmelkonian:
@ra0x3 
> Edit: Ah I see, the rm -fv does not have the proper filename (thus the files aren't removed) I'll fix in my PR

Was the PR you were referring to merged in master?

rvmelkonian:
@ra0x3 the errant wasm binaries are removed now.

deekerno (approved):

ra0x3 (approved):

PR #660, opened 2023-03-16
chore: introduce ephemeral DBs for e2e tests

deekerno:
## Changelog
- Add `FromStr` implementation for `DatabaseConfig`
- Add `DbUrlParseError` to `IndexerConfigError`
- Add `TestPostgresDb` fixture
-- `new()`: Uses the `DATABASE_URL` environment variable to connect to the `postgres` user on a Postgres server. It then generates a 10-character string and creates a database with that name
-- `teardown()`: Reconnects to the Postgres server, terminates all connections to the unique database, and then drops said database
- Add custom `Drop` implementation for `TestPostgresDb`
-- `drop()` calls `teardown()` by spawning a blocking thread as destructors cannot be run asynchronously
- Adjusts service and API server fixtures to take an optional DB connection string; if absent, the default database values will be used, just as before.
- Adjust Postgres E2E and integration tests to use `TestPostgresDb`, which isolates the data for each test

## Testing Plan
CI should pass as it uses the same `DATABASE_URL` environment variable that is checked in the DB construction.

I repeatedly ran the E2E and integration tests and ensured that all testing databases are in fact ephemeral. You can test this by doing the following (if you have Postgres installed through homebrew):
1. `psql -l` - this will list all of the databases on the server
```
❯ psql -l
                             List of databases
   Name    |  Owner   | Encoding | Collate | Ctype |   Access privileges
-----------+----------+----------+---------+-------+-----------------------
 postgres  | deekerno | UTF8     | C       | C     |
 template0 | deekerno | UTF8     | C       | C     | =c/deekerno          +
           |          |          |         |       | deekerno=CTc/deekerno
 template1 | deekerno | UTF8     | C       | C     | =c/deekerno          +
           |          |          |         |       | deekerno=CTc/deekerno
(3 rows)
```
2. In your terminal session (or in your shell config), make sure that the `DATABASE_URL` environment variable is set to use a user that has the `CREATEDB` permission. For example, my Postgres server is on my local host and default port and has a `postgres` superuser with `my-secret` as a password.
```
export DATABASE_URL="postgres://postgres:my-secret@localhost:5432"
```
3. Run the E2E tests: `cargo test --features e2e,postgres e2e -- --test-threads=1`. They should pass. Feel free to run the integration tests as well: `cargo test --features e2e,postgres integration -- --test-threads=1`
4. Run `psql -l` again; there should be no extra databases and the output should be the same as the original.

## Notes
Eventually, I'd like to do one more adjustment and spawn the indexer service/API server through `lazy_static!`. With some adjustments to the `Drop` behaviors of those structs and changes to the tests themselves, that would allow for us to run the entire workspace test suite in parallel without having to exclude `fuel-indexer-tests` for single threading (that means one less CI job). But again, _eventually_.


packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -160,16 +270,15 @@ pub fn get_test_contract_id() -> Bech32ContractId {
     Bech32ContractId::from(id)
 }
 
-pub async fn api_server_app_postgres() -> Router {
+pub async fn api_server_app_postgres(database_url: Option<&str>) -> Router {
+    let database: DatabaseConfig = if database_url.is_some() {
+        DatabaseConfig::from_str(database_url.unwrap()).unwrap()
+    } else {
+        DatabaseConfig::default()
+    };
```
rvmelkonian, line 278: if this is not set `export DATABASE_URL="postgres://postgres:my-secret@localhost:5432"` the password will be set from the default const value `"postgres"` and I think an error will be thrown.  Maybe we should just hardcode a pw for the ephemeral testdb?

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -16,29 +17,138 @@ use fuels::{
...
+            user,
+            password,
+            host,
+            port,
+            database: db_name.clone(),
+        };
+
+        let mut conn = PgConnection::connect(server_connection_str.as_str())
+            .await
+            .unwrap();
```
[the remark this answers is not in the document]
deekerno: Oops, don't know how that made it past me. 😅

packages/fuel-indexer-tests/src/fixtures.rs
rvmelkonian: This makes sense.

packages/fuel-indexer-lib/src/config/mod.rs
```diff
@@ -34,6 +34,8 @@ pub enum IndexerConfigError {
     ConfigFileError(#[from] Error),
     #[error("Error processing YAML file: {0:?}")]
     SerdeYamlError(#[from] serde_yaml::Error),
+    #[error("Error parsing database connection string into DatabaseConfig")]
```
ra0x3, line 37: If we use `url` to parse the connection URI we might be able to derive `url::Error` here

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -16,29 +17,142 @@ use fuels::{
...
+
+        conn.execute(
+            format!(
+                r#"DROP DATABASE IF EXISTS "{}" WITH (FORCE);"#,
+                self.db_name
+            )
+            .as_str(),
+        )
+        .await
+        .unwrap_or_else(|_| {
```
ra0x3, line 139: Could we maybe print these errors?

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -16,29 +17,142 @@ use fuels::{
...
-    };
-    match IndexerConnectionPool::connect(&config.to_string())
+pub struct TestPostgresDb {
+    pub db_name: String,
+    pub url: String,
+    pub pool: Pool<Postgres>,
+    server_connection_str: String,
+}
+
+impl TestPostgresDb {
```
ra0x3, line 41: There's a lot going on here, maybe a comment mentioning what this is doing?

packages/fuel-indexer-tests/tests/e2e/graphql_api_postgres.rs
```diff
@@ -71,8 +73,9 @@ async fn test_can_return_query_response_with_all_fields_required_postgres() {
 async fn test_can_return_query_response_with_nullable_fields_postgres() {
     let fuel_node_handle = tokio::spawn(setup_example_test_fuel_node());
 
-    let mut srvc = indexer_service_postgres().await;
-    let api_app = api_server_app_postgres().await;
+    let test_db = TestPostgresDb::new().await;
+    let mut srvc = indexer_service_postgres(Some(&test_db.url)).await;
```
ra0x3, line 77: This logic of `indexer_service_postgres(Some(&test_db.url))` and `api_server_app_postgres(Some(&test_db.url))` with the `Some()` arg seems pretty repetitive, is there some way we can hide that logic given that _all_ of these take the exact same arg of `Some(&test_db.url)`

deekerno:
@ra0x3 re: `fuel-indexer-tests/tests/integration/web_api_postgres` -- I didn't combine the testing component setup into a single function like the other test modules as they each use different parts; we can come back and clean it up later if we like.

packages/fuel-indexer-lib/src/config/database.rs
```diff
@@ -156,7 +141,10 @@ impl FromStr for DatabaseConfig {
                     database: database.to_string(),
                 })
             }
-            _ => Err(IndexerConfigError::DbUrlParseError),
+            _ => {
+                println!("Unsupported database. Please check your database URL.");
+                Err(ParseError::InvalidIpv4Address)
```
ra0x3, line 146: Is this really an `ParseError::InvalidIpv4Address` error thought? Maybe `unimplemented!()` ?

packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
```diff
@@ -439,8 +438,7 @@ async fn test_can_trigger_and_index_scriptresult_event_postgres() {
 async fn test_can_trigger_and_index_transferout_event_postgres() {
     let fuel_node_handle = tokio::spawn(setup_example_test_fuel_node());
 
-    let test_db = TestPostgresDb::new().await;
-    let mut srvc = indexer_service_postgres(Some(&test_db.url)).await;
+    let (test_db, mut srvc) = setup_test_components().await;
```
ra0x3, line 441: Looks much much cleaner 👏🏽

ra0x3 (commented):
@deekerno Isn't there a CI change here as well since we're auto-dropping the DB after each run?

packages/fuel-indexer-lib/src/config/database.rs
deekerno: Ooh, `unimplemented!()` is much better.

deekerno:
> Isn't there a CI change here as well since we're auto-dropping the DB after each run?

No CI change here as the `DATABASE_URL` env var in CI points to a Postgres server but no chosen database. What this does is that it creates a temp database on that server which the test uses, and then it drops the temp database when the test goes out of scope.

packages/fuel-indexer-lib/src/config/database.rs
```diff
@@ -111,3 +114,37 @@ impl From<DatabaseConfig> for Uri {
...
+                Ok(DatabaseConfig::Postgres {
+                    user: user.to_string(),
+                    password: password.to_string(),
+                    host: host.to_string(),
+                    port: port.to_string(),
+                    database: database.to_string(),
+                })
+            }
+            _ => {
+                println!("Unsupported database. Please check your database URL.");
```
ra0x3, line 145:
```suggestion
                unimplemented!("Unsupported database. Please check your database URL.");
```

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -160,16 +306,15 @@ pub fn get_test_contract_id() -> Bech32ContractId {
     Bech32ContractId::from(id)
 }
 
-pub async fn api_server_app_postgres() -> Router {
+pub async fn api_server_app_postgres(database_url: Option<&str>) -> Router {
+    let database: DatabaseConfig = if database_url.is_some() {
```
ra0x3, line 310:
```suggestion
    let database = database_url.map(|x| DatabaseConfig::from_str(x).unwrap()).unwrap_or_else(|| DatabaseConfig::default());
```

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -184,16 +329,15 @@ pub async fn api_server_app_postgres() -> Router {
     GraphQlApi::build(config, pool, None).await.unwrap()
 }
 
-pub async fn authenticated_api_server_app_postgres() -> Router {
+pub async fn authenticated_api_server_app_postgres(database_url: Option<&str>) -> Router {
+    let database: DatabaseConfig = if database_url.is_some() {
```
ra0x3, line 333:
```suggestion
    let database = database_url.map(|x| DatabaseConfig::from_str(x).unwrap()).unwrap_or_else(|| DatabaseConfig::default());
```

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -214,16 +358,15 @@ pub async fn authenticated_api_server_app_postgres() -> Router {
     GraphQlApi::build(config, pool, None).await.unwrap()
 }
 
-pub async fn indexer_service_postgres() -> IndexerService {
+pub async fn indexer_service_postgres(database_url: Option<&str>) -> IndexerService {
+    let database: DatabaseConfig = if database_url.is_some() {
```
ra0x3, line 362:
```suggestion
    let database = database_url.map(|x| DatabaseConfig::from_str(x).unwrap()).unwrap_or_else(|| DatabaseConfig::default());
```

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -16,29 +17,174 @@ use fuels::{
...
+        let test_db_config = DatabaseConfig::Postgres {
+            user,
+            password,
+            host,
+            port,
+            database: db_name.clone(),
+        };
+
+        // Connect directly to the Postgres server and create a database with the unique string
+        let mut conn = match PgConnection::connect(server_connection_str.as_str()).await {
```
ra0x3, line 78: @deekerno 
- For this entire `TestPostgresDb::new` function, I realize this is testing code (thus not as picky with it). 
- But as I'm looking over it, maybe we should use the same error pattern as we use in the service code? 
- Meaning, in this test code we're using a lot of `expect` and `panic`. I'm thinking maybe we can just `?` a lot of this stuff?
  - Maybe with some `TestError` enum that uses `thiserror` similar to how `IndexerError` does? Thoughts?
    - Thinking is, since this PR doesn't really change _a ton_ about how the testing works (we still have to use a single-threaded setup), a real benefit/side-effect of this PR can be cleaning up our testing setup

deekerno: I'm down for that. I was doing some work to get multi-threading to work and also came to the conclusion that we could do some cleanup in places. I'll work on this with the other suggestions, and I'll ping for re-review.

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -16,29 +17,133 @@ use fuels::{
...
+        let rng = thread_rng();
+        let db_name: String = rng
+            .sample_iter(&Alphanumeric)
+            .take(10)
+            .map(char::from)
+            .collect();
+
+        // The server connection string serves as a way to connect directly to the Postgres server.
+        // Example database URL: postgres://postgres:my-secret@localhost:5432
+        let connection_config: DatabaseConfig = std::env::var("DATABASE_URL")?.parse()?;
```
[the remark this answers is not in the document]
deekerno: Yep, will do!

rvmelkonian (approved):
Yep this is looking awesome, I like the new `pub enum TestError`. Just left one general thought but not a requested change.

PR #661, opened 2023-03-17
chore: update examples `hello-index` -> `hello-indexer`

rvmelkonian:
## changelog
• updates our examples from `index` to `indexer` for `hello-index`, `hello-index-native`, and `block-explorer`
• updates docs and commands
• updates the wasm scripts to work with the new file names.
• removes outdated commands from doc comments in `lib.rs` example files. 

### 1. build new wasm binaries
- Run the updated script to build the wasm test binaries (note the updated binary file names)
`bash scripts/utils/build_test_wasm_module.bash`

### 2. Hello World
- Start a local test Fuel node `cargo run --bin fuel-node`
- Start fuel-indexer with the example `cargo run --bin fuel-indexer -- run --manifest examples/hello-world/hello-indexer/hello_indexer.manifest.yaml`

### 3. Hello World Native
 - build the binary (the script doesn't build this one) `cargo build -p hello-indexer-native --release` (if this hangs run `cargo clean` , I had to do this).
 - Run it `cargo run --bin fuel-indexer -- run --manifest examples/hello-world-native/hello_indexer_native.manifest.yaml`

### 4. Block Explorer
- Run the block explorer example (the wasm binary is already built) `cargo run --bin fuel-indexer -- run --manifest examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml`

## Notes 

1. In `hello-indexer/src/lib.rs` the last doc comment says 
```
//! Now trigger an event.
//!
//! ```bash
//! cargo run --bin hello-bin
//! ```
```
this yields `error: no bin target named `hello-bin`.` 
I have removed this command as it must be out date.


ra0x3:
@rvmelkonian You will need to update all references in the book as well

rvmelkonian:
Yep. And also the native example.

ra0x3:
@rvmelkonian 
- Could you mention what the testing steps would be to test this PR? (in your PR description)
- I see both hello-world examples were changed so I imagine there are testing steps involved in making sure those both still work

examples/block-explorer/explorer-indexer/src/lib.rs
```diff
@@ -18,7 +18,7 @@
 //! assets from this example:
 //!
 //! ```bash
-//! cargo run --bin fuel-indexer -- --manifest examples/block-explorer/manifest.yaml
+//! cargo run --bin fuel-indexer -- --manifest examples/block-explorer/explorer_indexer.manifest.yaml
```
deekerno, line 21: This should match the manifest path in the `#[indexer(..)]` macro.

docs/README.md
```diff
@@ -194,13 +194,13 @@ You should see output indicating the successful creation of a database and start
 Now that we have our development environment set up, the next step is to create an index.
 
 ```bash
-forc index new hello-index --namespace my_project && cd hello-index
+forc index new hello-indexer --namespace my_project && cd hello-index
```
deekerno, line 197: `forc index new` will create a new directory with whatever string you give, so we should be `cd`-ing into `hello-indexer`.

deekerno (changes requested):
Left a couple comments on doc strings.

Echoing @ra0x3, we want to make sure that we're following the new instructions line-by-line any time we're adjusting any documentation (book, docstrings, etc.).

rvmelkonian:
I've added in testing steps and some notes. And I updated the necessary parts of the docs @deekerno, grep doesn't always work here as there are so many cases where we want to keep `index` and not use `indexer`. I've tried to comb through carefully.

examples/block-explorer/explorer-indexer/src/lib.rs
```diff
@@ -18,7 +18,7 @@
 //! assets from this example:
 //!
 //! ```bash
-//! cargo run --bin fuel-indexer -- --manifest examples/block-explorer/manifest.yaml
+//! cargo run --bin fuel-indexer -- --manifest examples/explorer-indexer/explorer_indexer.manifest.yaml
```
deekerno, line 21: Our `fuel-indexer` command requires a subcommand; in this case, that would be `run`. That's probably why you're encountering that error with the `--manifest` flag.

`cargo run --bin fuel-indexer -- run --manifest examples/explorer-indexer/explorer_indexer.manifest.yaml`

deekerno (changes requested):
Changes look good; left one comment related to your issue with the `--manifest` flag.

examples/block-explorer/explorer-indexer/src/lib.rs
rvmelkonian: noted down, thank you

deekerno (approved):
🚢

ra0x3 (approved):

PR #666, opened 2023-03-19
chore: write tests for plugins

### LINKED ISSUE #663: Write tests for our examples
As our examples will be developing, we should make sure the examples we currently have wont break. We should have a few basic tests for our examples. This will also be useful for testing locally and will speed our iteration time when developing examples. 

• The tests should make sure that that binaries build, the indexer can be ran with these binaries, events are logged. Make sure to cover any actions that are written in the examples section of the docs
• Add a job so that these tests run on CI

rvmelkonian:
closes #663

WIP


ra0x3:
@rvmelkonian I see CI is not ✅ ? I can review once CI passes

rvmelkonian:
@ra0x3 I think CI is failing because forc-indexer-plugin 0.5.0 has not yet been published to crates.io yet. But we can wait till after that's up.

rvmelkonian:
@ra0x3 CI is passing now.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ members = [
    "packages/fuel-indexer-types",
    "plugins/forc-index",
    "plugins/forc-postgres",
+   "plugins/plugin-tests",
```
ra0x3, line 28: Could we maybe name this to `forc-index-tests` for consistency?
- `fuel-indexer` --> `fuel-indexer-tests`
- `forc-index` --> `forc-index-tests`

plugins/plugin-tests/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+homepage = { workspace = true }
+license = { workspace = true }
+repository = { workspace = true }
+rust-version = { workspace = true }
+description = "Tests for the Forc Index plugins"
+
+[dependencies]
+
+forc-index = { workspace = true }
+fuel-indexer-plugin = { version = "0.5" }
```
ra0x3, line 15: I'm guessing we can't use the `workspace` version of `fuel-indexer-plugin` here?

plugins/plugin-tests/tests/commands.rs
```diff
@@ -0,0 +1,140 @@
+use forc_index::commands::{
+    build::Command as BuildCommand, init::Command as InitCommand,
+    new::Command as NewCommand,
+};
+use std::path::{Path, PathBuf};
+use tempfile::{Builder, TempDir};
+
+#[test]
+fn init_command_creates_correct_project_tree() {
```
ra0x3, line 9: It's convention that tests start with `test_` (checkout the tests in `fuel-indexer-tests`)

plugins/plugin-tests/tests/commands.rs
```diff
@@ -0,0 +1,140 @@
...
+            full_path.exists(),
+            "Expected path '{}' does not exist.",
+            full_path.to_string_lossy()
+        )
+    });
+    temp_dir.close().expect("Failed to close temp dir");
+}
+
+#[test]
+fn build_command_creates_artifact_at_expected_path() {
```
ra0x3, line 46: Same here with `test_` prefix

plugins/plugin-tests/tests/commands.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    assert!(
+        Path::new(&wasm_artifact_path).exists(),
+        "WASM artifact not found at expected path"
+    );
+    temp_dir.close().expect("Failed to close temp dir");
+}
+
+#[test]
+fn new_command_initializes_project_at_new_directory() {
```
ra0x3, line 87: Same here with `test_` prefix

.github/workflows/ci.yml
```diff
@@ -221,8 +221,15 @@ jobs:
         with:
           profile: minimal
           toolchain: stable
-      - name: cargo test --locked --all-targets --workspace --exclude fuel-indexer-tests
-        run: cargo test --locked --all-targets --workspace --exclude fuel-indexer-tests
+          target: wasm32-unknown-unknown
+
+      - uses: Swatinem/rust-cache@v1
```
ra0x3, line 226: - Can we break your CI additions out into their own `job`? (idk something like `forc-index-test`)
- This would have the same `if` as the current `check-forc-index` job

```yaml
if: needs.check-is-semver-branch.outputs.is_semver_branch != 'true'
```

So that we don't run it when cutting a new release or creating a new release PR

ra0x3 (changes requested):
@rvmelkonian Looks good mostly, thanks for doing this. Will be nice to have some certainty that the plugin works decently.

Left a few small pieces of feedback

plugins/plugin-tests/Cargo.toml
rvmelkonian: hmm .. we can!

ra0x3 (approved):
:shipit:

PR #672, opened 2023-03-20
fix: remove skipped dependent jobs for master/release

ra0x3:
- In Github actions a job is skipped if any one of it's dependencies is skipped, even if the `if:` statement of the job evaluates true
- So we need to unskip the docker/publish jobs by removing their skipped dependencies


deekerno (approved):
LGTM!

PR #675, opened 2023-03-20
ci: remove Postgres service from cargo-verifications build step

deekerno:
## Changelog
- Remove Postgres from the `services` for `cargo-verifications`
- Re-add `fuel-indexer` to package strategy matrix for `cargo-verifications`
- Add additional requirements to `validation-complete` step

## Notes
@Voxelot mentioned that we're pulling Postgres in on the build step for each package. However, we're not leveraging the database at all during the build step as we've disabled SQLx's compile-time query checking. So in order to reduce the frequency of rate limit errors, I'm removing Postgres from the build step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,7 +136,7 @@ jobs:
       matrix:
         command: [build]
         args: [--locked --all-features --all-targets]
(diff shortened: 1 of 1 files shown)


ra0x3 (approved):

ra0x3:
Approved ✅ but conflict ✖️

PR #682, opened 2023-03-21
fix: update deployment command

ra0x3:


tirkesi (approved):
looks good, hopefully this works

PR #676, opened 2023-03-20
chore: cleanup plugin command opts

ra0x3:
### Description
- Changes the `--output-dir-root` flag in `forc index build` and `forc index deploy` to a `--target-dir` flag in order to be more aligned [with the `cargo` `--target-dir` flag](https://doc.rust-lang.org/cargo/reference/config.html#configuration-keys)
- Also does some cleanup
- Renames some things to "indexer" for consistency slash so that they will work
- Fixes the `--run-migrations` flag in fuel-indexer so you don't have to `--run-migrations <VALUE>`, you just `--run-migrations`
- Updates docs

### Testing steps
- [ ] CI should pass
- [ ] Ensure flag still works
  - `cargo build -p forc-index --release`
  - `cargo run --bin fuel-indexer -- run --fuel-node-host node-beta-2.fuel.network --fuel-node-port 80`
  - `./target/release/forc-index deploy --path examples/block-explorer/explorer-indexer --target-dir /path/to/repos/fuel-indexer --release`

### Changelog
breaking: Swap `--output-dir-root` flag in `forc index {build,deploy}` to `--target-dir`

PR #686, opened 2023-03-21
docs: update docs

ra0x3:
### Description
- As you can see, the plugin docs were a bit dated, this PR changes that

### Testing steps
- [ ] N/A

### Changelog
- docs: bring plugin docs up to date

### DIFF
--- a/plugins/forc-index/src/commands/build.rs
+++ b/plugins/forc-index/src/commands/build.rs
@@ -3,14 +3,14 @@ use anyhow::Result;
 use clap::Parser;
 use std::path::PathBuf;
 
-/// Build an index.
+/// Build an indexer.
 #[derive(Debug, Parser)]
 pub struct Command {
     /// Manifest file name of indexer being built.
     #[clap(short, long, help = "Manifest file name of indexer being built.")]
     pub manifest: Option<String>,
 
-    /// Path of index project.
+    /// Path of indexer project.
     #[clap(short, long, help = "Path to the indexer project.")]
     pub path: Option<PathBuf>,
 
@@ -42,10 +42,10 @@ pub struct Command {
     #[clap(long, help = "Building for native execution.")]
     pub native: bool,
 
-    /// Path with which to prefix asset filepaths in the index manifest.
+    /// Path with which to prefix asset filepaths in the indexer manifest.
     #[clap(
         long,
-        help = "Path with which to prefix asset filepaths in the index manifest."
+        help = "Path with which to prefix asset filepaths in the indexer manifest."
     )]
     pub output_dir_root: Option<PathBuf>,
 }
--- a/plugins/forc-index/src/commands/deploy.rs
+++ b/plugins/forc-index/src/commands/deploy.rs
@@ -3,7 +3,7 @@ use anyhow::Result;
 use clap::Parser;
 use std::path::PathBuf;
 
-/// Deploy an index asset bundle to a remote or locally running indexer server.
+/// Deploy an indexer to an indexer service.
 #[derive(Debug, Parser)]
 pub struct Command {
     /// URL at which to deploy indexer assets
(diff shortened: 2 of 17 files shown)


plugins/forc-index/src/commands/revert.rs
```diff
@@ -4,22 +4,22 @@ use clap::Parser;
...
+    #[clap(short, long, help = "Path of indexer project.")]
     pub path: Option<PathBuf>,
 
-    /// Path to the manifest of indexer project being deployed.
+    /// Path to the manifest of the indexer project being reverted.
     #[clap(
         short,
         long,
-        help = "Path to the manifest of indexer project being deployed."
+        help = "Path to the manifest of indexer project being reverted."
```
deekerno, line 22: Should this match the docstring above?

plugins/forc-index/src/commands/remove.rs
```diff
@@ -3,11 +3,11 @@ use anyhow::Result;
...
+/// Stop and remove a running indexer.
 #[derive(Debug, Parser)]
 pub struct Command {
-    /// URL at which index is deployed.
-    #[clap(long, default_value = defaults::INDEXER_SERVICE_HOST, help = "URL at which index is deployed.")]
+    /// URL at which indexer is deployed.
+    #[clap(long, default_value = defaults::INDEXER_SERVICE_HOST, help = "URL at which indexer is deployed.")]
     pub url: String,
 
     /// Path to the manifest of indexer project being deployed.
```
deekerno, line 13: Should this also be changed to

> Path to the manifest of _the_ indexer project being deployed

to mirror the change in the revert operation?

deekerno (changes requested):
Looks good overall! Just some comments about consistency.

plugins/forc-index/src/commands/build.rs
```diff
@@ -42,10 +42,10 @@ pub struct Command {
     #[clap(long, help = "Building for native execution.")]
     pub native: bool,
 
-    /// Path with which to prefix asset filepaths in the index manifest.
+    /// Path with which to prefix asset filepaths in the indexer manifest.
     #[clap(
```
rvmelkonian, line 46: this comment / help has always confused me. Suggested change (or something similar) 
`/// output path of indexer assets in the manifest`

plugins/forc-index/src/commands/deploy.rs
```diff
@@ -55,10 +54,10 @@ pub struct Command {
     #[clap(long, help = "Building for native execution.")]
     pub native: bool,
 
-    /// Path with which to prefix asset filepaths in the index manifest.
+    /// Path with which to prefix asset filepaths in the indexer manifest.
```
rvmelkonian, line 57: If above comment is used, update this one as well.

plugins/forc-index/src/commands/init.rs
```diff
@@ -6,29 +6,29 @@ use std::path::PathBuf;
...
         parse(from_os_str),
-        help = "Path at which to create index."
+        help = "Path at which to create indexer."
     )]
     pub path: Option<PathBuf>,
 
-    /// Namespace in which index belongs.
-    #[clap(long, help = "Namespace in which index belongs.")]
+    /// Namespace in which indexer belongs.
+    #[clap(long, help = "Namespace in which indexer belongs.")]
```
rvmelkonian, line 23: In the file below `new.rs` we have `/// Name of the indexer namespace`. I'm neutral on which one we choose, but they should be the same for consistency.

rvmelkonian (changes requested):
Thanks for updating this. The old manifest args in the docs caused some confusion for me also. Just left a few comments.

plugins/forc-index/src/commands/build.rs
ra0x3: @rvmelkonian I agree this is confusing. It's actually cleaned up in https://github.com/FuelLabs/fuel-indexer/pull/676 (so don't mind it here)

rvmelkonian (approved):

PR #687, opened 2023-03-22
enhancement: make verbose logging configurable

ra0x3:
### Description
- PR adds `-v --verbose` flag to `fuel-indexer` and `fuel-indexer-api-server`
- Updates docs accordingly

### Testing steps
- Ensure `forc-index start` points to a latest `fuel-indexer` binary
- Run with verbose logging
  - [ ] Try running the block explorer with and without the `-v --verbose` flag

 
```bash
cargo run --bin fuel-indexer -- run \
  --fuel-node-host beta-3.fuel.network \
  --fuel-node-port 80 \
  --manifest examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml \
  --verbose
```
   - [ ] And the plugin as well (with and without verbosity)

```
cargo run --bin forc-index -- start -v
```

### Changelog
- enhancement: make verbose logging configurable

### DIFF
--- a/packages/fuel-indexer-lib/src/utils.rs
+++ b/packages/fuel-indexer-lib/src/utils.rs
@@ -328,11 +328,3 @@ pub fn host_triple() -> String {
         .expect("Failed to determine host triple via rustc.")[6..]
         .to_owned()
 }
-
-pub fn bool_to_str(b: bool) -> String {
-    if b {
-        "true".to_string()
-    } else {
-        "false".to_string()
-    }
-}
--- a/packages/fuel-indexer-lib/src/utils.rs
+++ b/packages/fuel-indexer-lib/src/utils.rs
@@ -1,18 +1,23 @@
-use crate::defaults;
+use crate::{config::IndexerConfig, defaults};
 use anyhow::Result;
 use fuel_indexer_types::Bytes32;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
-use std::str::FromStr;
 use std::{
+    env,
     fs::canonicalize,
     future::Future,
     net::{SocketAddr, ToSocketAddrs},
     path::Path,
     process::Command,
+    str::FromStr,
 };
 use tokio::time::{sleep, Duration};
 use tracing::{info, warn};
+use tracing_subscriber::filter::EnvFilter;
+
+const RUST_LOG: &str = "RUST_LOG";
+const HUMAN_LOGGING: &str = "HUMAN_LOGGING";
 
 // Testing assets use relative paths, while production assets will use absolute paths
 //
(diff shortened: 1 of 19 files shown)


packages/fuel-indexer-lib/src/utils.rs
```diff
@@ -325,3 +328,11 @@ pub fn host_triple() -> String {
...
         .to_owned()
 }
+
+pub fn bool_to_str(b: bool) -> String {
+    if b {
+        "true".to_string()
+    } else {
+        "false".to_string()
+    }
+}
```
deekerno, line 338: `bool` implements the `Display` trait, so you should just be able to call `boolean_value.to_string()` instead of having to create a function. Am I missing something here?

plugins/forc-index/src/ops/forc_index_start.rs
```diff
@@ -59,7 +59,9 @@ pub async fn init(command: StartCommand) -> anyhow::Result<()> {
         forc_postgres::commands::create::exec(Box::new(create_db_cmd)).await?;
     }
 
-    let mut cmd = Command::new("fuel-indexer");
+    let mut cmd = Command::new(
+        "/Users/rashad/development/repos/fuel-indexer/target/release/fuel-indexer",
+    );
```
deekerno, line 64: This needs to be changed before merging.

plugins/forc-index/src/ops/forc_index_start.rs
```diff
@@ -132,7 +137,7 @@ pub async fn init(command: StartCommand) -> anyhow::Result<()> {
             child.id()
         );
     } else {
-        anyhow::bail!("❌ Failed to spawn fuel-indexer child process.");
+        panic!("❌ Failed to spawn fuel-indexer child process.");
```
deekerno, line 140: Was this supposed to be changed from `anyhow::bail!` to `panic!`? If so, why?

deekerno (changes requested):
Testing steps work. So much less spam now, thank you so much! 😭 

In addition to the review comments, it looks like `sqlx` is ignoring the `verbose` parameter of the database connection string.
```ignoring unrecognized connect parameter: verbose=true```
```ignoring unrecognized connect parameter: verbose=false```

It doesn't affect the functionality at all, but I just thought you should be aware.

packages/fuel-indexer-lib/src/utils.rs
ra0x3: 🤦🏽 brain fart

plugins/forc-index/src/ops/forc_index_start.rs
ra0x3: @deekerno 
- Yea. If you run `forc index start` when `fuel-indexer` is not in the path it fails, but the error gets swallowed  by `anyhow::bail` (the process will die and you'll get nothing in stdout)
- I think we can`panic` here (probably elsewhere) where we don't need to do anything after the error (though I definitely should include the error in the message)

ra0x3:
> Testing steps work. So much less spam now, thank you so much! 😭
> 
> In addition to the review comments, it looks like `sqlx` is ignoring the `verbose` parameter of the database connection string. `ignoring unrecognized connect parameter: verbose=true` `ignoring unrecognized connect parameter: verbose=false`
> 
> It doesn't affect the functionality at all, but I just thought you should be aware.

@deekerno 
- Yea putting the param in the database URI is the only way I think to sneak `verbose_logging` into `IndexerConnection` w/o requiring a decent amount of api changes
- The warnings is obviously legit (since `verbose_logging` isn't an actual PostgreSQL URI param)
  - I didn't see anything [in the native list of URI params](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS) that would be a good substitute
- Definitely down to change it if there's a better method

deekerno:
@ra0x3: If anything, you could probably use [`query_pairs_mut`](https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut) after getting the `verbose` value by creating an iterator from the original query pairs and then setting them all back through a map while filtering out `verbose`...but that's a bit of work for something that doesn't affect the functionality anyways. 😅

But everything looks good, I'll approve when tests pass.

ra0x3:
> @ra0x3: If anything, you could probably use [`query_pairs_mut`](https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut) after getting the `verbose` value by creating an iterator from the original query pairs, clearing them on the original `url` object, setting them all back by iterating through the query iterator while filtering out `verbose` _and then_ use `url.to_string()`...but that's a bit of work for something that doesn't affect the functionality anyways. 😅
> 
> But everything looks good, I'll approve when tests pass.

@deekerno Ahh ok I see what you're saying - just grab the param w/o passing it. That's definitely possible. Change incoming

deekerno (approved):
:shipit:

PR #689, opened 2023-03-22
enhancement: support revert receipts

### LINKED ISSUE #669: Add support for `Revert` receipt
We don't yet support indexing for [`Revert` receipts](https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/abi/receipts.md#revert-receipt); we should add it to the requisite places in our codegen.

rvmelkonian:
Closes #669 

## Changelog
• updates codegen macro, schema, handler, web-api and creates new test for indexing Revert reciepts. 
• adds `pub enum FuelError` to `packages/fuel-indexer-types/src/abi.rs`

## Note
- When trying to store the `ra` value in the db, we get a postgres out of bounds error. The ra value emitted in the receipt `18446744073709486084`, is too large to store in the db. So I convert that to the equivalent value from the VM `PanicReason::Revert` found [here](https://github.com/FuelLabs/fuel-vm/blob/master/fuel-asm/src/panic_reason.rs). These seemed like a good solution, as the user could still see what error specifically was indexed, in this case a revert. 

- If we don't want developers to manually implement this error conversion in their indexer, we could create a module that they could use. I'll put up an issue.

## Testing 
Run the new test `cargo test test_can_trigger_and_index_panic_function_postgres --features e2e,postgres`


deekerno (changes requested):
@rvmelkonian You can represent the `u64` type with a `UInt8`. 👌🏽 You can also check [this section](https://fuellabs.github.io/fuel-indexer/master/reference-guide/data-types/types.html) of the indexer book for equivalent data types between Sway and GraphQL. Furthermore, you can check the receipt spec for descriptions for the other receipt types and see how entities represent receipt data fields of type `u64`.

Other than the missing `val` field, it looks good! Great job!

rvmelkonian:
@deekerno Nice. That table in our docs is very useful.

packages/fuel-indexer-types/src/abi.rs
```diff
@@ -155,3 +189,17 @@ impl NativeFuelType for Call {
         type_id(FUEL_TYPES_NAMESPACE, "Call") as usize
     }
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct Revert {
+    pub contract_id: ContractId,
+    pub ra: u64,
```
deekerno, line 196: What if we changed this to something like "error_val" in order to give users a more descriptive name? It's technically `ra` for "register A" but it would be nice to communicate the intended usage.

packages/fuel-indexer-types/Cargo.toml
```diff
@@ -11,6 +11,7 @@ description = "Fuel Indexer Types"
 
 [dependencies]
 chrono = { version = "0.4", features = ["serde"] }
+fuel-abi-types = "0.2.1"
```
deekerno, line 14: Since you're not using this crate anymore, you can remove this.

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -509,6 +509,12 @@ fn process_fn_items(
                                 let data = bincode::serialize(&abi::TransferOut{ contract_id: id, to, asset_id, amount, pc, is }).expect("Bad encoding,");
                                 decoder.decode_type(ty_id, data);
                             }
+                            Receipt::Revert { id, ra, .. } => {
+                                #contract
+                                let ty_id = abi::Revert::type_id();
+                                let data = bincode::serialize(&abi::Revert{ contract_id: id, ra }).expect("Bad encoding,");
+                                decoder.decode_type(ty_id, data);
+                            }
```
deekerno, line 517: Just a thought, but what if we moved the bitmasking from the handler to the codegen itself? For example (using the other suggestion of an `error_val` field in `abi::Revert`):
```
Receipt::Revert { id, ra, .. } => {
    #contract
    let ty_id = abi::Revert::type_id();
    let data = bincode::serialize(&abi::Revert{ contract_id: id, error_val: u64::from(ra & 0xF) }).expect("Bad encoding,");
    decoder.decode_type(ty_id, data);
}
```

It seems improbable that a user would use the very large decimal values directly from the receipt other than to apply a bitmask upon them so if we we do it for them here in the codegen, they wouldn't have to worry about it. We could then add a mapping of the much smaller values and their meanings to a `Revert` page of the Receipts section of the book. Additionally, they could also store the value straight into the database as well.

Furthermore, you wouldn't have to define a `FuelError` enum or implement the `unmask()` operation.

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -644,6 +644,28 @@ pub mod test_web {
...
+
+        println!("Result: {:?}", result);
+
+        match result {
+            Ok(_) => HttpResponse::Ok(),
+            Err(_) => {
+                //Force 200 on err response here, because we have successfully triggered a panic
+                HttpResponse::Ok()
+            }
+        }
```
deekerno, line 666: Nice catch here! I think you can get away with removing the comment and result printing, and then matching on just the exhaustive case since we're doing the same action either way. For example:
```
    enum MysteryIncSounds {
        Zoinks,
        Jinkies,
        RuhRoh
    }
    
    let mystery_inc_sound = MysteryIncSounds::Zoinks;
    
    match mystery_inc_sound {
        _ => println!("Why are kids investigating crimes anyways?"),
    }
```

packages/fuel-indexer-types/src/abi.rs
```diff
@@ -6,6 +6,40 @@ pub use fuel_tx::Receipt;
...
+        match self {
+            FuelError::FailedRequire => 0x0000,
+            FuelError::FailedTransferToAddress => 0x0001,
+            FuelError::FailedSendMessage => 0x00002,
+            FuelError::FailedAssertEq => 0x0003,
+            FuelError::FailedAssert => 0x0004,
+        }
+    }
+}
+
```
deekerno, line 42: See other comment about doing bitmasking in the codegen phase.

deekerno (changes requested):
Looking good! Just a few thoughts.

packages/fuel-indexer-tests/src/fixtures.rs
rvmelkonian: This is a vivd example, love it.

packages/fuel-indexer-macros/src/indexer.rs
rvmelkonian: Yes, you are right we shouldn't expose this to the user 👌.
I'll update the receipt section of the docs in this PR also.

rvmelkonian:
@deekerno I've addressed you're feedback, yeah it's much better just doing `error_val: u64::from(ra & 0xF)` in the codegen! Also, be sure to run `mdbook serve` as I have updated the book. I also re-ordered the items of `7.2 Receipts` so that they are in alphabetical order.

packages/fuel-indexer-tests/src/fixtures.rs
rvmelkonian: turns out this is fine 
```rust
async fn fuel_indexer_test_trigger_revert(
        state: web::Data<Arc<AppState>>,
    ) -> impl Responder {
        let _ = state
            .contract
            .methods()
            .trigger_revert()
            .tx_params(tx_params())
            .call()
            .await;

        HttpResponse::Ok()
    }
    ```
    I initially added the match as I was getting error response codes, but it must have been for some other reason.

File: docs/src/reference-guide/indexing/receipts/revert.md [hunk 6]
```diff
@@ -0,0 +1,31 @@
...
+| FailedAssertEq        | 3     |
+| FailedAssert          | 4     |
+
+- [Read more about `Revert` in the Fuel Protocol spec](https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/abi/receipts.md#log-receipt)
+
+You can handle functions that could produce a `Revert` receipt by adding a parameter with the type `abi::Revert`.
+
+```rust, ignore
+fn handle_revert(revert: abi::Revert) {
+  // handle the transfer out
```
deekerno, line 29: This should be `// handle the revert`.

deekerno (changes requested):
Left some final thoughts. Once they're addressed, I'll approve. ✅

packages/fuel-indexer-types/src/abi.rs
```diff
@@ -155,3 +155,17 @@ impl NativeFuelType for Call {
...
     }
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct Revert {
+    pub contract_id: ContractId,
+    pub error_val: u64,
+}
+
+pub struct SomeThing();
```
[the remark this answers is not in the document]
rvmelkonian: My apologies for leaving it! I was experimenting and left it in.

docs/src/reference-guide/indexing/receipts/revert.md
rvmelkonian: Whoops, thanks.

deekerno (approved):
Thanks for adding some documentation! Nice job on this!

🚢

PR #705, opened 2023-03-24
fix: do not panic when there is an empty data field in `MessageOut` receipt

deekerno:
## Changelog
- Safely decode type ID and data from `data` field of `MessageOut` receipt

## Testing Plan
We can't create a test for this just yet as I don't know how someone was able to generate an empty data field. It was suggested that one can use an empty `Bytes` array to trigger it, but I was unable to get it working. In any case, this should solve our issues in which the indexer was crashing on the beta-3 testnet.

### Manual Steps
0. Build WASM modules and ensure DB is cleared.
1. Run examples on beta-2; indexers should work as expected.
2. Run examples on beta-3; indexers should work as expected.

## Notes
This error was caused by a `MessageOut` receipt with an empty hex string for the `data` field:
```
{
  "receiptType": "MESSAGE_OUT",
  "recipient": "0x000000000000000000000000172eda4e99d41a97d7c1849a8d05f46de9eeb8dc",
  "sender": "0xb53d4780ab195abdbd164a4cd5692a4e4ebc61b8d048abb42841ccb6cee150cf",
  "amount": "100000000",
  "data": "0x", <---
  "nonce": "0x632aaa4129d018ac820982dba793c5ec22c12bd19bef57521cf42ba7b7b0aedf"
}
```

We were decoding the type by copying from a slice of the data field, which panics if the slice lengths are not the same; ultimately, this is what was causing the WASM `Trap` error codes.

### DIFF
--- a/packages/fuel-indexer-macros/src/indexer.rs
+++ b/packages/fuel-indexer-macros/src/indexer.rs
@@ -105,7 +105,6 @@ fn process_fn_items(
                 }
             }
         })
-        // Add an arm for a unit struct in case of empty data fields
         .chain(
             vec![quote! {
                 u64::MAX => {
@@ -475,15 +474,20 @@ fn process_fn_items(
                                 // It's possible that the data field was generated from an empty Sway `Bytes` array
                                 // in the send_message() instruction in which case the data field in the receipt will
                                 // have no type information or data to decode, so we decode an empty vector to a unit struct
-                                let type_id = match data.get(..8) {
-                                    Some(buffer) => u64::from_be_bytes(<[u8; 8]>::try_from(&buffer[..]).expect("Could not get type ID for data in MessageOut receipt")),
-                                    None => u64::MAX,
-                                };
+                                let type_id = data
+                                    .get(..8)
+                                    .map(|buffer| {
+                                        u64::from_be_bytes(
+                                            <[u8; 8]>::try_from(&buffer[..])
+                                                .expect("Could not get type ID for data in MessageOut receipt"),
+                                        )
+                                    })
+                                    .unwrap_or(u64::MAX);
 
-                                let data = match data.get(8..) {
-                                    Some(buffer) => buffer.to_vec(),
-                                    None => Vec::<u8>::new(),
-                                };
+                                let data = data
+                                    .get(8..)
+                                    .map(|buffer| buffer.to_vec())
+                                    .unwrap_or(Vec::<u8>::new());
 
                                 let receipt = abi::MessageOut{ message_id, sender, recipient, amount, nonce, len, digest, data };
                                 decoder.decode_messageout(type_id, receipt);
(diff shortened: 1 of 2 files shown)


packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -463,10 +472,20 @@ fn process_fn_items(
...
                             Receipt::MessageOut { message_id, sender, recipient, amount, nonce, len, digest, data } => {
-                                let mut buf = [0u8; 8];
-                                buf.copy_from_slice(&data[0..8]);
-                                let type_id = u64::from_be_bytes(buf);
-                                let receipt = abi::MessageOut{ message_id, sender, recipient, amount, nonce, len, digest, data: data[8..].to_vec() };
+                                // It's possible that the data field was generated from an empty Sway `Bytes` array
+                                // in the send_message() instruction in which case the data field in the receipt will
+                                // have no type information or data to decode, so we decode an empty vector to a unit struct
+                                let type_id = match data.get(..8) {
+                                    Some(buffer) => u64::from_be_bytes(<[u8; 8]>::try_from(&buffer[..]).expect("Could not get type ID for data in MessageOut receipt")),
```
ra0x3, line 479: Maybe?

```rust
let type_id = data.get(..8).map(|x| u64::from_be_bytes(<[u8; 8]>::try_from(&x[..]).expect("Whoops!"))).unwrap_or(u64::MAX);
```

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -463,10 +472,20 @@ fn process_fn_items(
...
+                                    None => u64::MAX,
+                                };
+
+                                let data = match data.get(8..) {
```
ra0x3, line 483: Maybe?
```rust
let data = data.get(8..).map(|x| x.to_vec()).unwrap_or(Vec::<u8>::new());
```

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -463,10 +472,20 @@ fn process_fn_items(
                                 }
                             }
                             Receipt::MessageOut { message_id, sender, recipient, amount, nonce, len, digest, data } => {
-                                let mut buf = [0u8; 8];
-                                buf.copy_from_slice(&data[0..8]);
-                                let type_id = u64::from_be_bytes(buf);
-                                let receipt = abi::MessageOut{ message_id, sender, recipient, amount, nonce, len, digest, data: data[8..].to_vec() };
+                                // It's possible that the data field was generated from an empty Sway `Bytes` array
```
ra0x3, line 475: 👌🏽

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -105,6 +105,15 @@ fn process_fn_items(
                 }
             }
         })
+        // Add an arm for a unit struct in case of empty data fields
```
ra0x3, line 108: @deekerno I think we can remove this comment (I think the code makes it clear what's goin on)

ra0x3 (approved):
Thanks @deekerno  🥇

PR #718, opened 2023-03-29
chore: upgrade `wasmer` to v2.3.0

deekerno:
## Changelog
- Upgrade `wasmer` from `v2.0` to `v2.3`

## Testing Plan
Tests should pass. Each example should pass as well.

## Notes
_No changes in functionality_; this is just an upgrade to bring in some fixes for memory leaks and the like that have been added after `wasmer`'s `v2.0` release. We should still investigate refactoring to use `v3.0` or above, just at a later time.


rvmelkonian (approved):
Test suite passes locally and examples building. ✅

PR #719, opened 2023-03-31
enhancement: support many contracts one manifest

### LINKED ISSUE #697: Subscribe to multiple contracts in the manifest file
The manifest file contains a field `contract_id` which is used to subscribe to a contract.
You may specify an id or leave it blank to listen to everything on the network.

It would be nice to subscribe to multiple contracts at once.

The contract id file may take the form of a vector e.g.

```sh
contract_id: [ "0x00..00", "0x00..01" ]
```

rvmelkonian:
Closes #697

@Braqzen requested this feature, where many contractIds can be passed into a single manifest for use by a single indexer. This PR adds that functionality.

## Changelog
• Updates the`Manifest.contract_id` field to be of enum type `ContractIds`
• Updates the `indexer.rs` to handle which variant of the enum is passed. 
• Implements `FromStr for ContractIds`
• Adds the functions `fn contract_id_from_str` and `contract_id_to_str`. See note Below

## Notes
1. In order to maintain ergonomic input from our manifest files, I had to implement custom serialization and deserialization on the `contract_id` field using the `serde_json` crate. This is so that a user can do any of these in their manifest, and it will all be valid. 

```
contract_id : ~
contract_id: fuel1234
contract_id: 
   - fuel1234
   - fuel5678
contract_id: [fuel1234, fuel5678]
```
The multi-line array format is the standard for yaml, but a single-line array is also valid, both will work here. 

2. Currently if the contracts have different ABIs and therefor different types, a single indexer will not able to index both of these different contract ABIs. This functionality is outside the scope of this PR.

## Testing 
Initially I had bolted on a lot of additional logic in order to write an integration test for this, but it seemed to add a lot bloat for one case, so that was removed.

For manual testing, change one of the test `manifest.yaml` files. For each of the following yaml field formats, build the wasm binary. If you want you can also run the ci-test-suite locally for each change; I have done this on my side.   
```
contract_id: [fuel123]
contract_id: 
   - fuel123
contract_id: fuel123
```


examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml
```diff
@@ -7,6 +7,6 @@ graphql_schema: examples/block-explorer/explorer-indexer/schema/explorer_indexer
 module:
   wasm: target/wasm32-unknown-unknown/release/explorer_indexer.wasm
 metrics: ~
-contract_id: ~
+contract_id: ~
```
deekerno, line 10: Seems that a space was added here. That's not required for the code to work properly, right?

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -255,18 +257,47 @@ fn process_fn_items(
...
+                    quote! {
+                        let manifest_contract_id = Bech32ContractId::from_str(#contract_id).expect("Failed to parse manifest 'contract_id' as Bech32ContractId");
+                        if receipt_contract_id == manifest_contract_id {
+                            found = true;
+                        }
+                    }
+                }
+                None => quote! {},
+            }
+        });
```
deekerno, line 287: It would be more efficient to instantiate a `HashSet` from the contract IDs and check whether it contains the receipt contract ID.

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -46,6 +47,31 @@ impl ToString for Module {
     }
 }
 
+/// in a `Manifest` struct. It can either be a single `Option<String>`
+/// or a `Vec<Option<String>>`.
+///
+/// - `Single`: Represents an optional single contract ID as an `Option<String>`.
+/// - `Multiple`: Represents a vector of optional contract IDs as a `Vec<Option<String>>`.
```
deekerno, line 54: Can you change this so that its first line is consistent with the other doc comments?

deekerno (changes requested):

rvmelkonian:
Putting this back into draft, as I just found a bug when testing again.

examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml
rvmelkonian: Nope, just a typo. Will fix.

packages/fuel-indexer-lib/src/manifest.rs
rvmelkonian: Yep!

rvmelkonian:
@deekerno I've integrated your feedback, let me know your thoughts. Also I added new testing steps, if you could try those out, would be grateful. 🙏

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -46,6 +47,30 @@ impl ToString for Module {
     }
 }
 
+/// Represents contract IDs in a `Manifest` struct.
+///
+/// - `Single`: Represents an optional single contract ID as an `Option<String>`.
```
ra0x3, line 52: Same comment as [here](https://github.com/FuelLabs/fuel-indexer/pull/755#discussion_r1165731050) with regard to Rust doc struct field comments

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -46,6 +47,30 @@ impl ToString for Module {
...
+    Single(Option<String>),
+    Multiple(Vec<Option<String>>),
+}
+
+impl FromStr for ContractIds {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.starts_with('[') {
+            serde_json::from_str::<Vec<Option<String>>>(s)
```
ra0x3, line 65: @rvmelkonian Curious as to why this is `Vec<Option<String>>` and not `Vec<String>`. Under what circumstances would any of these contract IDs be `None`?

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -46,6 +47,30 @@ impl ToString for Module {
...
+                .map(ContractIds::Multiple)
+                .map_err(|err| err.to_string())
```
ra0x3, line 67: We're not really doing anything in this `.map_err`. Do we need it?

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -61,19 +86,60 @@ pub struct Manifest {
...
+                .into_iter()
+                .filter_map(|val| match val {
+                    serde_yaml::Value::String(s) => Some(s),
+                    _ => None,
+                })
+                .collect::<Vec<_>>();
+            Ok(ContractIds::Multiple(ids.into_iter().map(Some).collect()))
+        }
+        serde_yaml::Value::Null => Ok(ContractIds::Single(None)),
+        _ => Err(serde::de::Error::custom("Invalid contract_id value")),
```
ra0x3, line 119: Looks like we might be able to integrate `serve::de::Error` into `ManifestError`?

ra0x3 (changes requested):
@rvmelkonian 
- Looks mostly OK
- Left a comment mentioning that we can assume good input from the user 
  - And if we don't get good input (contract IDs) from the user we can just panic

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -255,18 +257,42 @@ fn process_fn_items(
...
+            let contract_ids_str: Vec<String> = contract_ids
+                .iter()
+                .filter_map(|id| id.as_ref().cloned())
+                .collect();
+
             quote! {
-                let manifest_contract_id = Bech32ContractId::from_str(#contract_id).expect("Failed to parse manifest 'contract_id' as Bech32ContractId");
                 let receipt_contract_id = Bech32ContractId::from(id);
-                if receipt_contract_id != manifest_contract_id {
+                let contract_ids_set: HashSet<Bech32ContractId> = {
```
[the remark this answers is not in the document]
deekerno: There should be a way; you could do a map across the list of contract ID strings, turn it back into an iter, and pass it into `HashSet::from_iter()`. I played around with this to see the set could be constructed at compile time instead of at run time, but it looks like `ToTokens` isn't implemented for `HashSet<Bech32ContractId>`. 🤷🏽‍♂️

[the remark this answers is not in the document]
rvmelkonian: Yeah I also felt this was a little janky, I tried to write it initially in a more readable fashion, but the compiler led me into this direction for reasons I can't exactly recall. Let me have another crack at this block.

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -61,19 +86,60 @@ pub struct Manifest {
...
+        deserialize_with = "contract_id_from_str",
+        serialize_with = "contract_id_to_str"
+    )]
+    pub contract_id: ContractIds,
     pub start_block: Option<u64>,
     #[serde(default)]
     pub resumable: Option<bool>,
 }
 
+fn contract_id_from_str<'de, D>(deserializer: D) -> Result<ContractIds, D::Error>
```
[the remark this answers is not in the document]
rvmelkonian: Yep, agreed. Time to promote 🎖️

packages/fuel-indexer-lib/src/manifest.rs
rvmelkonian: Updated to: 
```
pub enum ContractIds {
    ///Single represents a single contract ID as an `Option<String>`.
    Single(Option<String>),
    ///Multiple represents a vector of contracts IDs as a Vec<String>.
    Multiple(Vec<String>),
}
```
Single still should be Option<String> as before, so that we can deal with ~ yaml values.

packages/fuel-indexer-lib/src/manifest.rs
rvmelkonian: If we remove `.map_err` the error type wont match the one required by the `FromStr` trait: `type Err = String;`

packages/fuel-indexer-macros/src/indexer.rs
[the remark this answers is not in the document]
rvmelkonian: Yeah I've been wrestling with this for a bit now. The only way is to implement `ToTokens` with a custom type wrapper like: 
```rust
pub struct Bech32ContractIdSet(HashSet<Bech32ContractId>);
```
The ToTokens custom method seems like a lot of code for a small bit of functionality, whereas what we have here currently is compiling with no issues. If we think, implementing this method will be useful down the line, I am happy to implement this.

packages/fuel-indexer-lib/src/manifest.rs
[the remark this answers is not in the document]
rvmelkonian: So in the end I implemented these methods on the `ContractIds` method as the `deserialize` method is returning a variant of `ContractIds`. Also no doc comments here as is a private method.

packages/fuel-indexer-lib/src/manifest.rs
rvmelkonian: I tried this for a while as I really wanted this. But because we are doing this funky this with serde in the `Manifest` struct :
```rust
#[serde(
        serialize_with = "ContractIds::serialize",
        deserialize_with = "ContractIds::deserialize"
    )]
    pub contract_id: ContractIds,
```
It always needs to return it's own very specific error and complains when I try to return `ManifestError`, I tried to get around this by tying those types together, but in short it got messy and wasn't working.

packages/fuel-indexer-tests/src/lib.rs
```diff
@@ -21,6 +21,11 @@ pub enum TestError {
     IndexerConfigError(#[from] IndexerConfigError),
 }
 
+pub enum ContractAmount {
```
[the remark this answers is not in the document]
rvmelkonian: Nope. It's gone now.

packages/fuel-indexer-lib/src/manifest.rs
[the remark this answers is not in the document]
rvmelkonian: I could have had the enum be of variant types `ContractId(String)`, but after adding the methods onto the enum, we don't seem to gain much from this at this point.

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -151,3 +156,67 @@ impl TryFrom<&Vec<u8>> for Manifest {
...
+pub enum ContractIds {
+    ///Single represents a single contract ID as an `Option<String>`.
+    #[serde(alias = "single")]
+    Single(Option<String>),
+    ///Multiple represents a vector of contracts IDs as a Vec<String>.
+    #[serde(alias = "multiple")]
+    Multiple(Vec<String>),
+}
+
+impl ContractIds {
```
ra0x3, line 172: 🏅

packages/fuel-indexer/src/service.rs
```diff
@@ -63,7 +63,10 @@ impl IndexerService {
         )
         .await?;
 
+        println!("Registered Index({})", &manifest.uid());
```
ra0x3, line 66: print statements are still here.  If these are informative let's change them to `info!`. If they were just used for testing, let's remove them?

packages/fuel-indexer-macros/src/indexer.rs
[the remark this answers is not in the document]
ra0x3: @rvmelkonian So I don't believe it requires you to implement `ToTokens`.

```rust
        ContractIds::Multiple(contract_ids) => {
            let contract_ids = contract_ids
                .iter()
                .map(|id| {
                    quote! {
                        Bech32ContractId::from_str(#id)
                            .expect("Failed to parse manifest 'contract_id' as Bech32ContractId")
                    }
                })
                .collect::<Vec<proc_macro2::TokenStream>>();

            quote! {
                let receipt_contract_id = Bech32ContractId::from(id);
                let contract_ids = HashSet::from([#(#contract_ids),*]);

                if !contract_ids.contains(&receipt_contract_id) {
                    Logger::info("Not subscribed to this contract. Will skip this receipt event. <('-'<)");
                    continue;
                }
            }
        }
```

Doesn't have to be this ☝🏽 but something like this is more so what I was referring to. 

Which produces:

```rust
                    Receipt::Call {
                        id: contract_id, amount, asset_id, gas, param1, to: id, .. }
                        => {
                        let receipt_contract_id = Bech32ContractId::from(id);
                        let contract_ids =
                            HashSet::from([Bech32ContractId::from_str("fuel1t8v0g6h2jcnjtlpa0p3z244qcqjc8tp2g6fuf63uc8jam6e4j4uqen2kpu").expect("Failed to parse manifest 'contract_id' as Bech32ContractId"),
                                        Bech32ContractId::from_str("fuel18hchrf7f4hnpkl84sqf8k0sk8gcauzeemzwgweea8dgr7eachv4s86r9t9").expect("Failed to parse manifest 'contract_id' as Bech32ContractId")]);
                        if !contract_ids.contains(&receipt_contract_id) {
                                Logger::info("Not subscribed to this contract. Will skip this receipt event. <('-'<)");
                                continue;
                            }
                        let fn_name = decoder.selector_to_fn_name(param1);
                        return_types.push(param1);
                        callees.insert(id);
                        let data =
                            bincode::serialize(&abi::Call {
                                            contract_id,
                                            to: id,
                                            amount,
                                            asset_id,
                                            gas,
                                            fn_name,
                                        }).expect("Bad encoding");
                        let ty_id = abi::Call::type_id();
                        decoder.decode_type(ty_id, data);
                    }
```

- Which does bring up the fact that we're creating this `HashSet` on each `Receipt` match which is obviously not ideal (but outside of the scope of this PR). 
   - However I do think we should follow up on this to prevent us having to allocate this `HashSet` every indexer call #EveryComputationCounts 
   - Doesn't necessarily have to be you that does it

ra0x3 (changes requested):
@rvmelkonian 
- I pulled this PR down and tested it
- Things compile fine and seem to work OK
- I left some feedback on cleaning up how we're creating the `Hashset` for multiple contracts
  - It might require a small follow up PR to optimize but that's fine

packages/fuel-indexer/src/service.rs
rvmelkonian: My bad, I removed them. But they turned up again after I copied in my older commit.

packages/fuel-indexer-macros/src/indexer.rs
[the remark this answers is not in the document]
rvmelkonian: Made a note of this. Will create an issue.

ra0x3:
@rvmelkonian Just tested this out. Looks good. Just will paste what I wrote to you on Slack.

> Small note on your PR, because it's updating the way manifest works, you'll need to update the contract_id section of docs/manifest.md to just add a blurb toward the end mentioning that the contract_id param can either be a single string or a list of strings

Once the docs are updated, then merge 👌🏽

ra0x3 (approved):
💪🏽

packages/fuel-indexer-lib/src/manifest.rs
ra0x3: @rvmelkonian So I don't see anything [in `FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) that means you _have_ to `.map_err` like this. But not a blocker for merge - can always be fixed in a follow up.

Thanks for the forbearance in all of this

deekerno (approved):

PR #726, opened 2023-04-03
chore: fix CI commands for plugin tests

deekerno:
## Changelog
- Ensure that `cargo-test-forc-index-plugins` CI step only tests the plugins
- Remove unnecessary exclusions
- Remove `check-forc-index` as it's superseded by `cargo-test-forc-index-plugins`

## Testing Plan
CI should pass.

## Notes
`cargo-test-forc-index-plugins` was testing the entire workspace _and_ running each test sequentially. This changes that step so that it only runs the plugin tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,38 +272,6 @@ jobs:
       - name: cargo test --locked --all-targets -p plugin-tests -- --test-threads=1
         run: cargo test --locked --all-targets -p plugin-tests -- --test-threads=1
 
-  check-forc-index:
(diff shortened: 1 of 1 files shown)


ra0x3:
@deekerno While we're in here, is there any way we can combine `check-forc-index` and `cargo-test-forc-index-plugins` ?

deekerno:
@ra0x3 It seems that we can remove `check-forc-index` as it fires under the same conditions and its functionality is already covered by `cargo-test-forc-index-plugins` as `test_build_command_creates_artifact_at_expected_path` does both `init` and `build`. What do you think?

ra0x3 (approved):

PR #728, opened 2023-04-03
chore: update quickstart

ra0x3:
### Description
- This PR just moves our Quickstart out of the README and over to docs/QUICKSTART.md
- The Quickstart is still referenced in the README

### Testing steps
- [ ] `cd docs && mdbook serve`
  - You should see the Quickstart link working as normal in the book

### Changelog 
- chore: update quickstart


deekerno (commented):
@ra0x3 I generally support this change.

What do you think about moving the link to the quickstart to the top of the README with a call to action, i.e. "Want to get started right away? Check out our Quickstart!"? That way, it's almost impossible to miss. If we do that, we should add a little section about installing WASM targets and `wasm-snip` to the beginning of the quickstart document.

ra0x3:
> @ra0x3 I generally support this change.
> 
> What do you think about moving the link to the quickstart to the top of the README with a call to action, i.e. "Want to get started right away? Check out our Quickstart!"? That way, it's almost impossible to miss. If we do that, we should add a little section about installing WASM targets and `wasm-snip` to the beginning of the quickstart document.

@deekerno Sounds fine w/ me. If you have an idea for how it might/should look feel free to co-author this PR so I can see what you mean exactly

deekerno (approved):
:ship:

PR #732, opened 2023-04-04
chore: default `true` for `--release` `forc index build` and `deploy`

### LINKED ISSUE #712: Make `forc index build` and `forc index deploy` use `--release` by default
- Title says it all
- We should default to passing a `--release` flag in those ☝🏽 `forc index` commands
- More info here https://github.com/FuelLabs/fuel-indexer/issues/684

Additional context
- This will also require updating the docs for the `deploy` and `build` commands

rvmelkonian:
Closes #712

## Changelog
• Updates Command structs to use default value for release `"true"`
• Updates docs and doc comments. 

## Notes
Clap's`default_value` field on accepts values of type `&str`, release is of type `bool`, however the string slice "true" will be parsed as a bool.

## Testing 
The simplest way I found to test this was to `println!` the release value when running the command in a dummy indexer project. I removed the log in this PR.


plugins/forc-index/src/commands/deploy.rs
```diff
@@ -30,11 +30,12 @@ pub struct Command {
...
     pub target: String,
 
-    /// Build optimized artifacts with the release profile.
+    /// Build optimized artifacts with the release profile. Set to true by default.
     #[clap(
         short,
         long,
-        help = "Build optimized artifacts with the release profile."
+        help = "Build optimized artifacts with the release profile.",
+        default_value = "true"
```
ra0x3, line 38: Could we keep the pattern of using default values?

```rust
// forc-index/src/utils/defaults.rs

// ...

pub const BUILD_RELEASE_PROFILE: bool = true; // or whatever you wanna call it
```
And then 

```rust
  #[clap(
         short,
         long,
         help = "Build optimized artifacts with the release profile.",
         default_value_t = defaults::BUILD_RELEASE_PROFILE,

```

plugins/forc-index/src/commands/build.rs
```diff
@@ -18,11 +18,12 @@ pub struct Command {
     #[clap(long, default_value = defaults::INDEXER_TARGET, help = "Target at which to compile.")]
     pub target: String,
 
-    /// Build optimized artifacts with the release profile.
+    /// Build optimized artifacts with the release profile. Set to true by default.
```
ra0x3, line 21: We don't have to mention that it's set to true by default. Clap will automagically show this when the user runs `--help`

ra0x3 (changes requested):
@rvmelkonian Looks good thanks, 2 small requested changes, then ✅

plugins/forc-index/src/commands/deploy.rs
rvmelkonian: Updated in the new commit

ra0x3 (changes requested):

docs/src/reference-guide/plugins/forc-index/build.md
```diff
@@ -17,7 +17,7 @@ OPTIONS:
         --native                     Building for native execution.
     -p, --path <PATH>                Path to the indexer project.
         --profile <PROFILE>          Build with the given profile.
-    -r, --release                    Build optimized artifacts with the release profile.
+    -r, --release                    Build optimized artifacts with the release profile. Set to true by default.
```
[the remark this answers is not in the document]
rvmelkonian: I created this issue to track your suggestion https://github.com/FuelLabs/fuel-indexer/issues/737

ra0x3 (approved):

PR #741, opened 2023-04-06
breaking: swap uint8 and id types to postgres numeric

ra0x3:
### Description
- Changes to some integer types to `numeric`
  - This gives us more flexibility by allowing us to manually define our own scale and precision for these values
  - Sqlx (and possibly others) use `BigDecimal` to interface with `numeric` PostgresQL types 
    - This `BigDecimal` can be converted into various other simple(r) types 
 - Cleans up the `manifest.rs` module
 - Cleans up and updates tests
 - Fixes native execution
   - We keep forgetting to update native execution when we change things :( 
- The previously broken/skipped tests hidden behind the feature flag now work 🥳 so that flag was removed

### Testing steps
⚠️ This is a non-trivial breaking change, so all testing steps should be checked

- [ ] CI should pass
- [ ] Hello world example should work
- [ ] Block explorer example should work
- [ ] Hello world native example should work

### Changelog
- breaking: swap uint8 and id types to postgres numeric


packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
```diff
@@ -358,178 +317,151 @@ async fn test_can_trigger_and_index_logdata_event_postgres() {
...
-#[cfg(all(feature = "e2e", feature = "postgres", feature = "pg-embed-skip"))]
-async fn test_can_trigger_and_index_messageout_event_postgres() {
-    let (fuel_node_handle, test_db, mut srvc) = setup_test_components().await;
-
-    let mut manifest: Manifest =
-        serde_yaml::from_str(assets::FUEL_INDEXER_TEST_MANIFEST).expect("Bad yaml file.");
+// #[actix_web::test]
+// #[cfg(all(feature = "e2e", feature = "postgres", feature = "pg-embed-skip"))]
+// async fn test_can_trigger_and_index_messageout_event_postgres() {
+//     let (node_handle, test_db, mut srvc) = setup_test_components().await;
```
deekerno, line 402: Is this test supposed to be commented out? None of the other `pg-embed-skip` tests are commented out, so I'm not sure.

deekerno (commented):
Testing steps work for me. Left one question.

packages/fuel-indexer-lib/src/manifest.rs
```diff
@@ -7,22 +7,10 @@ use std::{
...
-    pub contract_id: Option<String>,
-    pub start_block: Option<u64>,
-    #[serde(default)]
-    pub resumable: Option<bool>,
-}
-
+/// Result returned from Manifest operations.
 type ManifestResult<T> = Result<T, ManifestError>;
 
+/// Error type returned from Manifest operations.
```
rvmelkonian, line 13: Yay to doc comments. Will start doing more of this.

packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
ra0x3: @deekerno It's not. I'll address 👌🏽

rvmelkonian (approved):
This is a nice PR. Use of `numeric` type gives us nice controls over data types. I actually ran into an overflow issue last week when indexing Revert messages, so yeah I think this will help us when indexing all kinds of data from the chain. 

Left one suggestion which is more of a thought experiment. Testing steps working locally the doc comments already make the code vastly more digestable.

deekerno (approved):

rvmelkonian (approved):

PR #744, opened 2023-04-07
enhancement: compute message ID by using `MessageOut` receipt fields

### LINKED ISSUE #722: Calculate `message_id` from fields of `MessageOut` receipt
The next major version of `fuel-core` will include a change that removes the `message_id` field from `MessageOut` receipts; see https://github.com/FuelLabs/fuel-vm/pull/397. We'll need to adjust the `#[indexer(...)` macro to hash the necessary fields and compute the ID upon receiving the receipt.

deekerno:
Closes #722.

## Changelog
- Compute `message_id` from `MessageOut` receipt fields
- Fix a bug in which the message data from the receipt was being indexed but not the receipt itself
- Fix errant re-name of one of our E2E tests

## Testing Plan
CI should pass. Since the message ID is different on every pass, I couldn't add an assertion to the E2E test. But if you want to be sure that the calculated message ID is exactly the same as the one in the receipt, feel free to replace the first two lines of the `Receipt::MessageOut` match arm with the following:
```
Receipt::MessageOut { sender, recipient, amount, nonce, len, digest, data, message_id } => {
    let computed_message_id = decoder.compute_message_id(&sender, &recipient, nonce, amount, &data[..]);
    Logger::warn!(&format!("These two message IDs should match:\nmessage id from receipt: {}\ncalculated message id: {}\n", message_id, calculated_message_id));
```

Then run `cargo run --bin fuel-node` along and `cargo run --bin web-api` to start up the testing components. After that, run `cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml --verbose` to make sure logs are printed. Finally, run `curl -X POST http://127.0.0.1:8000/messageout` and inspect the logs to ensure that the IDs are the same.

### DIFF
--- a/examples/block-explorer/explorer-indexer/Cargo.toml
+++ b/examples/block-explorer/explorer-indexer/Cargo.toml
@@ -18,3 +18,4 @@ fuels-types = { version = "0.38.1", default-features = false }
 getrandom = { version = "0.2", features = ["js"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+sha2 = { version = "0.10" }
--- a/packages/fuel-indexer-macros/src/indexer.rs
+++ b/packages/fuel-indexer-macros/src/indexer.rs
@@ -96,12 +96,12 @@ fn process_fn_items(
         .iter()
         .flatten()
         .map(|typ| {
-            let message_id = typ.message_id;
+            let message_type_id = typ.message_id;
             let ty_id = typ.application.type_id;
 
             quote! {
-                #message_id => {
-                    self.decode_type(#ty_id, data.data.clone());
+                #message_type_id => {
+                    self.decode_type(#ty_id, data);
                 }
             }
         })
@@ -377,6 +377,21 @@ fn process_fn_items(
                 }
             }
 
+            fn compute_message_id(&self, sender: &Address, recipient: &Address, nonce: Bytes32, amount: Word, data: &[u8]) -> MessageId {
+
+                let raw_message_id = Sha256::new()
+                    .chain_update(sender)
+                    .chain_update(recipient)
+                    .chain_update(nonce)
+                    .chain_update(amount.to_be_bytes())
+                    .chain_update(data)
+                    .finalize();
+
+                let message_id = <[u8; 32]>::try_from(&raw_message_id[..]).expect("Could not calculate message ID from receipt fields");
+
+                message_id.into()
+            }
+
             fn decode_type(&mut self, ty_id: usize, data: Vec<u8>) {
                 match ty_id {
                     #(#decoders),*
@@ -402,7 +417,7 @@ fn process_fn_items(
                 }
             }
 
-            pub fn decode_messageout(&mut self, type_id: u64, data: abi::MessageOut) {
+            pub fn decode_messagedata(&mut self, type_id: u64, data: Vec<u8>) {
                 match type_id {
                     #(#message_types_decoders),*
                     _ => Logger::warn("Unknown message type ID; check ABI to make sure that message types are correct.")
@@ -470,7 +485,9 @@ fn process_fn_items(
                                     decoder.decode_return_type(selector, data);
                                 }
                             }
-                            Receipt::MessageOut { message_id, sender, recipient, amount, nonce, len, digest, data } => {
+                            Receipt::MessageOut { sender, recipient, amount, nonce, len, digest, data, .. } => {
+                                let message_id = decoder.compute_message_id(&sender, &recipient, nonce, amount, &data[..]);
+
                                 // It's possible that the data field was generated from an empty Sway `Bytes` array
                                 // in the send_message() instruction in which case the data field in the receipt will
                                 // have no type information or data to decode, so we decode an empty vector to a unit struct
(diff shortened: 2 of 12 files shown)


packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -402,7 +417,7 @@ fn process_fn_items(
                 }
             }
 
-            pub fn decode_messageout(&mut self, type_id: u64, data: abi::MessageOut) {
+            pub fn decode_messagedata(&mut self, type_id: u64, data: Vec<u8>) {
```
rvmelkonian, line 420: why are we not using snake_case here?

rvmelkonian (approved):
Code looks all good. I had to update the test to use these lines instead. 

```rust
let computed_message_id = decoder.compute_message_id(&sender, &recipient, nonce, amount, &data[..]);
Logger::warn(&format!("These two message IDs should match:\nmessage id from receipt: {}\ncalculated message id: {}\n", message_id, computed_message_id));
```

Which generated the expected Log:
```
2023-04-08T16:01:09.790519Z  WARN fuel_indexer::ffi: 109: These two message IDs should match:
message id from receipt: 3710b0a37187fcd21ce92323ac9dcac35af3957afff4fd4aae4fcaba388ef3f3
calculated message id: 3710b0a37187fcd21ce92323ac9dcac35af3957afff4fd4aae4fcaba388ef3f3
````

File: examples/block-explorer/explorer-indexer/Cargo.toml [hunk 2]
```diff
@@ -18,3 +18,4 @@ fuels-types = { version = "0.38.1", default-features = false }
 getrandom = { version = "0.2", features = ["js"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+sha2 = { version = "0.10" }
```
ra0x3, line 21: @deekerno It's not clear what this is used for? It's not used in the user's indexer (because I don't see a change to the explorer indexer library) so I'm imagining this was needed by the codgen? If so, this should go through the plugin preludes

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -377,6 +377,21 @@ fn process_fn_items(
                 }
             }
 
+            fn compute_message_id(&self, sender: &Address, recipient: &Address, nonce: Bytes32, amount: Word, data: &[u8]) -> MessageId {
```
ra0x3, line 380: If this is a helper function can we move it to helpers.rs ?

packages/fuel-indexer-tests/components/indices/simple-wasm/simple-wasm/Cargo.toml
```diff
@@ -17,3 +17,4 @@ fuels-macros = { version = "0.38.1" }
 fuels-types = { version = "0.38.1", default-features = false }
 getrandom = { version = "0.2", features = ["js"] }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
+sha2 = { version = "0.10" }
```
ra0x3, line 20: Same comment of all these imports, I don't think they're needed as this `sha2` is not used on the user-facing part of the indexer. If these are required for the codegen they should be made available via the plugin's prelude.

ra0x3 (commented):
@deekerno As opposed to reverting, I've just left my feedback here. You can follow up in another PR for anything that needs to be followed up on that would be 👌🏽

examples/block-explorer/explorer-indexer/Cargo.toml
deekerno: It is indeed required by the codegen. Now that you mention it, I had originally put it in the prelude, but it looks like it got blown away through a rebase and I forgot about it. In any case, wouldn't it still need to be in the dependencies as the calculation happens when the receipt arrives and the hashing comes from a crate that isn't a part of the standard library?

ra0x3: - So the only dependencies that go in an indexer's Cargo.toml are dependencies used in the user-facing part of the indexer (the codegen not being a user-facing part of the indexer because the user has no control over it outside of their own indexer handlers).
  - If I was calculating some sha2 of some byte array and creating an ID from that and assigning that ID to one of my GraphQL types - that's a case where I (as the indexer creator) would import sha2 in my indexer

- When you add a piece of code to the codegen that references an outside dependency (e.g., sha2), you add it to the plugin's Cargo.toml, then make it available via plugin's prelude (wasm and native) -- (e.g., `pub use sha2;` ).
  - This only makes the dependency available to be used by the macros - you still have to import it the dependency into the macros though

- To import this dependency from the plugin, into the indexer macros, you'll then add the dependency to the macros prelude. So the flow is (1) add dependency to plugin Cargo.toml (2) export dependency in plugin's prelude (3) import dependency in _macros_ prelude -- at which point it is available in the indexer macro now.

- [I see that there is already a sha2 available via the plugin in your PR](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-macros/src/wasm.rs#L42) as direct import. You don't want to import dependencies directly into the macros. We want to add that dependency to the _plugin_, then import that dependency from the plugin, into the macros (hope that makes sense).

CC @rvmelkonian cause that's ☝🏽 important

packages/fuel-indexer-macros/src/indexer.rs
deekerno: As far as I can tell, this has to reside on the decoder struct as it uses the values from the receipt fields to calculate the ID for the ABI type. I can create a helper function that returns a `TokenStream` that contains the function and inline it, if you like.

examples/block-explorer/explorer-indexer/Cargo.toml
deekerno: Yeah, that makes sense; I had understood it differently. I'll make the change in a follow-up PR.

PR #753, opened 2023-04-10
enhancement: support sway error enums

### LINKED ISSUE #754: Support Sway Enums
While working on #748 and creating an indexer for the Escrow contract, I ran into an error where we are unable to index sway enum types. As this is a common pattern in sway, we should provide support for this. 

• Parse out the tokes `TypeDeclarations` into a Rust enum that is available in the module using meta-programming.
• Add necessary helper functions to the codegen
• Update test contract, rebuild abi and create a new test for this.

rvmelkonian:
Closes #754 
 
For context around this PR checkout the issue. This PR adds the basic support and a test for indexing and data emitted by an Enum in sway. 

## Changelog
• Updates Test Contract to use a sway Enum type
• Updates test fixtures.
• Creates a new test. 

## Considerations 
When printing the response via the web-api you get the following receipts:
```
receipts: [
  Call {
    id: 0000000000000000000000000000000000000000000000000000000000000000,
    to: ff26a39789cdb733529ba55eb978bd2e55beb3bd41831ceb19b093b95838398d,
    amount: 0,
    asset_id: 0000000000000000000000000000000000000000000000000000000000000000,
    gas: 1000000,
    param1: 2226784473,
    param2: 69,
    pc: 11632,
    is: 11632
  },
  LogData {
    id: ff26a39789cdb733529ba55eb978bd2e55beb3bd41831ceb19b093b95838398d,
    ra: 0,
    rb: 0,
    ptr: 18072,
    len: 8,
    digest: af5570f5a1810b7af78caf4bc70a660f0df51e42baf91d4de5b2328de0e83dfc,
    data: 0000000000000000,
    pc: 12088,
    is: 11632
  },
  Revert {
    id: ff26a39789cdb733529ba55eb978bd2e55beb3bd41831ceb19b093b95838398d,
    ra: 18446744073709486080,
    pc: 12096,
    is: 11632
  },
  ScriptResult {
    result: Revert,
    gas_used: 715
  }
] 
```
When testing with other Error Enum variants in sway. The `ra` value, for the `Revert` receipt is always the same. There doesn't seem to be anyway to decode `ra` into the original Error Enum variant used in the contract source code.  

In this implementation, the data can only be indexed and captured. I don't think LogData is of much use either. 

I am hoping that when I try to use this enum indexing in the Escrow example, I will discover some more solutions about the best way to work with Enums. 

## Testing 
- CI Tests should pass.


packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
```diff
@@ -697,3 +697,31 @@ async fn test_can_trigger_and_index_panic_function_postgres() {
...
+    node_handle.abort();
+
+    let mut conn = test_db.pool.acquire().await.unwrap();
+    let row = sqlx::query("SELECT * FROM fuel_indexer_test_index1.enumerror LIMIT 1")
+        .fetch_one(&mut conn)
+        .await
+        .unwrap();
+
+    assert_eq!(row.get::<BigDecimal, usize>(0).to_u64().unwrap(), 42);
+    assert_eq!(row.get::<&str, usize>(1), EXPECTED_CONTRACT_ID);
```
deekerno, line 726: Can we add an assertion for `error_val` here?

deekerno (changes requested):
Looks good overall, left one comment.

packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
rvmelkonian: sure! Sorry missed this review somehow.

rvmelkonian: added the assert, regardless of which variant of the enum is used in the contract `error_val` is emmited as zero.

deekerno (approved):

PR #758, opened 2023-04-13
 fix: decode auth signature before recovery

ra0x3:
### Description
- PR fixes an authentication bug by decoding the message signatures before trying to derive a signature

### Testing steps
- [ ] Start a node `cargo run --bin fuel-node`
- [ ] Start the service with auth enabled `cargo  run --bin fuel-indexer -- run --run-migrations --auth-enabled --auth-strategy jwt --jwt-secret 12345677890`
- [ ] Try to authenticate `forc index auth`
- Everything should work the first time (and every subsequent time)

### Changelog
- fix: decode auth signature before recovery

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2917,6 +2917,7 @@ dependencies = [
  "fuel-indexer-lib",
  "fuel-indexer-metrics",
  "fuel-indexer-schema",
+ "hex",
  "http",
  "hyper",
  "hyper-rustls 0.23.2",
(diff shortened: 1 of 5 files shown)


deekerno (approved):
Great find. 💯

rvmelkonian (approved):
Glad you resolved this bug! Testing steps all work on my side. I tried to get it to flake by doing it about 6 times. Alas, sea of green.

PR #761, opened 2023-04-14
enhancement: add `PageInfo` object to paginated GraphQL responses

deekerno:
## Changelog
- Adds `PageInfo` object to paginated GraphQL responses
- Adjusts SQL query for paginated GraphQL requests
- Adds support for aliased fields
- Adjust tests
- Adjust doc comments
- Adds match arm in `ApiError` for `GraphQlError`

## Testing Plan
CI should pass.

### Manual Testing
0. Compile fresh WASM modules and clear database.
1. `cargo run --bin fuel-indexer -- run --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --manifest examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml`
2. Open the corresponding playground: `http://localhost:29987/api/playground/fuel_examples/explorer_indexer`
3. Run the following query.
```
query {
  blocks: block(order: { desc: height}, first: 5, offset: 5) {
    id
    hash
    height
    timestamp
  }
}
```
 You should see that the last 5 blocks are shown in reverse order of block height under a `blocks` key as well as a `PageInfo` object that contains pagination metadata. It will be similar in structure to the following example:
```
{
  "data": {
    "blocks": ...,
    "page_info": {
      "has_next_page": true,
      "limit": 5,
      "offset": 5,
      "pages": 282,
      "total_count": 1409
    }
  }
}
```

### DIFF
--- a/packages/fuel-indexer-api-server/src/uses.rs
+++ b/packages/fuel-indexer-api-server/src/uses.rs
@@ -34,6 +34,7 @@ use fuel_indexer_schema::db::{
 use hyper::Client;
 use hyper_rustls::HttpsConnectorBuilder;
 use jsonwebtoken::{encode, EncodingKey, Header};
+use serde::Serialize;
 use serde_json::{json, Value};
 use std::{
     convert::From,
@@ -42,6 +43,11 @@ use std::{
 };
 use tokio::sync::mpsc::Sender;
 use tracing::error;
+
+#[derive(Serialize)]
+pub(crate) struct QueryResponse {
+    data: Value,
+}
 
 #[cfg(feature = "metrics")]
 use fuel_indexer_metrics::{encode_metrics_response, METRICS};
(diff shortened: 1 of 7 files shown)


packages/fuel-indexer-schema/src/db/queries.rs
```diff
@@ -94,19 +96,105 @@ impl UserQuery {
...
+                        ),
+                        '{alias}', (
+                            SELECT json_agg(item)
+                            FROM (
+                                SELECT {json_selections_str} FROM selection_cte
+                                LIMIT {limit} OFFSET {offset}
+                            ) item
+                        )
+                    );"#
+                );
```
rvmelkonian, line 195: I am sure you have tested this case, but what happens the alias value is the `self.entity_name`, and then is the entity name is used in the placeholder `{'{alias}'}`, does it effect the query? Maybe it makes no difference, but perhaps we should remove this value altogether from being used, if no `alias` has been set.

deekerno: Nope, it doesn't affect the results in any way. It just sets the list of objects as a value to the key of the entity name:
```
query {
  block(order: { desc: height}, first: 5, offset: 5) {
    id
    hash
    height
    timestamp
  }
}
```
```
{
  "data": {
    "block": [
        ...
    ],
    "pageInfo": {
        ...
    }
  }
}
```

I think it should stay as any downstream users will be able to refer to the objects themselves or the `PageInfo` object with just a key. Without it, both parts would have to be elements in a list which would require users to do additional object checking on their end.

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -43,6 +44,11 @@ use std::{
 use tokio::sync::mpsc::Sender;
 use tracing::error;
 
+#[derive(Serialize)]
```
ra0x3, line 47: This should go in api/models.rs

ra0x3 (approved):
Looks ok. Real test will be Luiz testing this stuff out. Will leave feedback in the feature branch PR.

packages/fuel-indexer-api-server/src/uses.rs
deekerno: I'll move it over in the final docs/cleanup PR today. 👌🏽

PR #748, opened 2023-04-09
enhancement: create escrow example

rvmelkonian:
Closes #647 
🐳

WIP

PR #762, opened 2023-04-15
chore: bump fuels to `v0.40.0`

### LINKED ISSUE #757: Bump fuels sdk to `0.40.0`
Version 0.40.0 of the SDK is released. Check out the [release notes](https://github.com/FuelLabs/fuels-rs/releases/tag/v0.40.0). 
• Bump the versions in our toml files. 
• Update any tests and or production code if any breaking changes. 

- This should unblock #754 and #748 🥳. Thanks @segfault-magnet for the fast fix on the Parameterize / Tokenizable wasm issue.

rvmelkonian:
Closes #757 
## Context 

`fuels-rs` release a new update. As always we should stay in sync with their latest version, see the release notes [here](https://github.com/FuelLabs/fuels-rs/releases/tag/v0.40.0). This update will enable us to be able to index sway enum values and will unblock #754 and #748. 

## Changelog
• Updates SDK methods, notably using the new `Contract::load_from` and `contract_id()` methods. 
• Use the new cleaner import pattern, see [this](https://github.com/FuelLabs/fuels-rs/releases/tag/v0.40.0) PR for more context. 
• Updates `Cargo.toml` files. 
• Use the newly required generic type `<FuelIndexerTest<WalletUnlocked>`. This will change soon [see this issue](https://github.com/FuelLabs/fuels-rs/issues/925).
• Add new value for `contract_id` field in `fuel_indexer_test.yaml` (the sdk was outputting a new id after the update).

Thanks to @segfault-magnet who unblocked this. 

## Testing 
CI should pass ✅


plugins/forc-index/src/utils/defaults.rs
```diff
@@ -26,9 +26,9 @@ fuel-indexer-macros = {{ version = "0.9", default-features = false }}
 fuel-indexer-plugin = {{ version = "0.9", features = ["native-execution"] }}
 fuel-indexer-schema = {{ version = "0.9", default-features = false }}
 fuel-tx = "0.26"
-fuels = {{ version = "0.38.1" }}
-fuels-core = {{ version = "0.38.1" }}
-fuels-types = {{ version = "0.38.1", default-features = false }}
+fuels = {{ version = "0.40.0" }}
```
ra0x3, line 29: @rvmelkonian Unrelated to your PR, but can we just use the minor version (and not the patch version) `0.40` here? I think it's safe to assume the latest version is OK to use

CC @deekerno for thoughts

ra0x3 (commented):
@rvmelkonian Looks ok, left one comment. Will defer to @deekerno

plugins/forc-index/src/utils/defaults.rs
deekerno: The Rust SDK team has already moved on to `v0.41.0`, so in _this_ particular case, I don't think it matters too much. But in general, yeah, I'd also like us to use the minor version as they're quite good at making sure that patches don't contain breaking changes.

deekerno (approved):
It generally looks okay to me. 👍🏽 

The SDK has moved to v0.41.0 since you've opened this PR. It contains one breaking change in the way that our test contract would be instantiated. I don't mind if you would rather keep it on this version for now to unblock your other tasks.

plugins/forc-index/src/utils/defaults.rs
rvmelkonian: Noted, so we only want minor version bumps. I will test out the new `v0.41.0` after this one is merged.

Issue #656, opened 2023-03-15
Support GraphQL pagination


ra0x3:
- In addition to https://github.com/FuelLabs/fuel-indexer/issues/637 we need to support basic GraphQL pagination
- https://graphql.org/learn/pagination/
- Let's maybe just support `first` and `offset` for now, we can add `after` if someone requests it
- This functionality will be needed for the block explorer so high priority

PR #763, opened 2023-04-15
enhancement: add filtering, sorting, and pagination to GraphQL queries

### LINKED ISSUE #637: Add filtering to GraphQL API queries

ra0x3:
_This is deekerno speaking as ra0x3. 😉_ 

Closes #637, #656.

## Changelog

- Add `arguments` module for turning query arguments into the new `QueryParams` struct
- Add functionality and tests for the following filter types:
-- `id`: select a record by its ID
-- `and` / `or`: combine two filters together
-- `not`: negate the behavior of a filter
-- `has`: select records that have non-null values for columns in list
-- `in`: select records that have a value matching an element in set
-- `gt` / `gte` / `lt` / `lte` (greater than, etc.): select records whose value satisfies the comparison condition
-- `between`: select records whose value falls between the bounds of the comparison condition
- Add sorting and pagination functionality and tests
- Move stuff specific to GraphQL API queries to fuel-indexer-schema/db/queries.rs
- Add additional variants to GraphQLError
- Update documentation for GraphQL queries and new functionality
- Added support for field aliasing in GraphQL queries
- Add `fuel-indexer-graphql` and `fuel-indexer-graphql-parser` packages

## Testing Plan
I've added extensive testing (15 new tests in total) for filtering and pagination, and unit tests have been adjusted where necessary. CI should pass.

### Manual Testing
Feel free to peruse changes from the documentation PR for example queries that you can use for the block explorer indexer; there are pages for [filtering](https://github.com/FuelLabs/fuel-indexer/blob/7e2e798feebd70babf3b5d1caa12b7b59f120c9d/docs/src/reference-guide/components/graphql/queries/search-filtering.md) and [pagination](https://github.com/FuelLabs/fuel-indexer/blob/7e2e798feebd70babf3b5d1caa12b7b59f120c9d/docs/src/reference-guide/components/graphql/queries/pagination.md) as well as [this page](https://github.com/FuelLabs/fuel-indexer/blob/7e2e798feebd70babf3b5d1caa12b7b59f120c9d/docs/src/reference-guide/components/graphql/queries/full-example.md) that contains a fully-featured example of all of the GraphQL querying functionality being used together (nested entities, sorting, pagination, and filtering).


rvmelkonian (dismissed):
The code looks really readable and easy to reason about. I've tested things in the playground and couldn't find any bugs. We reviewed the docs already. So LGMT 🚀

deekerno (approved):
🥷

rvmelkonian (approved):
🚢

PR #765, opened 2023-04-17
fix: run migrations in docker compose

ra0x3:
### PR 
- PR adds `--run-migrations` to the docker compose `run` commands

### Testing steps
- [ ] Compose up the examples

### Changelog 
- fix: run migrations in docker compose


rvmelkonian:
mdbook build job has failed.

deekerno (approved):
Should we change the Fuel node host to beta-3?

ra0x3:
> Should we change the Fuel node host to beta-3?

@deekerno Yes (I believe we moved everything to beta-3 already). I'll make that change

PR #768, opened 2023-04-17
docs: add docs for GraphQL filtering and pagination

deekerno:
## Changelog
- Add documentation for GraphQL filtering and pagination
- Move `QueryResponse` to `fuel-indexer-api-server/models.rs`

## Testing Plan
CI should pass.


docs/src/reference-guide/components/graphql/pagination.md
```diff
@@ -0,0 +1,127 @@
...
+
+The Fuel indexer currently supports offset-based pagination in order to allow users to selectively request parts of a set of results.
+
+A paginated query can be made using three keywords:
+
+- `order` - sort order (required)
+- `first` - limit on number of results (required)
+- `offset` - the amount of records to skip before returning results (optional)
+
+In order to generate a paginated response, your query should include a limit and either an ascending or descending sort order applied to one of the fields of the searched entity. _If you include a limit in your query, you are required to include a sort order as well._ Otherwise, your query will return an error.
```
rvmelkonian, line 11: This is obviously correct, but I can't help but feel that with the lines 7,8 and 9, it explains all and any developer should understand the idea of `order`, `first` and `offset` params. It would read just as well (perhaps better?) without this paragraph altogether.

docs/src/reference-guide/components/graphql/queries.md
```diff
@@ -19,16 +19,19 @@ query {
 
 The `entity` field corresponds to the name of an entity defined in your [schema](./schema.md) and the sub-fields are the fields defined on that entity type; entities and fields are stored in the database using the names defined in the schema, so make sure that your query uses those same names as well.
 
-Let's refer back to the [block explorer](../../../examples/block-explorer.md) example for an illustration. After the block data has been indexed, we can retrieve information about the blocks by sending a query to the graph endpoint for our indexer.
+Let's refer back to the [block explorer](../../../examples/block-explorer.md) example for an illustration. After the block data has been indexed, we can retrieve information about the blocks by querying for information on the explorer indexer's playground; you can get to the playground by starting the block explorer example using the instructions on the page and navigating to `http://localhost:29987/api/graph/fuel_examples/explorer_indexer`.
```
rvmelkonian, line 22: Suggestion:
"by querying information on the indexer's playground ..."
(reader knows we are talking about the block explorer in this paragraph already).

docs/src/reference-guide/components/graphql/queries.md
```diff
@@ -137,16 +140,7 @@ query {
...
-
-```txt
-curl -X POST http://localhost:29987/api/graph/fuel_examples/nested_query_index \
-   -H 'content-type: application/json' \
-   -d '{"query": "query { character { name book { title library { name city { name } } } } }", "params": "b"}' \
-| json_pp
-```
-
-And we receive the following response:
+Let's assume that we've created an indexer for this data and indexed data has been stored in the database. If we send the query, we'll get the following response:
```
rvmelkonian, line 143: A missing 'the'

"... and "the" indexed data has been stored in the database."

docs/src/reference-guide/components/graphql/search-filtering.md
```diff
@@ -0,0 +1,297 @@
+# Search and Filtering
+
+Currently, the Fuel indexer supports the following search and filtering operations:
```
rvmelkonian, line 3: Suggested change:

"The Fuel indexer currently supports the following search and filtering operations:"

docs/src/reference-guide/components/graphql/search-filtering.md
```diff
@@ -0,0 +1,297 @@
...
+      "baz": 1000,
+      "foo": "blorp",
+      "id": 3,
+      "possibly_null_bar": 456
+    }
+  ]
+}
+```
+
+> Note: It is important to note that you should practice sensible database design when filtering records. Apply database indicies to the underlying columns in order to make search operations more efficient; however, be advised that an overuse of database indicies will lead to degraded performance.
```
rvmelkonian, line 50: Suggested change:
"You should practice sensible database design when filtering records."
This means we avoid repetition of the word "note" here.

docs/src/reference-guide/components/graphql/search-filtering.md
```diff
@@ -0,0 +1,297 @@
...
+
+## ID Selection
+
+You can query for a particular instance of an object by passing an ID value to with the `id` key. For example, let's say that you were storing blocks and you wanted details about a certain block, you would pass the requisite ID:
```
rvmelkonian, line 54: Suggested change:
"For example, let's say that you are storing blocks and you want details about a certain block, you would pass the requisite ID"

Just a feeling that the present continuous tense  (are storing, want) makes the sentence more immediate and relevant to the reader in the present context.

docs/src/reference-guide/components/graphql/search-filtering.md
```diff
@@ -0,0 +1,297 @@
...
+    }
+  ]
+}
+```
+
+> Note: Remember that IDs currently must be of type `u64`, and as such, the ID selection operation will only allow for a `u64` value. We're open to changing this in the future.
+
+## Excluding Null Values
+
+You can store null values in your records if the corresponding entity fields are configured to allow for it. As such, you can exclude records that contain null values in a particular column or set of coulmns by using the `has` operator inside of a `filter` object.
```
rvmelkonian, line 84: I don't mean to diminish your style! But I honestly feel this sentence reads better like:
"You can exclude records that ..."
Removing the "As such,"

docs/src/reference-guide/components/graphql/search-filtering.md
```diff
@@ -0,0 +1,297 @@
...
+      "id": 3,
+      "possibly_null_bar": 456
+    }
+  ]
+}
+```
+
+## Set Membership
+
+Additionally, you can exclude records in which a particular column's value does not any elements in a given set by using the `in` operator as part of a `filter` object.
```
rvmelkonian, line 120: Missing "contain"
"... column's value does not [contain] any elements ..."

docs/src/reference-guide/components/graphql/search-filtering.md
```diff
@@ -0,0 +1,297 @@
...
+
+### Less, Greater, and Equals
+
+You can do simple value comparisons using any of the following operators:
+
+- equals: `equals`
+- greater than: `gt`
+- greater than or equal to: `gte`
+- less than: `lt`
+- less than or equal to: `lte`
```
rvmelkonian, line 164: Maybe we should use table markdown here?
```
|   - equals:    | `equals`                 |
|   - greater than: | `gt`                     |
|   - greater than or equal to: | `gte`  |
|   - less than: | `lt`                        |
|   - less than or equal to: | `lte`   |
```

rvmelkonian (changes requested):
Great work on these docs. Two tiny mistakes and some style suggestions which you can take or leave, just things that occurred to me as I was reading.

ra0x3:
@deekerno 
- As @rvmelkonian thanks for the docs.
- None of my suggestions/requests are blockers for merge into the feature branch
- But yea in general I think we will want to:
  - Use _actual_ data in the examples (not _foo_ and fake variables like that)
  - Include one simple example and one complex example for _all_ operations (filter, pagination, etc)
- In general I would personally cut down on the verbosity (devs literally don't read the docs, as evidenced by my experience with Luiz yesterday 😆 )
- Thanks for the work 🆗

deekerno:
@ra0x3 @rvmelkonian Thanks for the reviews! I'll definitely reduce the verbosity as I can definitely start to sound academic after awhile. I'll also adjust the examples to use the block explorer example; I lazily lifted the fake example from the test cases that I wrote for all of this, but I definitely agree that a more relevant example would be more helpful.

I'll make the changes and re-ping before merging this into the feature branch.

ra0x3 (approved):
Will of course do a final review before merging into master. I think it looks good enough for the feature branch.

PR #772, opened 2023-04-18
fix: correct usage of `fuel-indexer` feature flags

ra0x3:
### Description
- Dockerized builds weren't working due to incorrect feature flagging
  - The main fuel-indexer Docker process was exiting with status `0` -- which is success -- and exiting within a second or two
    - This gave me the idea that some coroutine wasn't being awaited
    - From there I noticed that the `--no-default-features` flag resulted in weird behavior
      - If you built `fuel-indexer` with `--no-default-features` it would do the same `exit status 0` as Docker
        - Turns out this was due to the fact that `--no-default-features` wasn't including the `api-server` feature thus was not creating `channels` where futures were awaited
         - We never saw this because I don't think anyone has ever actually passed `--no-default-features` to `fuel-indexer` ever 😅 
- I'm OK with changing this (as a separate issue) -- i.e., using some other means of keeping our service running when not using the GraphQL API
  - Opened an issue for this https://github.com/FuelLabs/fuel-indexer/issues/776
Again, so far we haven't seen much (if any) usage of the service without the `api-server` feature so this bug would've been largely hidden from users
- We need to add some CI for this https://github.com/FuelLabs/fuel-indexer/issues/773
- PR also swaps a lot of `expect` for `?`
- Updates tests for the fix

### Testing steps

⚠️ The testing steps are a bit involved (especially the Docker steps) but please them fully so we can ensure this breakage doesn't happen again

### Ensure all features build properly
- [ ] Build for with `fuel-core-lib` dep with default features
  - `cargo build -p fuel-indexer --features fuel-core-lib`
- [ ] Build for `fuel-core-lib` dep without default features
  - `cargo build -p fuel-indexer --no-default-features --features fuel-core-lib`
- [ ] Build without `fuel-core-lib` dep
  - `cargo build -p fuel-indexer --default-features`
- [ ] Build without default features but still include `api-server` feature
  - `cargo build -p fuel-indexer --no-default-features --features api-server`

### Ensure binary runs properly
- [ ] If compiled with `--features fuel-core-lib` pass the `--local-fuel-node` flag to the binary to ensure that the local node starts
  - `./target/release/fuel-indexer run --run-migrations --local-fuel-node`
- [ ] If compiled with `--features fuel-core-lib` don't pass the `--local-fuel-node` flag to the binary to ensure that a local node is _not_ started
  - `./target/release/fuel-indexer run --run-migrations`
- [ ] If built without default features `--no-default-features` you should _not_ be able to access the GraphQL api when running `./target/release/fuel-indexer run --run-migrations`

### Ensure dockerized examples work
- [ ] The dockerized block explorer should work
  - Stop any running postgres instance
  - Build a local docker image `docker build -t fuel-indexer/local:latest -f deployment/Dockerfile .` 
  - `cd examples/block-explorer && docker compose up`
  - Do a deploy as test `forc-index deploy --path examples/block-explorer/explorer-indexer --target-dir /Users/rashad/dev/repos/fuel-indexer --url http://0.0.0.0:29987`
  - Health endpoint should work `curl http://0.0.0.0:29987/api/health`
- [ ] Dockerized hello-world should work
  - Stop any running postgres instance
- [ ] The hello-world-native example should work as well (instructions are in the examples lib.rs file)

### Changelog 
- fix: correct usage of `fuel-indexer` feature flags


deekerno (commented):
Code looks good to me and the non-Docker test steps work as described. However, I can't get any Docker stuff to successfully compile for my M1, so I'll defer to @rvmelkonian and I'll work on fixing that once and for all.

rvmelkonian:
When I run `cargo build -p fuel-indexer --features fuel-core-lib`
```
hread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at /usr/local/opt/llvm/lib/libclang.dylib could not be opened: dlopen(/usr/local/opt/llvm/lib/libclang.dylib, 0x0005): tried: '/usr/local/opt/llvm/lib/libclang.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/local/lib/libclang.dylib' (no such file), '/usr/lib/libclang.dylib' (no such file), '/usr/local/Cellar/llvm/15.0.7_1/lib/libclang.dylib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')), '/usr/local/lib/libclang.dylib' (no such file), '/usr/lib/libclang.dylib' (no such file)"', /Users/rvm/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bindgen-0.64.0/./lib.rs:2393:31
```
my zsh config:
```
export PATH="/usr/local/opt/llvm/bin/:$PATH"
export PATH="/opt/homebrew/opt/lua/bin/:$PATH"
export AR=/usr/local/opt/llvm/bin/llvm-ar
export CC=/usr/local/opt/llvm/bin/clang
export LIBCLANG_PATH=/usr/local/opt/llvm/lib
export LDFLAGS=/usr/local/opt/llvm/lib
export CPPFLAGS=/usr/local/opt/llvm/include
```
If I change these paths, then `cargo run --bin fuel-indexer -- run` won't run.

ra0x3:
@rvmelkonian does that command work on master for you ? 

Also, since @deekerno confirmed the non/docket changes worked for him, could you just confirm the docket changes work. ? 

I see the clang error in the error you posted, which looks slightly dependency (system dependency) related

rvmelkonian:
@ra0x3 It also occurs on master. I've never actually run the `fuel-indexer` command with the `----features fuel-core-lib`. Let me try change some paths in my shell.

deekerno (approved):

PR #778, opened 2023-04-19
enhancement: Add graphQL info output when indexer runs

### LINKED ISSUE #766: Add link to GraphQL playground to stdout when service is started
- A user should be able to click the link and be taken to the playground in their browser

rvmelkonian:
Closes #766 

Fairly self-explanatory. 

To test run :
```
cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml
```
You should see the output 
```
2023-04-19T15:07:46.654923Z  INFO fuel_indexer::commands::run: 70: GraphQL Playground: http://localhost:29987/api/playground/fuel_indexer_test/index1
```


packages/fuel-indexer/src/service.rs
```diff
@@ -226,6 +226,8 @@ async fn create_service_task(
             match rx.try_recv() {
                 Ok(service_request) => match service_request {
                     ServiceRequest::AssetReload(request) => {
+                        info!("GraphQL Playground: http://localhost:29987/api/playground/{}/{}", &request.namespace, &request.identifier);
```
ra0x3, line 229: We just need to make the link clear on startup, not really here

packages/fuel-indexer/src/commands/run.rs
```diff
@@ -67,6 +67,10 @@ pub async fn exec(args: IndexerArgs) -> anyhow::Result<()> {
         Manifest::from_file(&p).unwrap()
     }) {
         Some(m) => {
+            info!(
```
ra0x3, line 70: We shouldn't make the link dependent on whether the user included a manifest or not, we should do it regardless (maybe just above this line)

ra0x3 (changes requested):
@rvmelkonian 

We just want to show the link once when the service starts (or just before it starts).Might help to add a cool little emoji or two (helps draw your attention). We want this link to be super obvious to see when running without `--verbose`

We also want the same functionality for native execution as well

packages/fuel-indexer/src/commands/run.rs
rvmelkonian: @ra0x3 I've added emojis, and now show two different urls depending on the match arm. This line seems to be the earliest point in the program that I have access to the manifest values (without adding more logic to parse the PathBuf and read the file, which seemed unnecessary).

`exec` is first function  called after `run_cli()`. Let me know your thoughts.

ra0x3 (approved):
LGTM! :shipit:

PR #797, opened 2023-04-24
enhancement: update page_info gql response

ra0x3:
### Description
- PR adds a query param to the `/api/graph/:namespace/:identifier` route
- Query param is `page_info` and denotes whether or not metadata related to the page should be returned -- even if we request pagination
- Also refactors a bit of the web models used for the GraphQL stuff to make things a bit more Rusty

### Testing steps
- A test was updated so CI should pass
- [ ] Start explorer on beta-3  `cargo run --bin fuel-indexer -- run --run-migrations --manifest examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml --fuel-node-host beta-3.fuel.network --fuel-node-port 80`
- [ ] Run a non-paginated query, that returns a `PageType::Plain`

```graphql
query {
  block {
    id
    height
    timestamp
  }
}
```
- You should get back a payload similar to `{ data: [ data_is_here ] }`
- Even if you include `?page_info=true` in your playground URL (in the UI) the response structure should not change

- [ ] Run a query for paginated data that returns `PageType::Paginated`
```graphql
query {
  block(order: { asc: height }, first: 5) {
    hash
    id
  }
}
```
  - Response should return a page without any page metadata: `{ data: [ my_data ] }`
- [ ] Now run the same pagination query but update the GraphQL endpoint (in the UI) with the prefix: `?page_info=true`
  - Response should include page metadata `{ data: [my_data],  has_next_page: true, limit: 5, offset: 0, pages: 367, total_count: 1833 }`

### Changelog 
- enhancement: update page_info GraphQL response


ra0x3:
@deekerno Got it. Could we update the issue with this info as well (for anyone else who stumbles across it)?

deekerno:
This should probably be closed as I'll probably be integrating more of `async-graphql` into the GraphQL parts of the codebase, which would cause a lot of mess in this PR.

ra0x3:
@deekerno I closed this one. If you get back to this part of the codebase before me, let's try to implement these `From<>` and `To<>`. I know there's at least 2 `.clone()` we're doing in the hot path that can be removed with these conversions.

PR #800, opened 2023-04-25
enhancement: support multiple queries in a single request

### LINKED ISSUE #792: Allow for multiple queries inside of a GraphQL API request
Currently, our GraphQL API does not allow for multiple queries inside of the body of a request. For example, the following request will not succeed despite being totally valid GraphQL functionality:
```
query {
    block {
        id
    }
}

query {
    tx {
        id
    }
}
```

We should amend the API response to allow for multiple query result sets.

rvmelkonian:
Closes #792 

See the original issue for context around this feature. 

## Changelog 
• Adds a new function `async fn process_queries`

## Testing 
1. `cargo run --bin fuel-node`
2. `cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml`
3. Run the query in the original issue 
```bash
curl -X POST -H "Content-Type: application/json" -d '{"query": "query { block { id } } query { tx { id } }"}' http://localhost:29987/api/graph/fuel_indexer_test/index1
```
You should receive a response
```
{"data":{"block":{"data":[{"id":3546081342346770273},{"id":3474075630728078385}]},"tx":{"data":[{"id":7149246036009629541},{"id":4123389826724803897},{"id":3834875793769705525},{"id":7075492784554664246}]}}}
```

## Considerations 
I have tested this via postman and get the expected response, however when testing this inside the playground you get a response : 
```
{
  "data": {}
}
```

I originally tried to parse the request as a string and with a HashSet run the queries seperately, however this returned graphql erros from the endpoint, so I implemented `process_queries` using the types from `fuel_indexer_graphql::graphql:` and `fuel_indexer_graphql_parser::query:`, this worked. 

This will be a hotpath, I would hope that these alterations wouldn't slow down the response noticeably, but until we do real profiling it's difficult to tell.


rvmelkonian:
Closing this as `async-graphql` provides the functionality and we will be switching to this.

PR #802, opened 2023-04-25
fix: log schema api error

ra0x3:


rvmelkonian (approved):
Makes sense, we have no negative test cases covering these errors, otherwise we would have caught this.

PR #808, opened 2023-04-26
bugfix: `stop-previous` flag to remove wasm binary by `index_id`

### LINKED ISSUE #786: Allow `forc index deploy` to automagically stop the current version of the indexer
- We have the ability to deploy new indexers to an indexer service via `forc index deploy`
- When this new indexer is deployed, it starts to index data immediately
- However, any previous version of the indexer that has not been explicitly stopped, will continue to run
- Given that indexer migrations https://github.com/FuelLabs/fuel-indexer/issues/382 don't really make the most sense given our setup, the main way to "migrate data" is to update your schema, rebuild your indexer, then redeploy 
- This is simpler than it seems: you just edit your schema file then do another `forc index deploy`
- However, in order to ensure that users can deploy this new indexer without having to go back and manually stop an older version, we should add some flag to `forc index deploy` that will automagically stop any indexer running under that `namespace/identifier` combination
- Something like `forc index deploy --stop-previous` would upload the new indexer and stop any previous version
  - Call the flag whatever you want that makes sense

rvmelkonian:
Closes #786 

Note that the changes to the plugin side were already merged to master, and then reverted. The changes to the plugin are on this branch, but the diff doesn't show because of the previous merge. 

## Changelog
• Adds conditional logic based on query param to `remove_indexer`.
• Adds `remove_previous_indexer` queries function.
• Adds metrics for new query function.  

## Testing 
1. Build the new plugin binary `cargo build -p forc-index -p fuel-indexer --release --features fuel-core-lib`.
2. Refresh the DB and rebuild test binaries. 
3. Start local fuel-node.
4. Start the indexer.
5. Deploy the block explorer `./target/release/forc-index deploy --path examples/block-explorer/explorer-indexer `
6. Run the SQL query `select * from index_asset_registry_manifest;` 
   - Make a mental note of the single wasm binary in the DB. 👀 
7. Make some small change in the `lib.rs` of the block explorer.
8. Deploy the block explore with the new flag `./target/release/forc-index deploy --stop-previous --path examples/block-explorer/explorer-indexer`
 9. Run the same SQL query `select * from index_asset_registry_manifest;`
 
You should see that there is still only one wasm binary with the same index_id of 1. The old one was removed.


examples/block-explorer/explorer-indexer/explorer_indexer.manifest.yaml
```diff
@@ -7,6 +7,6 @@ graphql_schema: examples/block-explorer/explorer-indexer/schema/explorer_indexer
 module:
   wasm: target/wasm32-unknown-unknown/release/explorer_indexer.wasm
 metrics: ~
-contract_id: ~
+contract_id: ""
```
ra0x3, line 10: This is that bug that needs to be fixed in that separate issue

examples/block-explorer/explorer-indexer/src/lib.rs
```diff
@@ -73,7 +73,7 @@ mod explorer_index {
             match &tx.transaction {
                 #[allow(unused)]
                 Transaction::Script(t) => {
-                    Logger::info("Inside a script transaction. (>^‿^)>");
+                    Logger::info("Inside a script transaction./././ (>^‿^)>");
```
ra0x3, line 76: Do these additional characters serve a purpose?

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -119,16 +120,38 @@ pub(crate) async fn health_check(
...
+        let _ = queries::start_transaction(&mut conn).await?;
+
+        if let Err(e) =
+            queries::remove_previous_indexer(&mut conn, &namespace, &identifier).await
+        {
+            queries::revert_transaction(&mut conn).await?;
+
+            error!("Failed to remove Indexer({namespace}.{identifier}): {e}");
+
+            return Err(ApiError::Sqlx(sqlx::Error::RowNotFound));
```
ra0x3, line 145:
```suggestion
            return Err(ApiError::default());
```

Since we don't know that the error was `RowNotFound` for sure

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,6 +232,19 @@ pub(crate) async fn register_indexer_assets(
         while let Some(field) = multipart.next_field().await.unwrap() {
             let name = field.name().unwrap_or("").to_string();
             let data = field.bytes().await.unwrap_or_default();
+
+            if params.stop_previous.unwrap_or(false) {
```
ra0x3, line 236: This should be _after_ I've already successfully uploaded. Else if there is an issue with my upload you will have already removed my already running indexer.

Also, this logic should be outside of the `multipart` loop. Else you will try to run this logic for each multipart field.

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -841,6 +841,38 @@ pub async fn remove_indexer(
     Ok(())
 }
 
+pub async fn remove_previous_indexer(
```
ra0x3, line 844: Can we be a little more specific with the naming? There can be multiple "previous indexers". This function seems to remove the most recent version of the indexer. Can we update the naming to reflect that?

ra0x3 (changes requested):
Left some feedback. Did not test. Will try that next time.

@rvmelkonian In the mean time, can we update the PR with manual testing steps for when its ready to be tested?

PR #824, opened 2023-04-27
enhancement: add end_block field in manifest

### LINKED ISSUE #796: Add an `end_block` field in manifest
Brought up by @ra0x3 in team sync today, users should be able to add an `end_block` field, to determine when an indexer should stop indexing. Without this, all indexers by default will run and listen for events on new blocks indefinitely. 

• Add a new field in the manifest file, update `Manifest` struct and add the logic necessary to achieve this functionality. 

• Additionally, if no `end_block` is set, we should display a very clear warning message when the indexer starts, notifying users that this index will run until explicitly stopped. 

• In the PR an e2e test and documentation should be included.

rvmelkonian:
Closes #796 

## Changes
• Updates `executor.rs` with logic needed for `end_block`
• Updates `Manifest`
• Adds new tests. 
• Updates docs and example manifest files. 

CI with new test cases added should all pass.


packages/fuel-indexer/src/executor.rs
```diff
@@ -72,6 +72,7 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
     kill_switch: Arc<AtomicBool>,
 ) -> impl Future<Output = ()> {
     let start_block = manifest.start_block.expect("Failed to detect start_block.");
+    let end_block = manifest.end_block; // Value can be none.
```
ra0x3, line 75: Can remove this comment (I think the code is pretty clear)

packages/fuel-indexer/src/executor.rs
```diff
@@ -131,6 +132,15 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
 
             let mut block_info = Vec::new();
             for block in results.into_iter() {
+                if let Some(end_block) = end_block {
+                    if block.header.height.0 > end_block {
+                        info!("Stopping indexer at the specified end_block: {end_block}");
+                        break;
+                    }
+                } else {
+                    warn!("No end block is set. Indexer will run forever!");
```
ra0x3, line 141: This warning will print for every block? I don't think we want that. We can remove it (we can add the warning somewhere else, because I do think we should warn when indexers run forever).

packages/fuel-indexer/src/executor.rs
```diff
@@ -131,6 +132,15 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
 
             let mut block_info = Vec::new();
             for block in results.into_iter() {
+                if let Some(end_block) = end_block {
+                    if block.header.height.0 > end_block {
```
ra0x3, line 136: 👌🏽

ra0x3 (commented):
[redacted]

packages/fuel-indexer/src/executor.rs
deekerno, line 75: You can remove the comment here.

packages/fuel-indexer/src/executor.rs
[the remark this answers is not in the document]
rvmelkonian: Fixing this!

deekerno (approved):

ra0x3:
@deekerno @rvmelkonian Just FYI this PR forgot to add the `end_block` to the default `forc index` manifest. Let's fix that in a quick little follow up 🙏🏽

rvmelkonian:
@ra0x3 Noted thanks for the reminder, will put that up in a bit.

PR #828, opened 2023-04-27
bugfix: fix deserialize bug on contract_id

rvmelkonian:
Closes #827 

## Testing steps 
1. Rebuild wasm binaries and refresh the db. 
2. Rebuild the plugin binary `cargo build -p forc-index -p fuel-indexer --release --features fuel-core-lib`
3. Run the indexer and run the local fuel-node.
3. Run the command (this command before was producing the `""` in the yaml). 
`./target/release/forc-index deploy --path examples/block-explorer/explorer-indexer`
4. In `explorer_indexer.manifest.yaml` check that `contract_id: ~` and *not* `contract_id: ""`


ra0x3 (approved):
utACK

PR #844, opened 2023-05-02
chore: use `[default-members]` section in `Cargo.toml`

### LINKED ISSUE #842: Move indexer examples into their own workspace
We should move indexer examples into their own workspace so that users can run `cargo build` in the repository root without issues relating to different targets.

deekerno:
Closes #842.

## Changelog
- Introduce `[default-members]` section
- Change member inclusion to be more idiomatic

## Testing Plan
CI should pass.

## Notes
This change was done to prevent the issue of failed builds when running `cargo build` in the repository root. This structure comes from the [`axum` project](https://github.com/tokio-rs/axum/blob/main/Cargo.toml).

This change also reduces the overall build time of a bare `cargo build`, but makes changes to using the internal testing components. To use internal testing components, a dev should use `cargo run -p [package] --bin [package]`. So in this case of `fuel-node`, one would start it by using `cargo run -p fuel-node --bin fuel-node`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,23 @@ members = [
    "plugins/forc-postgres",
    "plugins/forc-index-tests",
 ]
+default-members = [
+   "packages/fuel-indexer",
+   "packages/fuel-indexer-api-server",
+   "packages/fuel-indexer-database",
+   "packages/fuel-indexer-database/database-types",
+   "packages/fuel-indexer-database/postgres",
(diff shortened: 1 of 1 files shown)


ra0x3 (changes requested):
@deekerno Looks like some testing components aren't working `cargo run --bin fuel-node`

deekerno:
Added a change, but then reverted it and added a description to the initial comment instead, so this is good to review again.

PR #845, opened 2023-05-02
enhancement: convert Scalar types into `Json` and `Blob`

### LINKED ISSUE #843: Allow GraphQL types to be converted into `Json` and `Blob` scalars
- An ask from @Braqzen 
- We need a way to easily convert one of our GraphQL scalar types into either `Json` or `Blob`
- We should either `#[derive]` these if possible, or add a manual `impl`
- Since we know the sizes of these objects already, this should be fairly straightforward to implement and test
- [`Json` type](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-types/src/lib.rs#L35)
- [`Blob` type](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-types/src/lib.rs#L32)

```rust
struct Foo {
  value: u64,
}

let foo = Foo { value: 1 };
let as_bytes: Vec<u8> = foo.into();
let as_json: Json = foo.into();

assert_eq!(as_json, "{"value":1}".to_string());
assert_eq!(as_bytes, [1]); // or whatever
```

rvmelkonian:
Closes #843

## Changelog
• Adds the macros`json_impl` and `blob_impl` macros for custom Scalar types. 
• Adds unit tests testing the underlying types of our scalar types.
• Changes the type of Blob from an alias to a custom type (see note below about this)
• Updates code elsewhere to integrate the Blob type. 

## NB
I had to make Blob a custom type as apposed to just an alias as it was before. This is because I could not implement a custom `From` trait on an alias of underlying type`Vec<u8>`. But when `Blob is :
```rust
pub struct Blob(pub Vec<u8>);
```
I can implement the From trait. This is because of the orphan rule.

##Testing
This unit test wont run in CI. I don't think we need it to necessarily, but to run the tests cd into 
`packages/fuel-indexer-types/src/` and run `cargo test` this will run the newly created unit tests in the `lib.rs` file

You want to see: 
```
running 8 tests
test tests::test_into_json_blob_int16 ... ok
test tests::test_into_json_blob_id ... ok
test tests::test_into_json_blob_int4 ... ok
test tests::test_into_json_blob_int8 ... ok
test tests::test_into_json_blob_timestamp ... ok
test tests::test_into_json_blob_uint16 ... ok
test tests::test_into_json_blob_uint4 ... ok
test tests::test_into_json_blob_uint8 ... ok
```

### DIFF
--- a/examples/hello-world-native/hello-indexer-native/src/main.rs
+++ b/examples/hello-world-native/hello-indexer-native/src/main.rs
@@ -74,7 +74,7 @@ mod hello_world_native {
 
                 // Here we show an example of an arbtrarily sized Blob type. These Blob types
                 // support data up to 10485760 bytes in length
-                visits: Blob(vec![1u8, 2, 3, 4, 5, 6, 7, 8]),
+                visits: vec![1u8, 2, 3, 4, 5, 6, 7, 8].into(),
             },
         };
 
--- a/examples/hello-world-native/hello-indexer-native/src/main.rs
+++ b/examples/hello-world-native/hello-indexer-native/src/main.rs
@@ -74,7 +74,7 @@ mod hello_world_native {
 
                 // Here we show an example of an arbtrarily sized Blob type. These Blob types
                 // support data up to 10485760 bytes in length
-                visits: vec![1u8, 2, 3, 4, 5, 6, 7, 8],
+                visits: Blob(vec![1u8, 2, 3, 4, 5, 6, 7, 8]),
             },
         };
 
--- a/examples/hello-world/hello-indexer/src/lib.rs
+++ b/examples/hello-world/hello-indexer/src/lib.rs
@@ -80,7 +80,7 @@ mod hello_world_indexer {
 
                 // Here we show an example of an arbtrarily sized Blob type. These Blob types
                 // support data up to 10485760 bytes in length
-                visits: vec![1u8, 2, 3, 4, 5, 6, 7, 8],
+                visits: Blob(vec![1u8, 2, 3, 4, 5, 6, 7, 8]),
             },
         };
 
--- a/examples/hello-world/hello-indexer/src/lib.rs
+++ b/examples/hello-world/hello-indexer/src/lib.rs
@@ -80,7 +80,7 @@ mod hello_world_indexer {
 
                 // Here we show an example of an arbtrarily sized Blob type. These Blob types
                 // support data up to 10485760 bytes in length
-                visits: Blob(vec![1u8, 2, 3, 4, 5, 6, 7, 8]),
+                visits: vec![1u8, 2, 3, 4, 5, 6, 7, 8].into(),
             },
         };
 
--- a/packages/fuel-indexer-schema/src/lib.rs
+++ b/packages/fuel-indexer-schema/src/lib.rs
@@ -130,9 +130,7 @@ impl FtColumn {
                 );
                 FtColumn::Timestamp(Some(int8))
             }
-            ColumnType::Blob => {
-                FtColumn::Blob(Some(fuel_indexer_types::Blob(bytes[..size].to_vec())))
-            }
+            ColumnType::Blob => FtColumn::Blob(Some(bytes[..size].to_vec().into())),
             ColumnType::ForeignKey => {
                 panic!("ForeignKey not supported for FtColumn.");
             }
@@ -262,9 +260,9 @@ impl FtColumn {
                 None => String::from(NULL_VALUE),
             },
             FtColumn::Blob(value) => match value {
-                Some(fuel_indexer_types::Blob(val)) => {
-                    let x = hex::encode(val);
-                    format!("'{}'", x)
+                Some(blob) => {
+                    let x = hex::encode(blob.as_ref());
+                    format!("'{x}'")
                 }
                 None => String::from(NULL_VALUE),
             },
--- a/packages/fuel-indexer-schema/src/lib.rs
+++ b/packages/fuel-indexer-schema/src/lib.rs
@@ -130,7 +130,9 @@ impl FtColumn {
                 );
                 FtColumn::Timestamp(Some(int8))
             }
-            ColumnType::Blob => FtColumn::Blob(Some(bytes[..size].to_vec())),
+            ColumnType::Blob => {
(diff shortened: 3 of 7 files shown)


examples/hello-world-native/hello-indexer-native/src/main.rs
```diff
@@ -74,7 +74,7 @@ mod hello_world_native {
 
                 // Here we show an example of an arbtrarily sized Blob type. These Blob types
                 // support data up to 10485760 bytes in length
-                visits: vec![1u8, 2, 3, 4, 5, 6, 7, 8],
+                visits: Blob(vec![1u8, 2, 3, 4, 5, 6, 7, 8]),
```
ra0x3, line 77: Should not have to wrap a `Blob` here. Using proper `impl` we can just `.into()` this

packages/fuel-indexer-schema/src/lib.rs
```diff
@@ -260,9 +262,9 @@ impl FtColumn {
                 None => String::from(NULL_VALUE),
             },
             FtColumn::Blob(value) => match value {
-                Some(val) => {
+                Some(fuel_indexer_types::Blob(val)) => {
                     let x = hex::encode(val);
-                    format!("'{x}'")
+                    format!("'{}'", x)
```
ra0x3, line 267: I think the previous way is preferred https://rust-lang.github.io/rust-clippy/master/

packages/fuel-indexer-types/src/lib.rs
```diff
@@ -29,11 +29,46 @@ pub type UInt16 = u128;
 pub type Timestamp = u64;
 pub type Charfield = String;
 pub type Boolean = bool;
-pub type Blob = Vec<u8>;
+
+#[derive(Deserialize, Serialize, Clone, Eq, PartialEq, Debug, Hash)]
+pub struct Blob(pub Vec<u8>);
+
+impl From<Vec<u8>> for Blob {
```
ra0x3, line 36: 👌🏽 

Can we also add `TryFrom` here as well?

packages/fuel-indexer-types/src/lib.rs
```diff
@@ -29,11 +29,46 @@ pub type UInt16 = u128;
 pub type Timestamp = u64;
 pub type Charfield = String;
 pub type Boolean = bool;
-pub type Blob = Vec<u8>;
+
+#[derive(Deserialize, Serialize, Clone, Eq, PartialEq, Debug, Hash)]
```
ra0x3, line 33: 👌🏽 

This is fine but we'll want a `impl From<Blob> for Ve<u8>` as well (as well as a `TryFrom`)

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/src/lib.rs
```diff
@@ -139,7 +139,7 @@ mod fuel_indexer_test {
             id: result,
             result,
             gas_used,
-            blob: vec![1u8, 1, 1, 1, 1],
+            blob: Blob(vec![1u8, 1, 1, 1, 1]),
```
ra0x3, line 142: Should not have to wrap a `Blob` here. Using proper `impl` we can just `.into()` this

examples/hello-world/hello-indexer/src/lib.rs
```diff
@@ -80,7 +80,7 @@ mod hello_world_indexer {
 
                 // Here we show an example of an arbtrarily sized Blob type. These Blob types
                 // support data up to 10485760 bytes in length
-                visits: vec![1u8, 2, 3, 4, 5, 6, 7, 8],
+                visits: Blob(vec![1u8, 2, 3, 4, 5, 6, 7, 8]),
```
ra0x3, line 83: Should not have to wrap a `Blob` here. Using proper `impl` we can just `.into()` this

packages/fuel-indexer-schema/src/lib.rs
```diff
@@ -260,9 +262,9 @@ impl FtColumn {
                 None => String::from(NULL_VALUE),
             },
             FtColumn::Blob(value) => match value {
-                Some(val) => {
+                Some(fuel_indexer_types::Blob(val)) => {
```
ra0x3, line 265: Using `impl AsRef<[u8]> for Blob` prevents needing to make this change.

packages/fuel-indexer-schema/src/lib.rs
```diff
@@ -130,7 +130,9 @@ impl FtColumn {
                 );
                 FtColumn::Timestamp(Some(int8))
             }
-            ColumnType::Blob => FtColumn::Blob(Some(bytes[..size].to_vec())),
+            ColumnType::Blob => {
+                FtColumn::Blob(Some(fuel_indexer_types::Blob(bytes[..size].to_vec())))
```
ra0x3, line 134: Should not have to wrap a `Blob` here. Using proper `impl` we can just `.into()` this

ra0x3 (changes requested):
@rvmelkonian Thanks for picking this up so fast. 

In general we want to use type conversion `From`, `Into`, `TryFrom`, etc so that we don't have to manually wrap everything `in Type(...)`.

packages/fuel-indexer-schema/src/lib.rs
rvmelkonian: Ok let me try this!

examples/hello-world-native/hello-indexer-native/src/main.rs
rvmelkonian: Of course!

packages/fuel-indexer-types/src/lib.rs
rvmelkonian: I had initially just used `TryFrom` but then just went with `From` as I couldn't think of what kind of validation we would want here, and what could produce the error, as it's just 
```
impl From<Vec<u8>> for Blob {
    fn from(value: Vec<u8>) -> Self {
        Blob(value)
    }
}
```

ra0x3 (approved):
LGTM. Will defer to @deekerno

deekerno (approved):
LGTM as well.

Issue #643, opened 2023-03-13
Use proper names when referencing indexer abstractions


ra0x3:
- We need to go throughout the codebase and rename things for clarity
- These things include routes, migrations, table fields, variables, etc...everything

- Definitions
  - Indexer: an abstraction that takes Fuel-VM data and produces indices
  - Indexer service: A service that runs one or more indexers
  - Executors: async tasks run by indexers
  - Indices: data produced by an indexer

- Everywhere we're referencing "index" we need to most likely replace with "indexer"
  - If we don't do this, we're creating confusion, as an "index" is the resultant data produced by an "indexer"
- This will most likely require migrations for name changes to tables and such

deekerno:
@ra0x3 What do you think about adding a glossary to our book as part of the documentation cleanup for this milestone? I feel that it may be helpful to have a page to refer users (and perhaps ourselves) to when they have questions about the terminology.

ra0x3:
@deekerno I think that would be super helpful. Like a section “Glossary” in the SUMMARY. It would be small for now but I think that’s fine

deekerno:
@ra0x3 Word. I'm finishing up the first pass of the documentation cleanup, so I'll make a quick page to knock it out as well as update the terminology in the book itself. 👌🏽

PR #848, opened 2023-05-03
fix: don't send wasm bytes over service channel

ra0x3:
### Description
- PR fixes a bug with `forc index revert`
- Previously `revert` was sending the entire WASM blob over the channel (👎🏽 ) which might have been crashing the channel and closing it
- This PR fixes that by moving the "delete" portion of this to the use-case/handler, and loading the wasm bytes inside the service task (so that only the indexer identifier and namespace are sent over the channel)
- Also renames queries as a part of https://github.com/FuelLabs/fuel-indexer/issues/643
- ⚠️ We won't be able to test for sure that this works til we deploy to beta-3-indexer (after merge to master)
  - But we can still confirm the change locally

### Testing steps
- [ ] Start the service connected to beta-3 `cargo run --bin fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80`
- [ ] Deploy the block-explorer example `forc-index deploy --path examples/block-explorer/explorer-indexer --target-dir /Users/rashad/dev/repos/fuel-indexer`
- [ ] Make a change to explorer-indexer/lib.rs. Add a `Logger::info("TESTING REVERT");` to it
- [ ] Redeploy this updated indexer
- [ ] In your database you should see two WASM assets under that indexer ID, and only a single schema and manifest
- [ ] Now call revert `forc-index revert --path examples/block-explorer/explorer-indexer`
- [ ] In the console, you should see the `"TESTING REVERT"` disappear, as the older indexer is reloaded
- [ ] In the database, you should see you still have 1 schema/manifest asset, and now you have 1 wasm asset as well

### Changelog
- fix: don't send wasm bytes over service channel

### DIFF
--- a/packages/fuel-indexer-database/postgres/src/lib.rs
+++ b/packages/fuel-indexer-database/postgres/src/lib.rs
@@ -587,9 +587,7 @@ pub async fn latest_assets_for_indexer(
     index_id: &i64,
 ) -> sqlx::Result<IndexAssetBundle> {
     let wasm = latest_asset_for_indexer(conn, index_id, IndexAssetType::Wasm).await?;
-
     let schema = latest_asset_for_indexer(conn, index_id, IndexAssetType::Schema).await?;
-
     let manifest =
         latest_asset_for_indexer(conn, index_id, IndexAssetType::Manifest).await?;
 
--- a/packages/fuel-indexer-database/postgres/src/lib.rs
+++ b/packages/fuel-indexer-database/postgres/src/lib.rs
@@ -383,7 +383,7 @@ pub async fn columns_get_schema(
 }
 
 #[cfg_attr(feature = "metrics", metrics)]
-pub async fn index_is_registered(
+pub async fn indexer_is_registered(
     conn: &mut PoolConnection<Postgres>,
     namespace: &str,
     identifier: &str,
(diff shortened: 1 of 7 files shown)


packages/fuel-indexer-lib/src/utils.rs
```diff
@@ -69,8 +69,6 @@ pub struct IndexStopRequest {
 
 #[derive(Debug)]
 pub struct IndexRevertRequest {
```
deekerno, line 71: If we're renaming some types, we should probably rename this to `IndexerRevertRequest` and the others ones should also be renamed accordingly.

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -834,3 +832,31 @@ pub async fn get_nonce(
 
     Ok(Nonce { uid, expiry })
 }
+
+pub async fn remove_latest_assets_for_indexer(
```
ra0x3, line 836: This is unused, but want to leave it as this action (removing all latest assets for an indexer) can be valuable

deekerno (approved):
Works for me! Left a non-blocking comment.

PR #851, opened 2023-05-03
enhancement: add plugin command to kill indexer process

lostman:
Killing the indexer process:

```
forc-index kill
✅ Sucessfully killed process 85790 listening on port 29987
```

No process to kill:

```
forc-index kill
Error: ❌ No process is listening on port 29987
```

Same, but with a user-specified port:

```
forc-index kill --port 12345
Error: ❌ No process is listening on port 12345
```


ra0x3 (approved):
Works for me! 🤌🏽

PR #857, opened 2023-05-04
Update docs with vars

rvmelkonian:
Some users have needed to add these additional vars to their bash or zshrc. This PR adds those vars into our documentation.


deekerno (approved):
There will soon come a day in which Apple ships `clang` with support for the `wasm32-unknown-unknown` target, and when it arrives, there will be much rejoicing.

Alas, today is not that day.

PR #863, opened 2023-05-05
chore: create e2e tests for `/examples`

### LINKED ISSUE #856: Create Intergration tests for `/examples`
As new features are being added, breaking changes are being introduced and causing uncaught bugs in our `/examples`. We need tests and a new CI job to test the indexing functionality of our examples similar to how we test `/components` inside `fuel-indexer-tests`. Over there we use `web-api` inside `fixtures.rs` to ping the contracts on a local fuel-inder. A similar approach should be used here.

Integration Tests should be added, asserting that we can successfully index the data we expect in the `/examples`. 

We need these tests to be as least-brittle as possible, write them with this in mind, we don't want to add massive overhead here in terms of maintenance.

rvmelkonian:
Closes #856

### Description
This PR adds tests for our examples. 

The intention with this PR was to write some essential tests and provide the functionality necessary to do full e2e tests on our examples. There is room for more cases to be added. If anyone knows of any frequent bug that comes up in the examples, feel free to create an issue for adding the new test case. 

### Testing steps
- The new CI job has been added and should pass. 
- For local testing:
`cargo test --workspace --features examples,e2e,postgres -- examples --test-threads=1` 

### Changelog
• Creates new test file `examples.rs`
• Adds the tests:
- `test_release_build_hello_world_wasm_artifact`
- `test_deploy_indexer_hello_world`
- `test_release_build_block_explorer_wasm`
- `test_deploy_block_explorer`

If we're happy with this structure, I have additional test cases for indexing these examples, but so to reduce the size of this PR, will put that up after this.


rvmelkonian:
New test is failing on CI, but passes locally. Checking this at the moment; I'm expecting this to be something small and should not block review.

ra0x3:
- Gonna close this as I don't think anyone has bandwidth to pick it up
- But we will keep the PR around for reference when someone picks this up
- CC @deekerno @lostman

PR #872, opened 2023-05-08
enhancement: add graphql enum support

ra0x3:
cc @lostman 

### Description
- PR does a ton of things
  - Fixes type issues in the explorer example 
  - Refactors GraphQL schema codegen and table creation (it was spaghetti code)
  - Adds to concept of "virtual columns" on type IDs
    - Not all types need tables to be created 
    - Not only was this needed for enums but is a precursor for https://github.com/FuelLabs/fuel-indexer/issues/940
  - Cleans up how we're doing imports for the codgen and more important adds comments/explanations
  - Adds new types
    - We now support indexing enum types
      - Enum types have to be persisted into the DB as some type of `FtColumn` so I chose `String`
        - Users can convert this `String` back into its enum on their side if they wish 
  - Adds tests

### Testing steps
- [ ] Start a local node `cargo run -p fuel-node --bin fuel-node`
- [ ] Start a local web api `cargo run -p web-api --bin web-api`
- [ ] Start the service using the test indexer 
```bash
cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml --run-migrations
```
- [ ] Trigger the enum event `curl -X POST http://localhost:8000/enum`
- [ ] Open the playground `http://localhost:29987/api/playground/fuel_indexer_test/index1`
- [ ] Query for the indexed enum 

```graphql
query { complexenumentity { id one }}
```
Should return the object with its enum as a String

```json
{
  "data": [
    {
      "id": 1,
      "one": "EnumEntity::One"
    }
  ]
}
```

### Changelog
- Add GraphQL enum support
- Refactor schema and table codegen


packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -39,32 +28,34 @@ pub struct SchemaBuilder {
...
             version: version.to_string(),
-            primitives,
+            parsed_schema,
+            is_native,
             ..Default::default()
         })
     }
 
+    /// Generate table SQL for each object in the given schema.
     pub fn build(mut self, schema: &str) -> IndexerSchemaDbResult<Self> {
```
deekerno, line 59: Here's a thought:

Is there an instance where we'd want to create a `SchemaBuilder` instance in `SchemaManager::new_schema()` and _not_ immediately call `.build()` on that instance? If not, perhaps we can combine the `new()` and `build()` since all the necessary parts for `build()` are covered in the parameters for `new().

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -156,25 +154,31 @@ impl SchemaBuilder {
         Ok(schema)
     }
 
+    /// Derive the ColumnType and indicator of whether or not this field is a `ColumnType::ForeighKey`.
```
deekerno, line 157: Consider:
> Return a field's ColumnType and whether it is nullable.

packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -67,232 +82,314 @@ fn process_field(
...
-    object_name: String,
+/// Type of special fields in GraphQL schema.
+enum FieldType {
+    ForeignKey,
+    Enum,
+}
+
+/// Process an object's 'special' field and return a group of tokens.
+///
+/// Specicial fields are limited to foreign key and enum fields.
```
deekerno, line 106: s/Specicial/Special

packages/fuel-indexer-schema/src/parser.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// All unique names of enums in the schema.
+    pub enum_names: HashSet<String>,
+
+    /// All unique names of types for which tables should _not_ be created.
+    pub non_indexable_type_names: HashSet<String>,
+
+    /// All unique names of types that have already been parsed.
+    pub parsed_type_names: HashSet<String>,
+
+    /// A mapping of fully qualitified field names to their field types.
```
deekerno, line 33: s/qualitified/qualified

deekerno (changes requested):
Looks good on my first pass. 👌🏽 Left a couple comments around efficiency and documentation.

packages/fuel-indexer-schema/src/db/tables.rs
ra0x3: @deekerno I would like to put a pin in that and save it for the next round of refactor (I think tables/schema still needs one more round of cleanup then we should be in a good place).

deekerno (approved):

PR #875, opened 2023-05-09
fix: set platform: linux/x86_64 for fuel-indexer

lostman:
otherwise, docker-compose up fails on m1 mac:

```
no matching manifest for linux/arm64/v8 in the manifest list entries
```


rvmelkonian:
Can you add a way for me to test this locally @lostman ? If this one liner is literally the fix I will be so happy.

rvmelkonian (approved):

PR #884, opened 2023-05-09
Codeowners

Braqzen:
Added codeowners file which tags the entire indexer team for review meaning that we do not have to manually add reviewers as the entire team will be pinged / added.

Optional I guess: could close the PR if undesired.


ra0x3:
@deekerno @lostman 
- Will defer to you both here. 
- 2 out of 3 votes wins
- If you want to have this, we can merge it

ra0x3 (approved):

PR #901, opened 2023-05-12
add created_at field to index_registry

lostman:
### Description

Add the `created_at` column to the `index_registry` table.

### Testing steps

1. Run `fuel-indexer run --run-migrations --manifest PATH`.
2. Connect to `postgres` and run `select * from index_registry;`. `created_at` column should be present and initialized to epoch for any indexer in the table during migration or time of its creation for any indexer added after the migration.

### Changelog

* Add migrations for `created_at` column.
* Update Rust queries, make `register_index` add a timestamp.
* Add `Serializable` instance `#[serde(with = "ts_microseconds")]` (PostgreSQL represents a point in time with microsecond precision).

### DIFF
--- a/packages/fuel-indexer-database/postgres/src/lib.rs
+++ b/packages/fuel-indexer-database/postgres/src/lib.rs
@@ -402,7 +402,7 @@ pub async fn indexer_is_registered(
     {
         Some(row) => {
             let created_at: DateTime<Utc> = {
-                let created_at: NaiveDateTime = row.try_get("created_at")?;
+                let created_at: NaiveDateTime = row.get(4);
                 DateTime::<Utc>::from_utc(created_at, Utc)
             };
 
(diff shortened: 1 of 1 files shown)


ra0x3 (dismissed):
LGTM ✅  will defer to @deekerno

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -414,32 +424,39 @@ pub async fn register_indexer(
...
     .await?;
 
     let id: i64 = row.get(0);
     let namespace: String = row.get(1);
     let identifier: String = row.get(2);
     let pubkey = row.get(3);
+    let created_at: DateTime<Utc> = {
+        let created_at: NaiveDateTime = row.try_get("created_at")?;
+        DateTime::<Utc>::from_utc(created_at, Utc)
+    };
```
deekerno, line 452: Are we able to use `row.get(4)` here to remain consistent with your implementation on lines 476-479?

deekerno (changes requested):
Also looks good to me, just left one comment about consistency.

deekerno (approved):

PR #902, opened 2023-05-12
fix: remove query root + refactor graphql schema

ra0x3:

### Description

- Removes the concept of a query root from our GraphQL schema
- Fixes a few bugs in our GraphQL usage
- Updates tests accordingly
- Adds a migration to remove the query root from `graph_registry_graph_root`
- PR also includes a simple hack needed by the GraphQL web functionality (we will remove this eventually)

### Testing steps

- [ ] CI should pass
- [ ] Create an deploy a new indexer project
   - But don't use published crates as a dependency (it will fail)
   - Point your indexer dependencies to this PR branch
- [ ] Examples should work

### Changelog

- fix: remove query root + refactor graphql schema


packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -388,15 +355,14 @@ pub(crate) fn process_graphql_schema(
         #version_tokens
     };
 
-    let query_root = get_query_root(&schema_types, &ast);
-
     let mut processed: HashSet<String> = HashSet::new();
     let schema_types_map: HashMap<String, String> =
-        build_schema_fields_and_types_map(&ast);
+        build_schema_fields_and_types_map(&ast).unwrap_or_else(|e| {
+            panic!("Failed to build GraphQL shcema field and types: {e}")
```
deekerno, line 361: s/shcema/schema

packages/fuel-indexer-graphql/src/graphql.rs
```diff
@@ -532,37 +527,25 @@ impl<'a> GraphqlQueryBuilder<'a> {
...
-                    &selection_set.node,
-                )?;
-                selections.resolve_fragments(
-                    self.schema,
-                    &self.schema.query,
-                    fragments,
-                )?;
+                let mut selections =
+                    Selections::new(self.schema, "QueryRoot", &selection_set.node)?;
+                selections.resolve_fragments(self.schema, "QueryRoot", fragments)?;
```
deekerno, line 544: Could we use a `const` for these "QueryRoot" strings and use it here instead?

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -427,99 +392,115 @@ impl Schema {
...
+    // defined in a `TypeSystemDefinition::Schema`)
+    pub fn register_queryroot_fields(&mut self) {
+        self.fields.insert(
+            "QueryRoot".to_string(),
+            self.fields
+                .keys()
+                .map(|k| (k.to_lowercase(), k.clone()))
+                .collect::<HashMap<String, String>>(),
+        );
+    }
```
deekerno, line 416: Just so I understand: we'll need to adjust the query functionality to not need the `QueryRoot` object to be able to start through the parsing, right?

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -400,21 +365,21 @@ impl Schema {
...
+        };
 
-    pub fn check_type(&self, type_name: &str) -> bool {
-        self.types.contains(type_name)
+        schema.register_queryroot_fields();
+
+        Ok(schema)
     }
 
+    /// Ensure the given type is included in this `Schema`'s types
```
deekerno, line 382: I think this comment needs to be above `check_type()` as this function returns the type for a field.

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -377,20 +350,12 @@ impl Schema {
...
-        for tid in typeids {
-            types.insert(tid.graphql_name.clone());
+        for tyid in &typeids {
+            types.insert(tyid.graphql_name.clone());
 
-            let columns = queries::list_column_by_id(&mut conn, tid.id).await?;
+            let columns = queries::list_column_by_id(&mut conn, tyid.id).await?;
             fields.insert(
-                tid.graphql_name,
+                tyid.graphql_name.clone(),
```
deekerno, line 358: Is a clone necessary here?

deekerno (changes requested):
Looks pretty good to me on the first pass. Left a couple comments.

(also the PR is still in draft mode 😂)

packages/fuel-indexer-schema/src/db/tables.rs
ra0x3: @deekerno  
> "...to be able to start through the parsing"
- I'm not totally sure what that means 🤔 . This hack is needed because the GraphQL query implementation is getting its object/type info from the old way that query root was structured.

```graphql
type QueryRoot {
  foo: Foo
  bar: Bar
  zoo: Zoo
}
```
- It _should_ be a pretty straightforward fix when the time comes to remove  the hack.
- We already have all of that ☝🏽 info on `Schema.fields` anyway, so we don't need to use a `QueryRoot`

deekerno (approved):

PR #943, opened 2023-05-22
enhancement: add support for unindexable types

ra0x3:

### Description

- PR adds support for a `@norelation` directive
  - Applying this directive to any field on an object renders the entire object as `NoRelation`, which means the object is stored as `Json`
    - I chose `Json` over `Blob` because I think it might be easier to extract the item into an object
      - This will definitely require a second pass later
- PR also moves some code around (moves `serialize` and `deserialize` from fuel-indexer-schema to fuel-indexer-lib)
- 

### Testing steps

- [ ] Test was added so CI should pass
- [ ] Build assets `bash scripts/utils/build_test_wasm_module.bash`
- [ ] Start a local fuel node `cargo run -p fuel-node --bin fuel-node`
- [ ] Start the web API `cargo run -p web-api --bin web-api`
- [ ] Start the service with the test indexer 

```bash
cargo run --bin fuel-indexer -- run --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml --run-migrations
```
- [ ] Trigger some data `curl -X POST http://localhost:8000/block`
- [ ] Open your playground `http://localhost:29987/api/playground/fuel_indexer_test/index1`
- [ ] Make the following query:

```graphql
query {
  usesnotableentity {
    id
    name
    no_table
  }
}
```
- [ ] Should see your type come back as JSON
```json
{
  "data": [
    {
      "id": 1,
      "name": "hello world",
      "no_table": {
        "name": "norelation",
        "size": 1
      }
    }
  ]
}
```
- Note that the item `NoRelationEntity` already comes back as an object straight from the DB, so no further deserialization is needed

### Changelog

- enhancement: add support for a `@norelation` directive that does not create SQL tables for entities


packages/fuel-indexer-schema/src/lib.rs
```diff
@@ -78,176 +74,10 @@ pub enum FtColumn {
     UInt16(Option<UInt16>),
     UInt4(Option<UInt4>),
     UInt8(Option<UInt8>),
+    NoRelation(Option<NoRelation>),
 }
 
 impl FtColumn {
-    pub fn new(ty: ColumnType, size: usize, bytes: &[u8]) -> FtColumn {
```
ra0x3, line 84: @deekerno If you have this commit up, I can cherry-pick it from you (I think you might get a conflict regardless what we do 🤔)

packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -295,8 +334,23 @@ fn process_type_def(
...
+
+                    impl From<Json> for #strct {
+                        fn from(value: Json) -> Self {
+                            let s: #strct = serde_json::from_str(&value.0).expect("Serde error.");
+                            s
+                        }
+                    }
                 })
             } else {
+                // TOOD: derive Default here
```
deekerno, line 353: Is there a reason why we're not adding `Default` to the derived traits right now?

deekerno (commented):
Functionality looks good; the latest changes from master need to be integrated as the playground shows an error and the docs/schema don't load. But I'm sure they'll work after updating.

Left a couple of comments about readability.

packages/fuel-indexer-macros/src/schema.rs
ra0x3: https://github.com/FuelLabs/fuels-rs/pull/977

deekerno (approved):

PR #946, opened 2023-05-23
enhancement: update fuel explorer

ra0x3:

### Description

Please include a short derscription of what this PR does.

### Testing steps

### Changelog

### DIFF
--- a/packages/fuel-indexer-types/src/scalar.rs
+++ b/packages/fuel-indexer-types/src/scalar.rs
@@ -1,0 +1,121 @@
+use bytes::Bytes;
+pub use fuel_types::{
+    Address, AssetId, Bytes32, Bytes4, Bytes64, Bytes8, ContractId, MessageId, Salt, Word,
+};
+pub use fuels::{
+    core::try_from_bytes,
+    types::{
+        bech32::{Bech32Address, Bech32ContractId},
+        Bits256, Identity, SizedAsciiString,
+    },
+};
+use serde::{Deserialize, Serialize};
+use tai64::Tai64;
+
+/// Scalar for object IDs.
+pub type ID = u64;
+
+/// Scalar for 4-byte signed integers.
+pub type Int4 = i32;
+
+/// Scalar for 8-byte signed integers.
+pub type Int8 = i64;
+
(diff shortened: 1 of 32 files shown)


packages/fuel-indexer/src/executor.rs
```diff
@@ -64,12 +65,19 @@ impl ExecutorSource {
     }
 }
 
+/// Run the executor task until the kill switch is flipped, or until some other
+/// stop criteria is met.
+///
+/// In general the logic in this function isn't very idiomatic, but that's because
+/// types in `fuel_core_client` don't compile to WASM.
```
deekerno, line 72: We can probably change these two lines to be a regular comment instead of a doc comment as a user won't see the function logic in the crate documentation without clicking `Source`.

packages/fuel-indexer-types/src/scalar.rs
```diff
@@ -0,0 +1,124 @@
...
+pub use fuel_types::{
+    Address, AssetId, Bytes32, Bytes4, Bytes64, Bytes8, ContractId, MessageId, Salt, Word,
+};
+pub use fuels::{
+    core::try_from_bytes,
+    types::{
+        bech32::{Bech32Address, Bech32ContractId},
+        Bits256, Identity, SizedAsciiString,
+    },
+};
```
deekerno, line 11: I'm not sure if these should be re-exported under `fuel-indexer-types::scalar` as I'd imagine that the scalar types would be used for GraphQL-specific things. Perhaps we could re-export the ones that don't have a match in `fuel-indexer-schema/src/base.graphql` under a different name?

In any case, I don't feel strongly about this; just a suggestion.

deekerno (commented):
Did a first pass and everything looks pretty good. Left some comments; I'll approve once they're addressed/resolved.

packages/fuel-indexer-types/src/scalar.rs
deekerno: Yeah, the `fuels` types were what prompted the suggestion. It can wait for a future PR.

deekerno (approved):

PR #953, opened 2023-05-26
enhancement: re-export indexer `fuel-indexer-*` dependencies under `fuel-indexer-utils`

### LINKED ISSUE #867: Move all user-facing indexer components into `fuel-indexer-utils`
- As of today users have to import [a few external crates](https://github.com/FuelLabs/fuel-indexer/blob/master/plugins/forc-index/src/utils/defaults.rs#L51) when building indexers
- Further, we have [a set of utility functions that users can use to improve indexer devx](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer-lib/src/utils.rs#L219)
- We should move all of this (as much as possible into `fuel-indexer-utils`
  - Crate already reserved https://crates.io/crates/fuel-indexer-utils
- `fuel-indexer-{macros, schema, plugin}`, should be re-exported if possible
  - Debating on whether or not we should just add `fuels` and `fuel-tx` as well
  - Open to having a discussion on this 
- The utility functions should be exported accordingly
- New exports should export a `prelude` mod where we export common things (as per the convention)

Additional context
- We should rust-doc _this entire crate_ before merging
  - The entire purpose of it is to provide a single interface for users to reference when building indexers

deekerno:
Closes #867.

### Description

Re-exports `fuel-indexer-{macros, plugin, schema}` under `fuel-indexer-utils`

We will be able to change the `forc-index` defaults when we publish a new version of the `fuel-indexer-utils` crate.

### Testing steps

CI should pass.

### Changelog

[Re-export indexer fuel-indexer-* dependencies under fuel-indexer-utils](https://github.com/FuelLabs/fuel-indexer/commit/285c9ee4b3f4228d302fc12a43aad64c1789755b)

### DIFF
--- a/examples/fuel-explorer/fuel-explorer/src/lib.rs
+++ b/examples/fuel-explorer/fuel-explorer/src/lib.rs
@@ -1,5 +1,8 @@
 extern crate alloc;
-use fuel_indexer_macros::indexer;
+use fuel_indexer_utils::macros::indexer;
+use fuel_indexer_utils::prelude::*;
+#[allow(unused_imports)]
+use fuel_indexer_utils::utilities::*;
 
 pub enum ConsensusLabel {
     Unknown,
--- a/examples/fuel-explorer/fuel-explorer/src/lib.rs
+++ b/examples/fuel-explorer/fuel-explorer/src/lib.rs
@@ -1,8 +1,5 @@
 extern crate alloc;
-use fuel_indexer_utils::macros::indexer;
 use fuel_indexer_utils::prelude::*;
-#[allow(unused_imports)]
-use fuel_indexer_utils::utilities::*;
 
 pub enum ConsensusLabel {
     Unknown,
--- a/examples/hello-world-native/hello-indexer-native/Cargo.toml
+++ b/examples/hello-world-native/hello-indexer-native/Cargo.toml
@@ -7,8 +7,6 @@ publish = false
 [dependencies]
 async-trait = { version = "0.1" }
 fuel-indexer = { workspace = true }
-fuel-indexer-macros = { workspace = true, default-features = false }
-fuel-indexer-plugin = { workspace = true, features = ["native-execution"] }
-fuel-indexer-schema = { workspace = true, default-features = false }
+fuel-indexer-utils = { workspace = true, features = ["native-execution"] }
 fuels = { workspace = true }
 serde = { workspace = true }
--- a/examples/hello-world-native/hello-indexer-native/src/main.rs
+++ b/examples/hello-world-native/hello-indexer-native/src/main.rs
@@ -18,10 +18,7 @@
 //! cargo run -p hello-world-data --bin hello-world-data
 //! ```
 extern crate alloc;
-use fuel_indexer::prelude::*;
-use fuel_indexer_utils::macros::indexer;
 use fuel_indexer_utils::prelude::*;
-use fuel_indexer_utils::utilities::*;
 
 #[indexer(
     manifest = "examples/hello-world-native/hello-indexer-native/hello_indexer_native.manifest.yaml"
--- a/examples/hello-world-native/hello-indexer-native/src/main.rs
+++ b/examples/hello-world-native/hello-indexer-native/src/main.rs
@@ -19,8 +19,9 @@
 //! ```
 extern crate alloc;
 use fuel_indexer::prelude::*;
-use fuel_indexer_macros::indexer;
-use fuel_indexer_plugin::prelude::*;
+use fuel_indexer_utils::macros::indexer;
+use fuel_indexer_utils::prelude::*;
+use fuel_indexer_utils::utilities::*;
 
 #[indexer(
     manifest = "examples/hello-world-native/hello-indexer-native/hello_indexer_native.manifest.yaml"
--- a/examples/hello-world/hello-indexer/src/lib.rs
+++ b/examples/hello-world/hello-indexer/src/lib.rs
@@ -27,9 +27,7 @@
 //! ```
 
 extern crate alloc;
-use fuel_indexer_utils::macros::indexer;
 use fuel_indexer_utils::prelude::*;
-use fuel_indexer_utils::utilities::*;
 
 #[indexer(manifest = "examples/hello-world/hello-indexer/hello_indexer.manifest.yaml")]
 mod hello_world_indexer {
--- a/examples/hello-world/hello-indexer/src/lib.rs
+++ b/examples/hello-world/hello-indexer/src/lib.rs
@@ -27,8 +27,9 @@
 //! ```
 
 extern crate alloc;
-use fuel_indexer_macros::indexer;
-use fuel_indexer_plugin::prelude::*;
+use fuel_indexer_utils::macros::indexer;
+use fuel_indexer_utils::prelude::*;
+use fuel_indexer_utils::utilities::*;
 
 #[indexer(manifest = "examples/hello-world/hello-indexer/hello_indexer.manifest.yaml")]
 mod hello_world_indexer {
--- a/packages/fuel-indexer-lib/Cargo.toml
+++ b/packages/fuel-indexer-lib/Cargo.toml
@@ -13,7 +13,6 @@ description = "Fuel Indexer Library"
 anyhow = "1.0"
 bincode = { workspace = true }
 clap = { features = ["cargo", "derive", "env"], workspace = true }
-fuel-indexer-types = { workspace = true }
 http = { version = "0.2", default-features = false }
 serde = { workspace = true }
 serde_json = { workspace = true }
--- a/packages/fuel-indexer-macros/src/native.rs
+++ b/packages/fuel-indexer-macros/src/native.rs
@@ -39,13 +39,11 @@ fn native_prelude() -> proc_macro2::TokenStream {
         // all dependencies explicity (preferably through a single crate).
         use fuel_indexer_utils::{
             plugin::{
+                bincode, deserialize,
+                serde_json, serialize,
                 native::*,
                 serde::{Deserialize, Serialize},
                 types::*,
-            },
-            prelude::{
-                bincode, deserialize,
-                serde_json, serialize,
             },
         };
         use fuels::{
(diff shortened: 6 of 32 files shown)


examples/fuel-explorer/fuel-explorer/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 extern crate alloc;
-use fuel_indexer_macros::indexer;
+use fuel_indexer_utils::macros::indexer;
+use fuel_indexer_utils::prelude::*;
+#[allow(unused_imports)]
+use fuel_indexer_utils::utilities::*;
```
ra0x3, line 5: 👌🏽

examples/hello-world-native/hello-indexer-native/Cargo.toml
```diff
@@ -7,8 +7,6 @@ publish = false
 [dependencies]
 async-trait = { version = "0.1" }
 fuel-indexer = { workspace = true }
-fuel-indexer-macros = { workspace = true, default-features = false }
-fuel-indexer-plugin = { workspace = true, features = ["native-execution"] }
-fuel-indexer-schema = { workspace = true, default-features = false }
+fuel-indexer-utils = { workspace = true, features = ["native-execution"] }
```
ra0x3, line 10: Much more simple IMO 👌🏽

examples/hello-world/hello-indexer/src/lib.rs
```diff
@@ -27,8 +27,9 @@
 //! ```
 
 extern crate alloc;
-use fuel_indexer_macros::indexer;
-use fuel_indexer_plugin::prelude::*;
+use fuel_indexer_utils::macros::indexer;
+use fuel_indexer_utils::prelude::*;
+use fuel_indexer_utils::utilities::*;
```
ra0x3, line 32: Could we re-export `fuel_indexer_utils::utilities::*` via `fuel_indexer_utils::prelude::*` here? And maybe `fuel_indexer_utils::macros::indexer` as well?

So ideally my import would be:

```rust
use fuel_indexer_utils::prelude::*
```

And that ☝🏽 would give me all utils, `#[indexer]` and whatever else is exported in `fuel_indexer_utils::prelude::*`

packages/fuel-indexer-lib/Cargo.toml
```diff
@@ -13,7 +13,6 @@ description = "Fuel Indexer Library"
 anyhow = "1.0"
 bincode = { workspace = true }
 clap = { features = ["cargo", "derive", "env"], workspace = true }
-fuel-indexer-types = { workspace = true }
```
ra0x3, line 16: Nice 👌🏽

packages/fuel-indexer-tests/components/indices/fuel-indexer-test/src/lib.rs
```diff
@@ -1,7 +1,8 @@
 extern crate alloc;
 
-use fuel_indexer_macros::indexer;
-use fuel_indexer_plugin::utils::*;
+use fuel_indexer_utils::macros::indexer;
```
ra0x3, line 3: Same comment here regarding ideally a single import

packages/fuel-indexer-plugin/src/lib.rs
```diff
@@ -16,13 +16,7 @@ pub mod types {
...
 
 pub mod utils {
-    pub use fuel_indexer_lib::utils::{
-        indexer_utils::{
-            bytes32_from_inputs, first32_bytes_to_bytes32, first8_bytes_to_u64,
-            trim_sized_ascii_string, u64_id, u64_id_from_inputs,
-        },
-        sha256_digest,
-    };
+    pub use fuel_indexer_lib::utils::sha256_digest;
```
ra0x3, line 19: 👌🏽

packages/fuel-indexer-utils/src/utilities.rs
```diff
@@ -0,0 +1,37 @@
+use crate::prelude::sha256_digest;
+use crate::prelude::{Bytes32, SizedAsciiString};
+
+pub fn u64_id(d: &[u8; 8]) -> u64 {
```
ra0x3, line 4: I realize we don't have much to put in `fuel-indexer-utils` yet, but could we rustdoc what we have? I think the SDK's rustdocs are a good example of what I'm referring to.

Ideally, users would only ever look up docs for this `fuel-indexer-utils` crate, so of all our crates, this one should be the easiest to follow (for example on crates.io)

packages/fuel-indexer-utils/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+fuel-indexer-macros = { default-features = false, workspace = true }
+fuel-indexer-plugin = { workspace = true }
+fuel-indexer-schema = { default-features = false, workspace = true }
+fuels = { default-features = false, workspace = true }
+serde = { workspace = true, features = ["derive"] }
+
+[features]
+native-execution = ["fuel-indexer-plugin/native-execution", "fuels/default"]
```
ra0x3, line 20: 👌🏽

packages/fuel-indexer-utils/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub use fuel_indexer_macros as macros;
+pub use fuel_indexer_plugin::*;
+pub use fuel_indexer_schema;
+pub use fuels;
```
ra0x3, line 4: If we're going to allow `fuels` with this level of scope (public), thoughts on removing it as a dependency from indexers? Cause with this ☝🏽 I can just `use fuel_indexer_utils::fuels::` as opposed to having to manage another external dependency

packages/fuel-indexer-utils/src/utilities.rs
```diff
@@ -0,0 +1,37 @@
+use crate::prelude::sha256_digest;
+use crate::prelude::{Bytes32, SizedAsciiString};
```
ra0x3, line 2: Could we import these from their actual module instead of the prelude? 

```rust
use fuel_indexer_plugin::types::{Bytes32, SizedAsciiString};
```

Just trying to keep the references as explicit as possible where possible

packages/fuel-indexer-macros/src/native.rs
```diff
@@ -31,20 +31,26 @@ pub fn handler_block_native(
...
-            core::abi_decoder::ABIDecoder,
-            macros::{Parameterize, Tokenizable},
-            types::{StringToken, traits::{Tokenizable, Parameterize}},
-        };
-
         type B256 = [u8; 32];
 
         static mut db: Option<Arc<Mutex<Database>>> = None;
 
+        use fuel_indexer_utils::{
```
ra0x3, line 38: Could we also keep the FYI comment that was here?

examples/hello-world-native/hello-indexer-native/src/main.rs
```diff
@@ -19,8 +19,9 @@
 //! ```
 extern crate alloc;
 use fuel_indexer::prelude::*;
-use fuel_indexer_macros::indexer;
-use fuel_indexer_plugin::prelude::*;
+use fuel_indexer_utils::macros::indexer;
```
ra0x3, line 22: Same comment here regarding a single import

ra0x3 (changes requested):
@deekerno 

Left some feedback and some questions.

This is a pretty non-trivial change so wanna make sure we're on the same page 👌🏽

examples/hello-world/hello-indexer/src/lib.rs
deekerno: Yep, I can do that.

packages/fuel-indexer-utils/src/utilities.rs
deekerno: Oops, forgot about that bit; will do.

packages/fuel-indexer-utils/src/lib.rs
deekerno: I was having quite a bit of trouble before with being able to remove it, but I think I may have fixed the issue with my last push, so I'll try this again. We'll be able to get rid of version mismatches that way. 👌🏽

packages/fuel-indexer-utils/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+pub use fuel_indexer_macros as macros;
+pub use fuel_indexer_plugin::*;
```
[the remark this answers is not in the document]
deekerno: Yep, I had it before, but it looks like I blew it away somehow; will add it back.

packages/fuel-indexer-utils/src/lib.rs
deekerno: It appears that `fuels` has to stay as part of the `Cargo.toml` as the SDK codegen itself depends on some attributes that I can't seem to include in this utils crate. 👎🏽

ra0x3: Ok in that case we can remove it here

packages/fuel-indexer-macros/src/native.rs
```diff
@@ -31,20 +31,30 @@ pub fn handler_block_native(
...
 
         // TODO: Eventually prevent these types of implicity imports and have users import
         // all dependencies explicity (preferably through a single crate).
+        use fuel_indexer_utils::{
+            plugin::{
+                native::*,
+                serde::{Deserialize, Serialize},
+                types::*,
+            },
+            prelude::{
```
ra0x3, line 46: @deekerno Any reason we're not importing through `plugin` here? Ideally, the `prelude` is only used by users. (Importing through the plugin makes things a bit more explicit for devs)

packages/fuel-indexer-macros/src/wasm.rs
```diff
@@ -38,25 +38,30 @@ pub fn handler_block_wasm(
...
 
         // TODO: Eventually prevent these types of implicity imports and have users import
         // all dependencies explicity (preferably through a single crate).
+        use fuel_indexer_utils::{
+            plugin::{
+                serde::{Deserialize, Serialize},
+                types::*,
+                wasm::{Digest, Entity, FromHex, Logger, Sha256},
+            },
+            prelude::{
```
ra0x3, line 53: Same here regarding always going through the `plugin`

ra0x3 (approved):

PR #956, opened 2023-05-26
enhancement: continue flushing out explorer

ra0x3:

### Description

- Reorganizes type imports hopefully one last time
- Common types (e.g., `BlockData`) and field traits (e.g., `ReceiptsRoot`) are exported directly in the plugins `types`
  - We don't allow schema objects to have these ☝🏽 names
  - _Everything_ copied from `fuel_client_schema` is now in the `fuel_indexer_types::fuels` module
    - When we refactor/replace these types, we only have to update one file
    - Namespacing these types under `fuel::` allows users to use these same common names in their GraphQL schema

### Testing steps

- [ ] Start the service connected to beta-3 using this explorer indexer
```bash
cargo run --bin fuel-indexer -- run --manifest examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --verbose
```
- [ ] Things should index properly 

### Changelog

- enhancement: continue flushing out explorer


deekerno (approved):
🆗

PR #962, opened 2023-06-01
fuel-indexer run --replace-indexer

lostman:

### Description

Add `--replace-indexer` flag which changes how `forc index deploy` works: deployed indexer completely replaces the existing indexer(s) under the same UID. All previous indexer assets, graphql entries, and indexer tables are removed.

### Testing steps

I have a simple contract/script/indexer which may be helpful: https://github.com/lostman/fuel-counter

1. Run `cargo run --features fuel-core-lib -p fuel-indexer -- run --local-fuel-node --replace-indexer`
2. `git clone https://github.com/lostman/fuel-counter`
3. Create an indexer with colliding UID `examples.counter_indexer` that we will replace later:
```
mkdir dummy && cd dummy && forc index new --namespace examples counter_indexer && cd counter_indexer && forc deploy
```
4 Deploy the `counter` contract, an indexer, and run a script to trigger an event: `cd fuel-counter && make`
5. Confirm that the dummy indexer was replaced and that an event was indexed:
```
postgres=> select * from examples_counter_indexer.increntity;
         id          | value |                                 object
---------------------+-------+------------------------------------------------------------------------
 7003491485461406310 |   123 | \x02000000000000000d0000000166626662666631611e000000017b00000000000000
 ```

### Changelog

* Add `--replace-indexer` flag
* Change `remove_indexer` to remove the indexer, grapgl related entries, and schema tables.
* Remove `forc index revert` and all mentions from the docs.

### DIFF
--- a/packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
+++ b/packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
@@ -1024,11 +1024,11 @@ async fn test_deploying_twice_returns_an_error() {
     assert!(result.is_err());
 
     match result.unwrap_err() {
-        fuel_indexer::IndexerError::OtherError(msg) => {
+        fuel_indexer::IndexerError::Unknown(msg) => {
             assert_eq!(&msg, "Indexer(fuel_indexer_test.index1) already exists")
         }
         err => {
-            panic!("Expected OtherError but got: {}", err)
+            panic!("Expected Unknown but got: {}", err)
         }
     }
 }

### CODE REVIEW

**Comment by ra0x3** [comment]
> I’m thinking: perhaps instead of `--replace-indexer` (or `--auto-replace-indexer` as suggested in the issue), this flag should be called `--deploy-behavior=version|replace`?

@lostman 
- Only thing about that is there is no more versioning (after we merge this)
- Either your indexer (and its graph registry components) gets replaced with the new one, or an error is returned 
  - Which is why I suggested that it just be a `--cli-flag`

File: packages/fuel-indexer-database/postgres/src/lib.rs [hunk 1]
```diff
@@ -757,40 +757,113 @@ pub async fn revert_transaction(
...
 }
 
+#[cfg_attr(feature = "metrics", metrics)]
+pub async fn remove_indexers(
+    conn: &mut PoolConnection<Postgres>,
+    namespace: &str,
+    identifier: &str,
+    limit: Option<usize>,
+) -> sqlx::Result<()> {
+    let limit = limit.map(|l| format!("LIMIT {l}")).unwrap_or_default();
```
ra0x3, line 813: Due to the foreign key constraints we're using, there cannot be an indexer with the same namespace and identifier, so I think we can remove this?

packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -757,40 +757,113 @@ pub async fn revert_transaction(
...
+                WHERE namespace = '{namespace}' AND identifier = '{identifier}' {limit})"
+        ),
+    )
+    .await?;
+
+    Ok(())
+}
+
+#[cfg_attr(feature = "metrics", metrics)]
+pub async fn remove_indexer(
```
ra0x3, line 859: - Any reason that `remove_indexers` is broken out? 
- `remove_indexers` implies that we'll remove multiple indexers, but as the code, reads, we're only ever removing a single one ?

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -261,6 +261,35 @@ pub(crate) async fn register_indexer_assets(
...
+                    queries::remove_indexers(&mut conn, &namespace, &identifier, None)
+                        .await
+                {
+                    error!("Failed to remove Indexer({namespace}.{identifier}): {e}");
+                    queries::revert_transaction(&mut conn).await?;
+                    return Err(e.into());
+                }
+
+                if let Err(e) =
+                    queries::remove_graph(&mut conn, &namespace, &identifier).await
```
ra0x3, line 284: - This gets to my "single transaction" comment above ☝🏽 
- I think `remove_graph` should be nested inside `remove_indexer`. 
  - There is never a case in which we want to call `remove_graph` without calling `remove_indexer
  - And if its nested, that means we can use a single transaction for the entire removal

packages/fuel-indexer-lib/src/config/mod.rs
```diff
@@ -380,6 +387,7 @@ impl Default for IndexerArgs {
             rate_limit: defaults::RATE_LIMIT_ENABLED,
             rate_limit_request_count: Some(defaults::RATE_LIMIT_REQUEST_COUNT),
             rate_limit_window_size: Some(defaults::RATE_LIMIT_WINDOW_SIZE),
+            replace_indexer: false,
```
ra0x3, line 390: Let's follow our pattern of using defaults here with `defaults::REPLACE_INDEXER` (or some such)

packages/fuel-indexer-lib/src/config/mod.rs
```diff
@@ -570,6 +580,8 @@ impl From<ApiServerArgs> for IndexerConfig {
                 request_count: args.rate_limit_request_count,
                 window_size: args.rate_limit_window_size,
             },
+            // TODO: is this the right thing to do here?
+            replace_indexer: false,
```
ra0x3, line 584: Since the API server doesn't run an indexer, yes we can just use a `false` here

ra0x3 (changes requested):
@lostman 
- Thanks for the work on this 
- Overall, looks pretty good
- This is a really non-trivial change, thus the thorough feedback
- Mainly we need to:
  - Remove the new indexer (we can just use fuel-indexer-test indexer)
  - Do all removal in a single transaction preferably 
  - If the operator has `replace-indexer = false` we need to return an error if the indexer already exists
    - [409 status seems appropriate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409)
 
- Additional functionality that isn't in this PR
  - If I have `replace-indexer = true` then you need to (1) remove the older indexer (2) register the new indexer being uploaded (similar to what `register_indexer_assets` does)
  - We need to remove _all_ references to `forc index revert` since this functionality no longer  is needed
     - Removal from the code, and docs

ra0x3 (changes requested):
@lostman 
- Thanks for the work on this 
- Overall, looks pretty good
- This is a really non-trivial change, thus the thorough feedback
- Mainly we need to:
  - Remove the new indexer (we can just use fuel-indexer-test indexer)
  - Do all removal in a single transaction preferably 
  - If the operator has `replace-indexer = false` we need to return an error if the indexer already exists
    - [409 status seems appropriate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409)
 
- Additional functionality that isn't in this PR
  - If I have `replace-indexer = true` then you need to (1) remove the older indexer (2) register the new indexer being uploaded (similar to what `register_indexer_assets` does)
  - We need to remove _all_ references to `forc index revert` since this functionality no longer  is needed
     - Removal from the code, and docs

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -254,6 +200,22 @@ pub(crate) async fn register_indexer_assets(
     if let Some(mut multipart) = multipart {
         queries::start_transaction(&mut conn).await?;
 
+        if (queries::get_indexer_id(&mut conn, &namespace, &identifier).await).is_ok() {
```
ra0x3, line 203: Nit: maybe we could make this a bit more legible?

```rust
let indexer_exists = queries::get_indexer_id(&mut conn, &namespace, &identifier).await;
if (indexer_exists) {
  // do the thing..
}
```

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -345,13 +347,16 @@ pub async fn authenticated_api_server_app_postgres(database_url: Option<&str>) -
     GraphQlApi::build(config, pool, tx).await.unwrap()
 }
 
-pub async fn indexer_service_postgres(database_url: Option<&str>) -> IndexerService {
+pub async fn indexer_service_postgres(
+    database_url: Option<&str>,
+    modify_config: Option<Box<dyn Fn(&mut IndexerConfig)>>,
```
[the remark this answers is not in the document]
ra0x3: @lostman That's fine. Unrelated (and you totally don't have to do this). But if you get a minute, we should replace what we were previously doing with this ☝🏽  (I know we would need to add this for the authenticated postgres app fixture).

packages/fuel-indexer/src/lib.rs
```diff
@@ -83,4 +83,6 @@ pub enum IndexerError {
     NativeExecutionRuntimeError,
     #[error("Tokio time error: {0:?}")]
     Elapsed(#[from] tokio::time::error::Elapsed),
+    #[error("{0}.")]
+    OtherError(String),
```
ra0x3, line 87: Nit
```suggestion
    Unknown(String),
```
(To be consistent with how we do thing elsewhere)

packages/fuel-indexer/src/service.rs
```diff
@@ -60,6 +55,32 @@ impl IndexerService {
         mut manifest: Manifest,
     ) -> IndexerResult<()> {
         let mut conn = self.pool.acquire().await?;
+
+        if (queries::get_indexer_id(&mut conn, &manifest.namespace, &manifest.identifier)
```
ra0x3, line 59: Same comment here with regard to readability 

```rust
```rust
let indexer_exists = queries::get_indexer_id(&mut conn, &namespace, &identifier).await;
if (indexer_exists) {
  // do the thing..
}
```

ra0x3 (commented):
@lostman 
- Left some feedback with regard to readability and small nits
- As I'm testing this out I'm seeing indexer errors after re-uploading the indexer with `--replace-indexer`. Did manual testing work for you?
- Repro steps
  - Start the service with `--replace-indexer`
```bash
cargo run --bin fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --replace-indexer
```
  - Deploy an indexer, then redeploy the same indexer
- Errors:

```text
2023-06-09T16:37:52.005338Z ERROR fuel_indexer::database: 127: TypeId(-1599046217856387010) not found in tables: {}.

Does the schema version in SchemaManager::new_schema match the schema version in Database::load_schema?

Do your WASM modules need to be rebuilt?
2023-06-09T16:37:52.005520Z  INFO fuel_indexer::ffi: 110: Processing a block. (>'.')>
2023-06-09T16:37:52.005722Z ERROR fuel_indexer::database: 127: TypeId(1406149282443221513) not found in tables: {}.

Does the schema version in SchemaManager::new_schema match the schema version in Database::load_schema?

Do your WASM modules need to be rebuilt?
2023-06-09T16:37:52.005863Z  INFO fuel_indexer::ffi: 110: Handling a transaction (>'.')>
2023-06-09T16:37:52.006000Z ERROR fuel_indexer::database: 127: TypeId(-944680147995759449) not found in tables: {}.

Does the schema version in SchemaManager::new_schema match the schema version in Database::load_schema?

Do your WASM modules need to be rebuilt?
2023-06-09T16:37:52.006269Z ERROR fuel_indexer::database: 127: TypeId(-1599046217856387010) not found in tables: {}.

Does the schema version in SchemaManager::new_schema match the schema version in Database::load_schema?

Do your WASM modules need to be rebuilt?
2023-06-09T16:37:52.006432Z  INFO fuel_indexer::ffi: 110: Processing a block. (>'.')>
2023-06-09T16:37:52.006604Z ERROR fuel_indexer::database: 127: TypeId(1406149282443221513) not found in tables: {}.

Does the schema version in SchemaManager::new_schema match the schema version in Database::load_schema?

Do your WASM modules need to be rebuilt?
```
- As I'm thinking about this, I believe this is due to the fact that the database reference being used by the indexer cannot see that the schema has changed
  - We would need some way of letting the `fuel_indexer::database::Database` object used by the indexer executor, know that it needs to reload the schema 

CC @deekerno 🤔

ra0x3:
@lostman 
- To clarify my earlier comment, I only see this error when:
  - Start the service with `--replace-indexer`
  - Deploy a default `forc index new` indexer
  - Add the following to your schema 
```graphql
type Foo {
  id: ID!
  name: Charfield!
}
```
   - Redeploy this indexer after making these changes
   - Now you get the error

I was originally thinking we're not loading the new DB schema in the executor, but that is not the case [since we just create a new executor on asset reload](https://github.com/FuelLabs/fuel-indexer/blob/master/packages/fuel-indexer/src/service.rs#L254).

PR #968, opened 2023-06-01
chore: reduce log spam

deekerno:
### Description

When debugging a separate issue with full blocks, it was quite difficult to parse through the output as certain macro and database logs were printed for each contract and transaction in the block. As such, I elected to place said logs at the `DEBUG` log level so that users could more easily find the logs from their indexer modules.

### Testing steps

CI should pass. You can see the debug logs by appending `RUST_LOG="debug"` to the beginning of `fuel-indexer run`.

### Changelog
- [Move 'unknown X; check Y' type logs to debug; move stash conn log to debug](https://github.com/FuelLabs/fuel-indexer/commit/2a6c7508897d124cb4894d55e8005f06f37d7bc9)
- [Do not print playground link when starting without a manifest](https://github.com/FuelLabs/fuel-indexer/commit/ca6710d9499215fd6d0015cd9cc56979fe38c6a6)

### DIFF
--- a/packages/fuel-indexer-lib/src/utils.rs
+++ b/packages/fuel-indexer-lib/src/utils.rs
@@ -192,7 +192,10 @@ pub async fn init_logging(config: &IndexerConfig) -> anyhow::Result<()> {
         .unwrap_or("info".to_string());
 
     if !config.verbose {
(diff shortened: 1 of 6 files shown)


ra0x3 (approved):
- Nice. 
- Was thinking this same thing needs to be done. 
- Way to be proactive!

PR #995, opened 2023-06-09
chore: implement union types in explorer

ra0x3:

### Description

- PR integrates `union` types into the explorer
- Renames everything related to the "no relation" directive to "virtual"
  - "no relation" was always a temporary name
  - Thankfully no one is using this directive yet so not gonna break anyone's things 
  - Also updates directives docs

### Testing steps

- [ ] CI should pass

### Changelog

- chore: implement union types in explorer


docs/src/reference-guide/components/graphql/directives.md
```diff
@@ -65,3 +66,22 @@ type Library {
...
+
+type Book {
+    id: ID!
+    title: Title!
+}
+```
+
+When SQL tables are generated for the entities above, a table will be created for `Book`, but no table will be created for `Title`. Rather, the `title` field on the `Book` object will exist on the `book` table as a `JSON` field.
+
+> Important: When using the `@virtual` directive with GraphQL `union` types, each member of the `union` type must either include _only_ types that are not virtual, or _only_ types that are virtual. We do not support mixing and matching virtual types with non-virtual types in unions.
```
deekerno, line 87: I think it would be helpful to include an example of when a user may want to mark a field with `@virtual`.

deekerno (approved):
Looks good to me, left one comment.

docs/src/reference-guide/components/graphql/directives.md
ra0x3: @deekerno Will add this to the to-do list

Issue #963, opened 2023-06-01
Add `.get_or_create` method to `Entity` traits


ra0x3:
- While implementing explorer types I'm noticing a common pattern I'm using _a lot_ is this `get_or_create` pattern
  - [Here is the version of this in Django](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create)
- Here is effectively what I'm constantly doing

```rust
// If the entity doesn't exist, create it
let id = 1;
let entity = Entity::load(id).unwrap_or_else(|| {
  let e = Entity{ id, name: "Rashad" };
  e.save();
  e
});
```

- It would be great if I could just use a `.get_or_create` method

```rust
// If the entity doesn't exist, create it
let id = 1;

// Since we only find by ID, we'd only use ID for the query 
let e = Entity::get_or_create(id=id, Entity{ id, name: "Rashad"});
// `Entity` was already saved to the DB in `get_or_create`
```
- This is definitely a pattern that users will use often

PR #1003, opened 2023-06-13
enhancement: implement instantiation methods for entity types

### LINKED ISSUE #951: Calculate `u64` value from the hash of a struct's fields
A user should be able to automatically create an ID for a struct based on the content of a struct.

deekerno:
Closes #951 and #963.

### Description
This PR adds a `::new()` method to entity types that automatically calculates an ID for the user so that they do not have to come up with their own way to create one. It also adds a `::get_or_create()` method that will load a record from the DB if present or create the struct as described above.

### Testing steps

The `hello-world` example has been changed to illustrate the new instantiation methods.
1. Ensure the test DB is cleared and that the `hello-world` WASM module has been freshly built.
2. `cargo run -p hello-world-node`
3. `cargo run --bin fuel-indexer -- run --run-migrations --manifest examples/hello-world/hello-indexer/hello_indexer.manifest.yaml --verbose`
4. `cargo run -p hello-world-data`
5. Check the output of the indexer; you should see that the original and `::new()` methods insert data into the DB, and that `::get_or_create()` loads a record from the database.

### Changelog

- Add a number of `HashSet`s for checking whether a field type implements the requisite traits
- Add a helper function for creating the necessary method implementations

### DIFF
--- a/packages/fuel-indexer-macros/src/helpers.rs
+++ b/packages/fuel-indexer-macros/src/helpers.rs
@@ -83,7 +83,7 @@ pub fn to_bytes_method_for_external_type(
 ) -> proc_macro2::TokenStream {
     match field_type_name.as_str() {
         "Identity" => quote! { .0 },
-        "Tai64Timestamp" => quote! { .0.to_be_bytes() },
+        "Tai64Timestamp" => quote! { .0.to_le_bytes() },
         _ => panic!(
             "No bytes conversion method available for external type {field_type_name}"
         ),
@@ -380,7 +380,7 @@ pub fn generate_struct_new_method_impl(
     object_name: String,
     struct_fields: proc_macro2::TokenStream,
 ) -> proc_macro2::TokenStream {
-    let new_method_impl = if !INTERNAL_INDEXER_ENTITIES.contains(object_name.as_str()) {
+    if !INTERNAL_INDEXER_ENTITIES.contains(object_name.as_str()) {
         quote! {
             impl #strct {
                 pub fn new(#parameters) -> Self {
@@ -388,7 +388,7 @@ pub fn generate_struct_new_method_impl(
 
                     let id_bytes = <[u8; 8]>::try_from(&raw_bytes[..8]).expect("Could not calculate bytes for ID from struct fields");
 
-                    let id = u64::from_be_bytes(id_bytes);
+                    let id = u64::from_le_bytes(id_bytes);
 
                     Self {
                         id,
@@ -399,9 +399,7 @@ pub fn generate_struct_new_method_impl(
         }
     } else {
         quote! {}
-    };
-
-    new_method_impl
+    }
 }
 
 /// Given a set of idents and tokens, construct the `Entity` and `Json` implementations
--- a/packages/fuel-indexer-macros/src/helpers.rs
+++ b/packages/fuel-indexer-macros/src/helpers.rs
@@ -23,6 +23,16 @@ lazy_static! {
         "Bytes8",
         "Blob",
         "Charfield",
+        "Json",
+        "Bytes",
+        "Bytes20",
+        "Salt",
+        "MessageId",
+        "Address",
+        "ContractId",
+        "AssetId",
+        "HexString",
+        "TxId",
     ]);
 }
 /// If TypeDeclaration is tuple type
--- a/packages/fuel-indexer-macros/src/helpers.rs
+++ b/packages/fuel-indexer-macros/src/helpers.rs
@@ -371,7 +371,7 @@ pub fn field_extractor(
     }
 }
 
-/// Construct a `.new()` method for a particular struct; this `.new()` method
+/// Construct a `::new()` method for a particular struct; `::new()`
 /// will automatically create an ID for the user for use in a database.
 pub fn generate_struct_new_method_impl(
     strct: Ident,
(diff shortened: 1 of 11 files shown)


packages/fuel-indexer-plugin/Cargo.toml
```diff
@@ -26,6 +26,7 @@ hex = "0.4"
 serde = { workspace = true }
 serde_json = { workspace = true }
 sha2 = { version = "0.10" }
+tai64 = { version = "4.0.0" }
```
ra0x3, line 29: Can we not import this from `fuel-indexer-types` ?

packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -350,22 +361,75 @@ fn process_type_def(
...
                 strct_fields,
                 type_id,
                 field_extractors,
                 from_row,
                 to_row,
                 schema.is_native,
-            ))
+            );
+
+            let new_method_impl = if field_set.contains(&"id".to_string()) {
```
ra0x3, line 415: Can we use `fuel_indexer_database_types::IdCol::to_lowercase_str()` (or something) here?

packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -350,22 +361,75 @@ fn process_type_def(
...
+                let to_bytes = if EXTERNAL_FIELD_TYPES.contains(field_typ_name.as_str()) {
+                    to_bytes_method_for_external_type(field_typ_name.clone())
+                } else if !ASREF_BYTE_TYPES.contains(field_typ_name.as_str()) {
+                    quote! { .to_le_bytes() }
+                } else {
+                    quote! {}
+                };
+
+                if field_set.contains(&"id".to_string())
+                    && !INTERNAL_INDEXER_ENTITIES.contains(object_name.as_str())
```
ra0x3, line 384: Same comment here with regard to the magic `"id"` string. Let's use `IdCol` if possible.

packages/fuel-indexer-macros/src/helpers.rs
```diff
@@ -1,10 +1,107 @@
...
+/// Provides a TokenStream to be used as a conversion to bytes
+/// for external types; this is done because traits cannot be
+/// implemented on external types due to the orphan rule.
+pub fn to_bytes_method_for_external_type(
+    field_type_name: String,
+) -> proc_macro2::TokenStream {
+    match field_type_name.as_str() {
+        "Identity" => quote! { .0 },
+        "Tai64Timestamp" => quote! { .0.to_le_bytes() },
+        _ => panic!(
```
ra0x3, line 87: - Maybe?
```suggestion
        _ => panic!("From<{field_type_name}> not implemented for AsRef<u8>.");
```
- Suggesting that the rust-y explanation might be easier to understand

ra0x3 (commented):
@deekerno 

- Looks mostly ok in terms of functionality
- We discussed on a call the "style" changes
- I left some small pieces of feedback
- The doc comments make this much more understandable 👌🏽

packages/fuel-indexer-macros/src/schema.rs
```diff
@@ -350,22 +361,75 @@ fn process_type_def(
...
+                    to_bytes_method_for_external_type(field_typ_name.clone())
+                } else if !ASREF_BYTE_TYPES.contains(field_typ_name.as_str()) {
+                    quote! { .to_le_bytes() }
+                } else {
+                    quote! {}
+                };
+
+                if field_set.contains(&IdCol::to_lowercase_string())
+                    && !INTERNAL_INDEXER_ENTITIES.contains(object_name.as_str())
+                    && field_name != *"id"
```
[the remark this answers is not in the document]
ra0x3: @deekerno Still magic string?

packages/fuel-indexer-macros/src/helpers.rs
```diff
@@ -1,10 +1,105 @@
...
+
+/// Provides a TokenStream to be used for unwrapping `Option`s
+/// for external types; this is done because traits cannot be
+/// implemented on external types due to the orphan rule.
+pub fn unwrap_or_default_for_external_type(
+    field_type_name: String,
+) -> proc_macro2::TokenStream {
+    match field_type_name.as_str() {
+        "Tai64Timestamp" => quote! { .unwrap_or(Tai64(0u64)) },
+        _ => panic!("No default value available for external type {field_type_name}"),
```
ra0x3, line 99: Maybe the same error message as `to_bytes_method_for_external_type`
```suggestion
        _ => panic!("Default is not implemented for {field_type_name}"),
```

packages/fuel-indexer-types/src/lib.rs
```diff
@@ -29,6 +29,8 @@ pub mod prelude {
     pub use crate::{TypeId, FUEL_TYPES_NAMESPACE};
 }
 
+pub use tai64::Tai64;
```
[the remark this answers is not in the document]
deekerno: Oh whoops, I knew something felt wrong. 👍🏽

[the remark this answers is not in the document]
deekerno: Since type aliases can't be used as constructors, I opted to do:

```rust
quote! {
    .unwrap_or(Tai64Timestamp::from({
        <[u8;8]>::try_from(0u64.to_be_bytes()).expect("Failed to create byte slice from u64")
    }))
}
```

as it's not possible to do:

```rust
quote! {
    .unwrap_or(Tai64Timestamp(0u64))
}
```

ra0x3 (approved):

PR #999, opened 2023-06-13
Upgrade Fuel ecosystem dependencies

lostman:
### Description

This PR updates Fuel dependencies.

### Testing steps

Please provide the _exact_ testing steps for the reviewer(s) if this PR requires testing.

### Changelog

* fuels-code-gen 0.43
* fuels-macros 0.43
* fuels-types 0.42
* Sway (for fuel-indexer-test) 0.40.1
* fuel-core 0.18.1
* fuel-core-client 0.18.1
* fuel-crypto 0.31.2
* fuels 0.43

PR #1007, opened 2023-06-14
chore: rustdoc + cleanup fuel-indexer-lib

ra0x3:

#### ⚠️ I'm going through all crates and fixing hacks, cleaning up, adding clarity and rustdoc comments. Starting with `fuel-indexer-lib`

- For merge after https://github.com/FuelLabs/fuel-indexer/pull/999
- This cleans up the fuel-indexer-lib module
- Adds rustdoc comments
- Updates README briefly
- **No code logic is changed, code is only moved around for clarity**

### Testing steps

- N/A

### Changelog

- chore: rustdoc + cleanup fuel-indexer-lib


packages/fuel-indexer-lib/src/config/cli.rs
```diff
@@ -0,0 +1,300 @@
+pub use crate::{
```
ra0x3, line 1: This module is configuration related to the CLI.

packages/fuel-indexer-lib/src/config/mod.rs
```diff
@@ -66,292 +45,29 @@ pub fn env_or_default(var: EnvVar, default: String) -> String {
     std::env::var(var.as_ref()).unwrap_or(default)
 }
 
-#[derive(Debug, Parser, Clone)]
```
ra0x3, line 69: Removed the cli-related stuff to its own module

packages/fuel-indexer-lib/src/utils.rs
```diff
@@ -229,71 +236,3 @@ pub async fn init_logging(config: &IndexerConfig) -> anyhow::Result<()> {
     }
     Ok(())
 }
-
-pub fn format_exec_msg(exec_name: &str, path: Option<String>) -> String {
```
ra0x3, line 233: These functions are only used by forc-index so they were moved to forc-index

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -86,7 +86,7 @@ pub(crate) async fn get_fuel_status(config: &IndexerConfig) -> ServiceStatus {
                 .await
                 .unwrap_or_default();
 
-            let fuel_node_health: FuelNodeHealthResponse =
+            let fuel_node_health: FuelClientHealthResponse =
```
deekerno, line 89: I see that we're referring to `FuelClient*`, which is clearer; can we also rename `fuel_node_health` to `fuel_client_health` to match?

packages/fuel-indexer-tests/src/fixtures.rs
```diff
@@ -287,7 +287,7 @@ pub async fn api_server_app_postgres(
         verbose: true,
         local_fuel_node: false,
         indexer_net_config: false,
-        fuel_node: FuelNodeConfig::default(),
+        fuel_node: FuelClientConfig::default(),
```
deekerno, line 290: Can we also rename `fuel_node` to `fuel_client`?

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -22,8 +22,8 @@ use fuel_indexer_lib::{
     config::{auth::AuthenticationStrategy, IndexerConfig},
     defaults,
     utils::{
-        AssetReloadRequest, FuelNodeHealthResponse, IndexStopRequest, ServiceRequest,
-        ServiceStatus,
+        FuelClientHealthResponse, ReloadRequest, ServiceRequest, ServiceStatus,
```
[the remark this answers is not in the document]
deekerno: Agreed.

packages/fuel-indexer-lib/src/utils.rs
```diff
@@ -67,36 +66,40 @@ pub fn local_repository_root() -> Option<String> {
...
 #[derive(Debug)]
-pub struct AssetReloadRequest {
+pub struct ReloadRequest {
     pub namespace: String,
     pub identifier: String,
 }
 
+/// Request to remove the specified indexer executor from the indexer service.
+///
+/// Sent from API server to indexer service.
```
deekerno, line 81: We can probably remove this line as it's already mentioned in the doc comment for `ServiceRequest`.

packages/fuel-indexer-lib/src/utils.rs
deekerno: Nice. 👌🏽

deekerno (changes requested):
Looks great! Just left some comments about consistency and redundancy.

packages/fuel-indexer-api-server/src/uses.rs
ra0x3: @deekerno  That's the thing. If we rename _everything_ to `FuelClient` the footprint is obviously much larger, and includes a few breaking changes. I originally actually had this (changing _everything_) but felt that was a little too large of a footprint for this cleanup - and that should be in its own separate breaking PR.

@lostman If we're all aligned to change these namings to `FuelClient*` let's do that separately

packages/fuel-indexer-tests/src/fixtures.rs
ra0x3: [See comment](https://github.com/FuelLabs/fuel-indexer/pull/1007#discussion_r1238597585)

deekerno (approved):
Approving in light of @ra0x3's comment.

deekerno (approved):

PR #1009, opened 2023-06-15
enhancement: add schema cache to API server

deekerno:
### Description

Adds an LRU cache for `Schema`s that have been loaded from the DB.

### Testing steps

CI should pass.

#### Manual Testing
Compare the logs and speed of loading the schema between develop and this branch. This branch should be noticeably faster.

### Changelog

- [Add schema cache to API server](https://github.com/FuelLabs/fuel-indexer/commit/613f79ed576253e5c60cb5a9753a205afacb65b4)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2993,6 +2993,7 @@ dependencies = [
  "tower",
  "tower-http",
  "tracing",
+ "uluru",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


ra0x3:
@deekerno 
- Now that https://github.com/FuelLabs/fuel-indexer/pull/1040 is merged, I think you can pick this back up
- Keep in mind we'll want some idea of "before this PR" vs "after this PR" timing when it comes to `IndexerSchema::load` timing

deekerno:
@ra0x3: Funnily enough, this is already done (ignoring the conflicts, of course). However, I do want to get some basic bench-marking in so that we actually have quantifiable timings. Thus, I'm going to close it for now so that it doesn't clutter the PR list and re-open it after I put up the bench-marking PR later this week.

PR #1016, opened 2023-06-17
chore: impl IDs into explorer

ra0x3:

### Description

- This PR implements the `::new()` and `.get_or_create()` functionality into the explorer
- Now all entities finally have IDs! 🥳 

### Testing steps

- [ ] Should be able to build and deploy the `fuel-explorer` indexer
- [ ] Can't test on beta-3

### Changelog

- chore: impl IDs into explorer


examples/fuel-explorer/fuel-explorer/src/lib.rs
deekerno, line 1: It's so beautiful with the new instantiation stuff. 🥲

examples/fuel-explorer/fuel-explorer/schema/fuel_explorer.schema.graphql
```diff
@@ -43,17 +43,20 @@ type Genesis {
   coins_root: Bytes32!
   contracts_root: Bytes32!
   messages_root: Bytes32!
+  is_genesis: Boolean!
```
deekerno, line 46: I may be missing something, but what is the purpose of these boolean flags? Is there something in the `union` parsing that relies upon them?

ra0x3: So that you can query a union for the single type that you want. (I also included the label field, which you could also use, but more simple to just use the boolean)

PR #1018, opened 2023-06-19
Add logging macros

### LINKED ISSUE #1014: Make `Logger` take args like `println`
- Indexers use our `Logger` abstraction to log output to stdout/stderr 
  - `println!` will not work in an indexer
- If I want to log a string to stdout with an arbitrary number of arguments in an indexer I have to:

```rust
Logger::info!(&format!("{} {} {}", "one", "two", "three"));
```

- This is a bit tedious, it would be better if we made `Logger` work like `println!`
- So that I could:

```rust
Logger::info("{} {} {}", "one", "two", "three");
```

Additional context
- Would need this for WASM execution and native execution

lostman:
### Description

This PR adds logging macros, which simplify the usage of `Logger`.

Closes #1014.

### Testing steps

No need for testing beyond automated testing.

### Changelog

* Add `log_error, log_warn, log_info, log_debug, log_trace` macros.

### DIFF
--- a/plugins/forc-index/src/utils/defaults.rs
+++ b/plugins/forc-index/src/utils/defaults.rs
@@ -26,7 +26,7 @@ name = "{indexer_name}"
 path = "src/main.rs"
 
 [dependencies]
-fuel-indexer-utils = {{ version = "0.17", features = ["native-execution"] }}
+fuel-indexer-utils = {{ version = "0.16", features = ["native-execution"] }}
 fuels = {{ version = "0.43", default-features = false, features = ["std"] }}
 serde = {{ version = "1.0", default-features = false, features = ["derive"] }}
 "#


examples/hello-world-native/hello-indexer-native/src/main.rs
```diff
@@ -29,6 +29,7 @@ use fuel_indexer_utils::prelude::*;
 mod hello_world_native {
 
     async fn index_logged_greeting(event: Greeting, block: BlockData) {
+        log_info!("Handling {:?}", event);
```
ra0x3, line 32: Would it be possible to change these to their `tracing` equivalent? So in this case instead of `log_info!`, just `info!` ?

ra0x3 (approved):

deekerno (approved):
Looks great, thanks for this!

PR #1019, opened 2023-06-20
enhancement: allow for list fields to be stored in database

deekerno:
_Note: This is being merged into `deekerno/support-entity-list-fields-in-indexers` branch._

### Description

This PR primarily adds the ability to persist values for list fields into the database for so that a user can query for them. Note that this PR does not implement querying; that will be included in the next PR. It also adds doc comments and does some light renaming.

### Testing steps

CI should pass. A new E2E test case was added with both a list of scalar values and a list of foreign key IDs.


deekerno:
Closing this as the work has been moved into #1040.

PR #1038, opened 2023-06-28
forc index kill: terminate or kill

lostman:
### Description

This PR changes the default behavior of `forc index kill` to send the `TERM` signal and adds a `-9` flag to send the `KILL` signal instead. This way, the behavior matches the Linux `kill` command.

The motivation for this is that `fuel-indexer` may be running with `--embedded-database`, in which case the current behavior of `forc index kill` won't allow `fuel-indexer` to shut down the embedded `postgres` process—it will be left alive.

If the user attempts to start `fuel-indexer` again with `--embedded` database, they will get an error.

### Testing steps

### Changelog


ra0x3 (approved):
utACK

PR #1044, opened 2023-06-29
fix: use proper type for block height

ra0x3:

### Description

- The `resumable` flag in the manifest never worked due to a type issue
- Indexer would always start at block `1` regardless
- How did no one catch this yet? 😭 

### Testing steps

- [ ] Start the service with a test indexer `fuel-indexer-test`
  - Make sure the manifest for the indexer specifies `resumable: true`
- [ ] Let it index a few blocks
- [ ] Stop the service
- [ ] Restart the service (without a `--manifest`, and let it use the indexer in the DB)
- [ ] Should see the output saying it's resuming

```text
2023-06-29T19:22:58.657046Z  INFO fuel_indexer::service: 339: Resuming index from block 26369
```

### Changelog

- fix: use proper type for block height


deekerno (approved):
Works as described.

PR #1059, opened 2023-07-04
reuse IndexerConnectionPool

lostman:
### Description

We're passing `db_url` down and creating multiple `sqlx::Pool` objects. I think this is unnecessary. We can instead pass the `Pool`.


ra0x3 (approved):
- I wasn't able to see a discernible difference (in resource usage) between re-creating the pool, and cloning the pool, but this looks fine

deekerno (approved):
Ran multiple indexers at once and things appear to work as before with no degradation on my side.

PR #1060, opened 2023-07-04
Enable WASM metering and add benchmarks

lostman:
### Description

This PR adds an option to enable WASM metering middleware in `WasmIndexExecutor`. The metering assigns a cost of one point to each opcode. Executor is supplied with points, and when these run out, the execution is stopped.

This is intended to replace the `tokio::time::timeout`, which can't interrupt `spawn_blocking` threads, or stop WASM modules in an infinite loop.

### Testing steps

To run `criterion` benchmarks:
```
cargo bench
```
To also run Rust's benchmarks:
```
cargo bench --features benchmarks
```
The feature flag is required because benchmarks are still an unstable feature 🤷‍♂️.

I ran into issues with `criterion`, so I'd like to have both for now.

After some tweaking, the results from both benchmarks align.

#### fuel-indexer with metering

```
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --manifest packages/fuel-indexer-tests/components/indices/fuel-indexer-test/fuel_indexer_test.yaml --indexer-handler-metering-points 100000000 --replace-indexer
```

will fail with:

```
2023-07-06T08:47:33.465076Z ERROR fuel_indexer::executor: 386: Indexer executor run time limit exceeded. Giving up. <('.')>. Consider increasing metering points
```

after encountering a large block.

Increasing `--indexer-handler-metering-points 10000000000` let's `fuel-indexer` run smoothly.

#### looping executors

Another test case is submitting an executor with a `loop {}`. For instance, changing the `fuel-indexer-test` executor to:

```rust
mod fuel_indexer_test {

    fn fuel_indexer_test_blocks(block_data: BlockData) {
        loop {}
    }
}
```

Deploying the above should result in
```
2023-07-07T14:10:19.239462Z ERROR fuel_indexer::executor: 387: Indexer executor run time limit exceeded. Giving up. <('.')>. Consider increasing metering points
```

### Changelog

* Add an option to enable metering middleware.
* Replace timeout with metering.
* Add criterion benchmarks.
* Add ordinary Rust benchmarks.

### DIFF
--- a/packages/fuel-indexer-benchmarks/Cargo.toml
+++ b/packages/fuel-indexer-benchmarks/Cargo.toml
@@ -1,7 +1,9 @@
 [package]
 name = "fuel-indexer-benchmarks"
-version = "0.1.0"
+version = "0.0.0"
 edition = "2021"
+license = "BUSL-1.1"
+publish = false
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3411,7 +3411,7 @@ dependencies = [
 
 [[package]]
 name = "fuel-indexer-benchmarks"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "criterion",
  "fuel-indexer",


packages/fuel-indexer-benchmarks/benches/wasm_metering_benchmark.rs
```diff
@@ -0,0 +1,79 @@
...
+    .unwrap();
+
+    let wasm_bytes = match &manifest.module {
+        fuel_indexer_lib::manifest::Module::Wasm(ref module) => {
+            let mut bytes = Vec::<u8>::new();
+            let mut file = File::open(module).unwrap();
+            file.read_to_end(&mut bytes).unwrap();
+            bytes
+        }
+        _ => panic!("unexpected"),
```
ra0x3, line 30: Can we get an actual error message here?

packages/fuel-indexer-benchmarks/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+        for _ in 0..16 {
+            b.iter(|| {
+                rt.block_on(executor.handle_events(blocks.clone())).unwrap();
+            })
+        }
+    }
+
+    #[bench]
+    fn executor_with_metering(b: &mut Bencher) {
+        make_executor_metering_benchmark(b, Some(9_000_000_000_000));
```
ra0x3, line 72: What is this magic number?

packages/fuel-indexer-lib/src/config/cli.rs
```diff
@@ -162,13 +162,13 @@ pub struct IndexerArgs {
...
+    /// The number of WASM opcodes after which the indexer's event handler will stop execution.
     #[clap(
         long,
-        default_value_t = defaults::INDEXER_HANDLER_TIMEOUT,
-        help = "Maximum length of time (in seconds) that an indexer's event handler can run before timing out."
+        help = "The number of WASM opcodes after which the indexer's event handler will stop execution.",
+        default_value_t = 30_000_000_000
     )]
-    pub indexer_handler_timeout: u64,
+    pub indexer_handler_metering_points: u64,
```
ra0x3, line 171: - I like the descriptiveness of the naming here, but let's shorten this. Maybe `max_metering_points` ? 
  - The fact that these points are for an indexer handler are implied (and mentioned explicitly in the opt description)

packages/fuel-indexer-lib/src/config/cli.rs
```diff
@@ -162,13 +162,13 @@ pub struct IndexerArgs {
...
     pub rate_limit_window_size: Option<u64>,
 
-    /// Maximum length of time (in seconds) that an indexer's event handler can run before timing out.
+    /// The number of WASM opcodes after which the indexer's event handler will stop execution.
     #[clap(
         long,
-        default_value_t = defaults::INDEXER_HANDLER_TIMEOUT,
-        help = "Maximum length of time (in seconds) that an indexer's event handler can run before timing out."
+        help = "The number of WASM opcodes after which the indexer's event handler will stop execution.",
+        default_value_t = 30_000_000_000
```
ra0x3, line 169: Let's also remove this magic number as well. We use defaults

packages/fuel-indexer-lib/src/defaults.rs
```diff
@@ -112,8 +112,8 @@ pub const RATE_LIMIT_WINDOW_SIZE: u64 = 5;
 /// Log level.
 pub const LOG_LEVEL: &str = "info";
 
-/// Maximum amount of time (seconds) that an indexer's `handle_events` function can take before being timed out.
-pub const INDEXER_HANDLER_TIMEOUT: u64 = 5;
+/// The number of WASM opcodes after which the indexer's `handle_events` function will stop execution.
+pub const INDEXER_HANDLER_METERING_POINTS: u64 = 30_000_000_000;
```
ra0x3, line 116: Nice! Let's use this default everywhere instead of any magic numbers

packages/fuel-indexer-tests/tests/integration/service.rs
```diff
@@ -22,6 +27,76 @@ abigen!(Contract(
         "packages/fuel-indexer-tests/contracts/simple-wasm/out/debug/contracts-abi.json"
 ));
 
+#[tokio::test]
+async fn test_wasm_executor_can_meter_execution() {
```
ra0x3, line 31: 🏅

packages/fuel-indexer-tests/tests/integration/service.rs
```diff
@@ -22,6 +27,76 @@ abigen!(Contract(
...
+                    match executor.get_instance_metering_points().await {
+                        wasmer_middlewares::metering::MeteringPoints::Remaining(pts) => {
+                            assert!(pts > 0)
+                        }
+                        _ => panic!("Expected remaining points > 0"),
+                    }
+                }
+            }
+        }
+        _ => panic!("Unexpected!"),
```
ra0x3, line 96: Can we get an error message here?

packages/fuel-indexer/src/executor.rs
```diff
@@ -657,17 +675,44 @@ impl WasmIndexExecutor {
             }
         }
     }
+
+    pub async fn get_instance_metering_points(
```
ra0x3, line 679: @lostman Going forward, we're trying to add `rustdoc` comments to any new functions we create or any previous functions/methods that we change. Can we do that here?

Makes the code much more easy to follow for others.

packages/fuel-indexer/src/executor.rs
```diff
@@ -683,38 +728,32 @@ impl Executor for WasmIndexExecutor {
...
+                fun.call(&mut store_guard, ptr, len)
             }
-            Ok(Err(e)) => {
-                error!("WasmIndexExecutor handle_events failed: {e:?}.");
+        })
+        .await?;
+
+        if let Err(e) = res {
+            // get_instance_metering_points panics if metering is not enabled.
+            if self.metering_points.is_some()
```
ra0x3, line 743: I think this can be changed a bit to be more readable. This `if` extends for 4 lines 😅

packages/fuel-indexer-benchmarks/benches/wasm_metering_benchmark.rs
```diff
@@ -0,0 +1,79 @@
...
+            })
+        });
+    }
+    group.finish();
+}
+
+criterion_group! {
+    name = benches;
+    // This can be any expression that returns a `Criterion` object.
+    config = Criterion::default().sample_size(50).measurement_time(std::time::Duration::from_secs(20));
```
ra0x3, line 76: More magic numbers here.

packages/fuel-indexer-benchmarks/benches/wasm_metering_benchmark.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let mut group: criterion::BenchmarkGroup<'_, criterion::measurement::WallTime> =
+        c.benchmark_group("executor");
+    for t in [true, false] {
+        let rt = tokio::runtime::Runtime::new().unwrap();
+
+        let mut executor = rt.block_on(async {
+            let mut config = IndexerConfig::default();
+            config.indexer_handler_metering_points =
+                if t { Some(9_000_000_000_000) } else { None };
```
ra0x3, line 41: Magic 🪄

ra0x3 (changes requested):
@lostman 
- Nice work here! Testing steps work as described
- Left a few comments regarding style/form
- This code, while it works, seems a bit thrown-together -- I realize this is just for benchmarks but let's try to clean some of this stuff up (no magic numbers, add rustdoc comments, try to unwrap as little as possible, specific/descriptive error messages -- for the person who might have to work in this part of the codebase later :)

Again, works fine (one failing test on CI?). Nice work

packages/fuel-indexer-benchmarks/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+            .await
+            .unwrap();
+
+            WasmIndexExecutor::new(&config, &manifest, wasm_bytes, pool)
+                .await
+                .unwrap()
+        });
+
+        let blocks: Vec<fuel_indexer_types::fuel::BlockData> = vec![];
+        for _ in 0..16 {
```
ra0x3, line 63: @lostman Why does this only go from `0..16`? More magic numbers?

packages/fuel-indexer-lib/src/config/cli.rs
```diff
@@ -162,13 +162,13 @@ pub struct IndexerArgs {
...
+    /// The number of WASM opcodes after which the indexer's event handler will stop execution.
     #[clap(
         long,
-        default_value_t = defaults::INDEXER_HANDLER_TIMEOUT,
-        help = "Maximum length of time (in seconds) that an indexer's event handler can run before timing out."
+        help = "The number of WASM opcodes after which the indexer's event handler will stop execution.",
+        default_value_t = defaults::METERING_POINTS
     )]
-    pub indexer_handler_timeout: u64,
+    pub metering_points: u64,
```
ra0x3, line 171: @lostman Don't forget to update `config.yaml` with this new variable (might have to replace the old one) as well. I've created an issue to make sure we add CI to detect this automatically

packages/fuel-indexer-lib/src/config/mod.rs
```diff
@@ -316,8 +316,7 @@ impl IndexerConfig {
 
         let log_level_key = serde_yaml::Value::String("log_level".into());
         let replace_indexer_key = serde_yaml::Value::String("replace_indexer".into());
-        let indexer_handler_timeout_key =
-            serde_yaml::Value::String("indexer_handler_timeout".into());
+        let metering_points_key = serde_yaml::Value::String("metering_points".into());
```
ra0x3, line 319: Nice 👌🏽  but I don't see any changes to `config.yaml`. See my comment above

packages/fuel-indexer/src/executor.rs
```diff
@@ -488,7 +495,15 @@ where
         pool: IndexerConnectionPool,
         handle_events_fn: fn(Vec<BlockData>, Arc<Mutex<Database>>) -> F,
     ) -> IndexerResult<Self> {
-        let db = Database::new(pool, manifest).await?;
+        let mut db = Database::new(pool.clone(), manifest).await;
+        let mut conn = pool.acquire().await?;
+        let version = fuel_indexer_database::queries::type_id_latest(
```
ra0x3, line 500: Unrelated to this PR, but as you previously mentioned `Manifest::graphql_schema` should produce a `GraphQLSchema` from which we can get the `version` by just calling `GraphQLSchema::version`.So this could eventually be updated to:

```rust
let version = manifest.graphql_schema().version();
```

ra0x3 (commented):
@lostman 
- Left 2 pieces of feedback related to updating config.yaml and some possible more magic numbers
- @deekerno Will wait for you to review this as well - let's make sure we review the docker components as well
  - This change contains some non-trivial lower-level changes, so wanna make sure everything works
  - The change looks ok to me but better thorough than sorry

packages/fuel-indexer-benchmarks/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+                .await
+                .unwrap()
+        });
+
+        let blocks: Vec<fuel_indexer_types::fuel::BlockData> = vec![];
+
+        b.iter(|| {
+            rt.block_on(executor.handle_events(blocks.clone())).unwrap();
+        })
+    }
```
deekerno, line 67: Nice. You can also consider returning a closure similar to what I did in my micro-benchmarking branch: https://github.com/FuelLabs/fuel-indexer/compare/develop...deekerno/1034-add-benchmarking#diff-d85d40698746760eb4b7a4927f8f10044bc118408b876197a42d4146425d38bfR56-R93

packages/fuel-indexer/src/ffi.rs
```diff
@@ -202,8 +232,21 @@ impl<'a> WasmArg<'a> {
...
-        dealloc_fn
-            .call(store, self.ptr, self.len)
-            .expect("Dealloc failed");
+        // Need to track whether metering is enabled or otherwise getting or setting points will panic
+        if self.metering_enabled {
+            let pts = match get_remaining_points(store, self.instance) {
+                MeteringPoints::Exhausted => 0,
+                MeteringPoints::Remaining(pts) => pts,
+            };
+            set_remaining_points(store, self.instance, 1_000_000);
```
deekerno, line 241: Is there some significance to the 1M number here? It seems like it may not be necessary for `dealloc_fn` as I can't see that costing too much; additionally, we set it immediately after the call. Also, is there a reason that we get the remaining points but don't use it until after the deallocation?

packages/fuel-indexer/src/executor.rs
```diff
@@ -683,38 +770,29 @@ impl Executor for WasmIndexExecutor {
...
+                fun.call(&mut store_guard, ptr, len)
             }
-            Ok(Err(e)) => {
-                error!("WasmIndexExecutor handle_events failed: {e:?}.");
+        })
+        .await?;
+
+        if let Err(e) = res {
+            if e.clone().to_trap() == Some(TrapCode::UnreachableCodeReached)
+                && self.metering_points_exhausted().await
```
deekerno, line 785: Just to be clear: Reaching the metering limit results in `TrapCode::UnreachableCodeReached` and never anything else, yeah?

packages/fuel-indexer/src/executor.rs
```diff
@@ -551,10 +565,23 @@ impl WasmIndexExecutor {
...
     ) -> IndexerResult<Self> {
+        let mut compiler_config = Cranelift::new();
+
+        if let Some(metering_points) = config.metering_points {
+            // `Metering` needs to be configured with a limit and a cost
+            // function. For each `Operator`, the metering middleware will call
+            // the cost function and subtract the cost from the remaining
+            // points.
+            let metering =
+                Arc::new(wasmer_middlewares::Metering::new(metering_points, |_| 1));
```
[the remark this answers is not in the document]
ra0x3: @deekerno Sounds like a great issue to create (as an FYI or future piece of work so we don't forget).

packages/fuel-indexer/src/ffi.rs
ra0x3: I missed this earlier, but @lostman #NoMagicNumbers - very very rarely should be including raw values in production code :) - and if you see us doing this elsewhere in the repo bring it to our attention so we can correct it

ra0x3:
@lostman Last piece of feedback from me on this. Don't forget to update the deployment with this new WASM metering option (and remove the old option). `fuel-indexer-deployment.yaml`

deekerno (approved):

ra0x3 (approved):
LGTM. Will defer to @deekerno. Thanks for this @lostman

packages/fuel-indexer-benchmarks/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "fuel-indexer-benchmarks"
+version = "0.1.0"
```
ra0x3, line 3: @lostman If this is a non-publishable crate, this should be `0.0.0`

ra0x3, line 3: @lostman If this is a non-publishable crate, this should be `0.0.0`

Here's what's in our `fuel-indexer-tests` manifest.


```toml
name = "fuel-indexer-tests"
version = "0.0.0"
edition = "2021"
license = "BUSL-1.1"
publish = false
```

ra0x3 (changes requested):

ra0x3 (approved):

deekerno (approved):

PR #1073, opened 2023-07-10
v0.18.0

ra0x3:

### Description

Please include a short derscription of what this PR does.

### Testing steps

### Changelog


plugins/forc-index/src/defaults.rs
deekerno, line 1: We don't usually do patch versions in the defaults; is there a reason we're doing it here?

deekerno (changes requested):

deekerno (approved):
💯

PR #1076, opened 2023-07-10
chore: Remove `QueryRoot` workaround

deekerno:
### Description

Removes dependence on the concept of a `QueryRoot` in queries and schema construction.

### Testing steps

CI should pass. All functionality should still be the same; this should have no outward effect on the user.

### DIFF
--- a/packages/fuel-indexer-lib/src/graphql/parser.rs
+++ b/packages/fuel-indexer-lib/src/graphql/parser.rs
@@ -635,7 +635,7 @@ impl ParsedGraphQLSchema {
     }
 
     /// Return the GraphQL type for a given field name.
-    pub fn field_type(&self, cond: &str, name: &str) -> Option<&String> {
+    fn field_type(&self, cond: &str, name: &str) -> Option<&String> {
         match self.object_field_mappings().get(cond) {
             Some(fieldset) => fieldset.get(name),
             _ => {
@@ -648,10 +648,12 @@ impl ParsedGraphQLSchema {
         }
     }
 
+    /// Return the GraphQL type for a given entity.
     fn entity_type(&self, name: &str) -> Option<&String> {
         self.entity_names_to_types.get(name)
     }
 
+    /// Return the GraphQL type for a given entity or field.
     pub fn graphql_type(&self, cond: Option<&String>, name: &str) -> Option<&String> {
         match cond {
             Some(c) => self.field_type(c, name),
--- a/packages/fuel-indexer-lib/src/graphql/parser.rs
+++ b/packages/fuel-indexer-lib/src/graphql/parser.rs
@@ -144,6 +144,10 @@ pub struct ParsedGraphQLSchema {
     /// All unique names of types in the schema (whether objects, enums, or scalars).
     type_names: HashSet<String>,
 
+    /// Mapping of lowercase entity names to their actual entity types.
+    // Used to refer to top-level entities in GraphQL queries.
+    entity_names_to_types: HashMap<String, String>,
+
     /// Mapping of object names to objects.
     objects: HashMap<String, ObjectType>,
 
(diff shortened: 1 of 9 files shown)


packages/fuel-indexer-lib/src/graphql/parser.rs
```diff
@@ -144,6 +144,10 @@ pub struct ParsedGraphQLSchema {
     /// All unique names of types in the schema (whether objects, enums, or scalars).
     type_names: HashSet<String>,
 
+    /// Mapping of lowercase entity names to their actual entity types.
+    // Used to refer to top-level entities in GraphQL queries.
+    entity_names_to_types: HashMap<String, String>,
```
ra0x3, line 149: - So going forward let's be a bit better/consistent about the naming. 
- If these are `TypeDefinition` names (looks like they are?) then let's refer to them as that instead of "entity"

packages/fuel-indexer-lib/src/graphql/parser.rs
```diff
@@ -633,6 +648,19 @@ impl ParsedGraphQLSchema {
         }
     }
 
+    /// Return the GraphQL type for a given entity.
+    fn entity_type(&self, name: &str) -> Option<&String> {
```
ra0x3, line 652: Same comment here regarding this "entity" naming stuff.

packages/fuel-indexer-schema/src/db/tables.rs
```diff
@@ -220,37 +219,13 @@ impl IndexerSchema {
...
             db_type: DbType::Postgres,
-        };
-
-        schema.register_queryroot_fields();
-
-        Ok(schema)
-    }
-
-    /// Register the `QueryRoot` type and its corresponding field types.
-    pub fn register_queryroot_fields(&mut self) {
```
ra0x3, line 238: 🥇 🥇 🥇

packages/fuel-indexer-lib/src/graphql/parser.rs
```diff
@@ -633,6 +648,19 @@ impl ParsedGraphQLSchema {
         }
     }
 
+    /// Return the GraphQL type for a given entity.
```
ra0x3, line 651:
```suggestion
    /// Return the GraphQL type for a given `TypeDefinition` name.
```

packages/fuel-indexer-lib/src/graphql/parser.rs
```diff
@@ -620,7 +635,7 @@ impl ParsedGraphQLSchema {
     }
 
     /// Return the GraphQL type for a given field name.
```
ra0x3, line 637:
```suggestion
    /// Return the GraphQL type for a given `FieldDefinition` name.
```

packages/fuel-indexer-lib/src/graphql/parser.rs
```diff
@@ -633,6 +648,19 @@ impl ParsedGraphQLSchema {
         }
     }
 
+    /// Return the GraphQL type for a given entity.
+    fn entity_type(&self, name: &str) -> Option<&String> {
+        self.entity_names_to_types.get(name)
+    }
+
+    /// Return the GraphQL type for a given entity or field.
```
ra0x3, line 656:
```suggestion
    /// Return the GraphQL type for a given `TypeDefinition` name or `FieldDefinition` name.
```
- I think there should also be a little comment here explaining why we use this in `ParsedGraphQLSchema` as opposed to letting the caller decide to call `ParsedGraphQLSchema::field_type` then `ParsedGraphQLSchema::entity_type` itself. If the answer is "convenience" that's totally fine

ra0x3 (changes requested):
@deekerno 
Awesome! Looks good.

Left a few small comments regarding clarity and naming. Otherwise looks good 👍🏽

packages/fuel-indexer-lib/src/graphql/parser.rs
deekerno: It's just convenience. 👍🏽 `QUERY_ROOT` is essentially communicated using a `None` value for `field_type` in `Selection::new()` now, so I decided to hide that detail away so that all we have to do is just pass `field_type` into `ParsedGraphQLSchema::graphql_type()` to get the correct value.

ra0x3 (approved):

PR #1086, opened 2023-07-11
chore: re-enable TRYBUILD tests

deekerno:
### Description

Re-enable TRYBUILD tests.

### Testing steps

CI should pass.

### Notes

The change to `fuel-indexer-macros` dev-dependencies was taken from this PR from the `futures` crate: https://github.com/rust-lang/futures-rs/pull/2305.


packages/fuel-indexer-macros/Cargo.toml
```diff
@@ -32,7 +32,8 @@ sha2 = "0.10"
 syn = { version = "2.0", features = ["full"] }
 
 [dev-dependencies]
-fuel-indexer-plugin = { workspace = true }
+# This is a workaround to avoid cyclic dependency problems
```
ra0x3, line 35: @lostman What do you think of this?

@deekerno Maybe might be better to create a Github issue explaining the background here? We could tag it with "fyi" and link the issue here instead of this comment. Looks harmless to me?

deekerno: @ra0x3 I'll add a link and a small blurb. The `futures` crate had the same issue with publishing with a circular dependency in `dev-dependencies`; this addition worked for them due to some of the intricacies of how `rustc > 1.40` strip some information from `dev-dependencies`.

ra0x3: @deekerno Yea see stuff like that, with supporting links is super helpful context to have. We can dump that into the Github issue as well 👌🏽

deekerno: @ra0x3 I decided to keep the comment and link to the issue from the `futures` crate that contains their explanation and links to other projects that have used the same workaround. Let me know if that doesn't seem like a good idea.

ra0x3 (approved):
utACK

PR #1092, opened 2023-07-12
fix: dont panic on empty panicreceipt contract id

ra0x3:

### Description

- PR fixes an issue where we were calling `.unwrap()` on a `None` contract ID, triggering a panic, triggering a (previously) hidden WASM runtime error
- Also swaps `try_from` for `from` (if `from` fails, we will get a panic anyway, so might as well use that instead of `try_from` - less code)

### Testing steps

- [ ] Update your `start_block` to `8320` in `fuel_explorer.manifest.yaml`
- [ ] Rebuild your modules `bash scripts/utils/build_test_wasm_module.bash`
- [ ] Run the explorer on beta-3
```bash
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --manifest examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml --replace-indexer
```
- [ ] Should see no WASM runtime errors

### Changelog

- fix: dont panic on empty panicreceipt contract id


deekerno (approved):
Works as described. 💯

PR #1097, opened 2023-07-12
fix: prevent serde error from crashing service

ra0x3:

### Description

- Prevents from panic'ing when `put_object` fails. 
  - Currently this crashes the entire service
  - Why does it fail? Not sure, but we can't bring down the entire service because of this
  - We successfully indexed 178,775 blocks before encountering this error - some type of blip?
- [Here is the elastic doc with the exact error](https://search-fuel-prod-sp5nn3gesvhmb5qjjcxarse5di.us-east-1.es.amazonaws.com/_dashboards/app/discover#/doc/a9bb0e30-7cfa-11ed-bb0a-9d1026406ebc/cwl-2023.07.12?id=37670455749185772964673089726982812396625972086305981568)

### Testing steps

- [ ] New indexer `forc index new indexer-test --namespace fuel`
- [ ] Use `178775` as your `start_block`
- [ ] Start the service on beta-3
```bash
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --replace-indexer
```
- [ ] Deploy `forc index deploy`
- [ ] Should see no panic

### Changelog

- fix: prevent serde error from crashing service

### DIFF
--- a/packages/fuel-indexer-macros/src/decoder.rs
+++ b/packages/fuel-indexer-macros/src/decoder.rs
@@ -673,14 +673,14 @@ impl From<ObjectDecoder> for TokenStream {
 
             impl From<#ident> for Json {
                 fn from(value: #ident) -> Self {
-                    let s = serde_json::to_string(&value).expect("Serde error.");
+                    let s = serde_json::to_string(&value).expect("Failed to serialize Entity.");
                     Self(s)
                 }
             }
 
             impl From<Json> for #ident {
                 fn from(value: Json) -> Self {
-                    let s: #ident = serde_json::from_str(&value.0).expect("Serde error.");
+                    let s: #ident = serde_json::from_str(&value.0).expect("Failed to deserialize Entity.");
                     s
                 }
             }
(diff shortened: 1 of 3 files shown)


packages/fuel-indexer-macros/src/decoder.rs
```diff
@@ -673,14 +673,14 @@ impl From<ObjectDecoder> for TokenStream {
 
             impl From<#ident> for Json {
                 fn from(value: #ident) -> Self {
-                    let s = serde_json::to_string(&value).expect("Serde error.");
```
ra0x3, line 676: - All of these errors were `"Serde error."` 
- Hard to sort through this in the repo, so I made them unique.

packages/fuel-indexer/src/ffi.rs
```diff
@@ -154,7 +154,16 @@ fn put_object(env: &IndexEnv, type_id: i64, ptr: u32, len: u32) {
         bytes.extend_from_slice(&mem.data_unchecked()[range]);
     }
 
-    let columns: Vec<FtColumn> = bincode::deserialize(&bytes).expect("Serde error.");
```
ra0x3, line 157: This is where the actual bug is. Not sure what causes 1 out of 178,778 of these to be incorrect but we can look at the data later. For now it just shouldn't crash the service.

deekerno (approved):

PR #1101, opened 2023-07-13
Put whole test modules behind feature flags

lostman:
### Description

Instead of feature-gating individual tests, we can feature-gate entire test modules.

The benefit is that it silences ~50 warnings in Visual Studio Code (and potentially other IDEs that report all errors/warnings).

### Changelog

* move feature flags for tests


ra0x3:
@lostman I see this was ping'd for review but some tests are failing? Can review once that's fixed.

PR #1102, opened 2023-07-13
chore: add more executor logging

ra0x3:

### Description

- More detailed logging so we can always know which indexer is being referred to in the logs

### Testing steps

- N/A

### Changelog

- chore: add more executor logging


deekerno (approved):
utACK

deekerno (dismissed):

PR #1106, opened 2023-07-13
enhancement: add endpoint for raw sql queries

ra0x3:

- CC @luizstacio 

### Description

- [ ] PR renames all `GraphQL` config related things to `Web`
  - As it relates to the web server, we have a _GraphQL endpoint_, but it is not a GraphQL server
- [ ] Adds support for a web endpoint that supports raw SQL queries at `/api/sql/:namespace/:identifier`
  - Also validates that the queries are read only queries
    - Might have to add a bit more validation later

### Testing steps

- [ ] Rebuild WASM
- [ ] Refresh DB
- [ ] Start service with the new `--accept-sql-queries` flag
```bash
cargo run --bin fuel-indexer -- run --run-migrations --accept-sql-queries --manifest examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --replace-indexer
```
- [ ] Query the endpoint for data
```bash
curl -X POST http://localhost:29987/api/sql/fuel/explorer -d '{"query":"select json_agg(t) from (select count(*) from fuel_explorer.block) t"}' -H 'Content-type: application/json'
```
- [ ] Should return results `{"data":[[{"count":5289}]]}`
- [ ] Try submitting a mutation/write query 
```bash
curl -X POST http://localhost:29987/api/sql/fuel/explorer -d '{"query":"drop table postgres"}' -H 'Content-type: application/json'
```
  - Again, I'm sure the validation will get more complex in the future
- [ ] Should get `OperationNotSupported` error `{"details":"Error: Operation is not supported.","success":"false"}`
- [ ] Dockerized examples should work (with new args)
  - `docker build -t fuel-indexer/local:latest -f deployment/Dockerfile .`

### Changelog

- enhancement: add endpoint for raw sql queries


packages/fuel-indexer-tests/tests/integration/web_api_postgres.rs
deekerno, line 1: Is it possible to add a test for ensuring that only read-only queries are allowed?

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -377,7 +390,9 @@ pub(crate) async fn verify_signature(
     unreachable!();
 }
 
-pub async fn gql_playground(
+/// Given a `GraphQLRequest` for a given indexer's namespace and identifier, return the results
+/// of the GraphQL query.
```
deekerno, line 394: Can we change this to the following?:
> /// Return the source for a GraphQL playground page for display in a user's browser.

It looks like the current comment was copied from `query_graph(...)`.

packages/fuel-indexer-api-server/src/sql.rs
deekerno, line 1: 👏🏽

deekerno (commented):
Looks great, just left a couple comments.

deekerno (dismissed):

deekerno (approved):

PR #1113, opened 2023-07-15
chore: reduce CI run time for releases

deekerno:
### Description

This PR makes a number of changes to the CI pipeline:
- `cargo build`, `cargo clippy`, and `cargo test` will not run on merges to master, release publishes, or on version bump PRs
-- We disallow merging to `develop` unless these steps successfully pass, so we are essentially running redundant tasks whenever we release a new version.
- Publishing should now happen in parallel.
-- The [GitHub action that we use to publish crates](https://github.com/katyo/publish-crates) states that it determines the correct order for publishing according to dependencies.

### Testing steps

CI should pass. 

### Notes

I kept our default indexer checks for the release as it seems like a good idea to ensure those pass before a version is completely released to the public. Also, I adjusted the `publish-docker-image` step to only run on releases instead of both releases and merges to master.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,8 @@ jobs:
     needs:
       - cargo-toml-fmt-check
       - set-env-vars
-    if: needs.set-env-vars.outputs.IS_RELEASE != 'true'
+      - check-is-semver-branch
+    if: ${{ (needs.set-env-vars.outputs.IS_RELEASE_OR_MASTER != 'true') && (needs.check-is-semver-branch.outputs.is_semver_branch != 'true' ) }}
     strategy:
       matrix:
         command:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -104,7 +104,8 @@ jobs:
     needs:
       - cargo-toml-fmt-check
       - set-env-vars
-    if: needs.set-env-vars.outputs.IS_RELEASE != 'true'
+      - check-is-semver-branch
+    if: ${{ (needs.set-env-vars.outputs.IS_RELEASE_OR_MASTER != 'true') && (needs.check-is-semver-branch.outputs.is_semver_branch != 'true' ) }}
```
ra0x3, line 108: @deekerno Looks fine. Is there any way to make this a single check instead of an `&&` check? Would help with readability. Some type of `IS_RELEASE_OR_MASTER_OR_SEMVER_BRANCH` ?

deekerno: @ra0x3 I tried a few different ways, but it looks like we can't do the regex check in the `env` step to create a `IS_RELEASE_OR_MASTER_OR_SEMVER_BRANCH` env var. 👎🏽

ra0x3 (approved):
utACK. Let's remember to be around for the next release just in case something goes awry

Issue #1111, opened 2023-07-14
Figure out why indexers are dying implicitly


ra0x3:
- We have an issue where an indexer will just die, no errors, no warnings, nothing - it just stops
- I created https://github.com/FuelLabs/fuel-indexer/issues/1100 as a temporary fix for this until we can get the time to see what's actually happening under the hood here
- Repro steps:
  - Just deploy a default indexer to the playground, then look at [the Grafana dashboard](https://graphana-beta-1.fuel.network/d/oMe0biyVk/fuel-indexer-playground?orgId=1) 
  - You will see that indexers run for about an hour before they just stop
- It was suggested that we use [tokio-console](https://github.com/tokio-rs/console) to help debug the issue 

Additional context
- High priority on this cause obviously, even with https://github.com/FuelLabs/fuel-indexer/issues/1100 we need to find the root of the issue here and fix it ASAP
- https://github.com/FuelLabs/fuel-indexer/issues/1100 also contains more notes and details/links

ra0x3:
@deekerno Could you paste anything you found on this?

deekerno:
Here are some links that may be helpful to read:
- https://ryhl.io/blog/async-what-is-blocking/ - written by one of the Tokio devs
- https://users.rust-lang.org/t/tokio-from-async-to-sync-and-back-to-async-block-on-vs-spawn-blocking/83438 - same Tokio dev responds to some questions

Some thoughts:
- [`tokio-console`](https://github.com/tokio-rs/console) would be pretty helpful here, even if only to verify that the freeze is due to a task that infinitely blocking. Note that the implementer will have to add `console-subscriber` to the dependencies but also disable the use of `tracing-subscriber` in the codebase, as I found out the hard way that you cannot have use both at the same time. This would only be done to use the tool and not be a permanent change, but it is a bit of a pain to do.
- One can use something like [`lockbud`](https://github.com/BurtonQin/lockbud/tree/all/Code) to see if it can detect any locking bugs due to the use of the `async_std::sync::Mutex`; however, it only supports `std::sync::Mutex` and other similar primitives and one would have to downgrade their Rust version to a older nightly to be able to use it.
- After looking through the code for the `Executor` trait, I wonder if we even need to use `spawn_blocking(...)` at all. Presumably, it's ultimately because of calls to `ff_put_object` and `ff_get_object` in the `Entity` trait which use a blocking call to place data in the database.
[...]

PR #1116, opened 2023-07-17
fix: dispatch per block instead of per transaction

deekerno:
### Description

The indexer currently dispatches a handler function _per transaction_ in a block; in effect, we're repeating an indexing function up to the max amount of transactions per block, if the only parameter is `BlockData`.

This PR ensures that we index _per block_ by moving the `decoder.dispatch()` call to the block iteration loop rather than the transaction iteration loop.

### Testing steps

CI should pass.

#### Reproduction

You can reproduce the bug by doing the following on the `develop` branch:
- Ensure DB is cleared.
- Add an logging call to an indexer
-- Feel free to add `info!("Block #{}", block_data.header.height);` to the `fuel-explorer` handler code.
- Compile the WASM module.
- Run the service with your adjusted indexer:
-- `cargo run -p fuel-indexer -- run --run-migrations --manifest examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml --fuel-node-host beta-3.fuel.network --fuel-node-port 80`
- After about 30 seconds of running the indexer, stop it and look at the logs. You should see that some of the blocks are logged multiple times.
-- In my testing, this was apparent around block 1000.

For further proof, feel free to include the `--verbose` flag and check the database operations when a block is logged multiple times; you should see that the same database query is being executed the same amount of times.

#### Manual Testing

Run the same steps with this branch. You should see that there are no repeated logs or database queries for the same block.

### Notes

This may be related to #1111 as the amount of transactions and receipts dramatically increases in the 200K block range which means that the amount of `get_object` and `put_object` calls increases as well.

### DIFF
--- a/packages/fuel-indexer-macros/src/indexer.rs
+++ b/packages/fuel-indexer-macros/src/indexer.rs
@@ -576,9 +576,8 @@ fn process_fn_items(
                             }
                         }
                     }
-
-                    decoder.dispatch()#awaitness;
                 }
+                decoder.dispatch()#awaitness;
 
                 let metadata = IndexMetadataEntity{ id: block.height as u64, time: block.time as u64, block_height: block.height };
                 metadata.save()#awaitness;


ra0x3 (approved):
🏅 Wow, great (and hilarious) find. Needs a patch release

PR #1125, opened 2023-07-18
enhancement: welcome wizard

deekerno:
### Description

This PR revamps `forc index welcome` to ask a few questions and create a manifest accordingly.

#### Example 1
[asciicast](https://asciinema.org/a/G4ttExKJ3kXWSb3UOfxAOMInl)

#### Example 2
[asciicast](https://asciinema.org/a/hOioaqSsxOv0LbIHA2lFTPTsF)

### Testing steps
1. `cargo run -p forc-index -- welcome`
2. Read the prompts and answer the questions.
3. Ensure that the resultant project builds correctly.

### Notes
I'm down to change the name to something else. Additionally, perhaps we can augment `forc index new` with this functionality; for example, running `forc index new` with no args runs this series of prompts while including args would be considered "power user mode" and would require the usual args that are passed.


deekerno:
> @deekerno 
> 
> - I think it looks great! I like the personal tone of the prompts. 
> 
> - For namespace I'd change the prompt to `Name of your organization` or something like that (since namespace is the same thing that you use for Apple/Java builds - it's like an organization identifier).

Will do. I can also keep it as "Namespace" and change the help message to something like "name of your organization", if that also works.

> - I noticed that the default manifest doesn't seem to be produced from `forc index new` ? I didn't see any comments in the manifest

Right, as I couldn't populate the necessary options using `forc index new`. However, I can add the necessary bits to `NewCommand` and pass it through that flow instead; it should be a fairly easy change. 

> If we can get this second 50%, I think we'll have something that helps us by being super educational, therefore we have to answer less questions 

Heard that. I'll put this back in draft and add those bits to make it an end-to-end thing instead.

PR #1129, opened 2023-07-19
enhancement: add micro-benchmarking

deekerno:
### Description

This PR adds micro benchmarks for WASM.

### Testing steps

- Build the WASM modules.
- `cd /packages/fuel-indexer-benchmarks`
- `cargo bench`

### Changelog
- Breaks block retrieval functionality in `executor.rs` into its own function for use in the benchmarking suite
- Add Criterion benchmark utilities for generic WASM executor
- Add specific benchmarks for `fuel-explorer` indexer
- Adds a small benchmark for building GraphQL queries

### Notes

After rebasing on `develop`, I decided to remove the original WASM executor creation function as it was superseded by the more generic work that I did for benchmarking `fuel-explorer`. I would like to find a way to show the amount of points used by an executor and show that as part of the micro-benchmarks, but I couldn't find a clean way of doing it with `criterion` and didn't want this to sit around much longer. In any case, we can add to this as we see fit.


packages/fuel-indexer-benchmarks/benches/wasm/examples/fuel_explorer.rs
```diff
@@ -0,0 +1,144 @@
...
+
+pub(crate) fn fuel_explorer_10_blocks_start_200000(c: &mut Criterion) {
+    let f =
+        create_wasm_indexer_benchmark(200000, 10, "fuel_explorer_10_blocks_start_200000");
+    f(c, setup_fuel_explorer_manifest(), IndexerConfig::default())
+}
+
+pub(crate) fn fuel_explorer_100_blocks_start_200000(c: &mut Criterion) {
+    let f = create_wasm_indexer_benchmark(
+        200000,
```
ra0x3, line 62: Mentioned this to Maciej as well, but anyway to remove these magic numbers? 🥲

ra0x3 (commented):
@deekerno 
- Looks mostly fine, thanks for this
- What's the end result of this ?
  - Is this something we include on CI?
  - Is this a report that we can save and compare over time?

packages/fuel-indexer-benchmarks/benches/wasm/examples/fuel_explorer.rs
deekerno: Unfortunately not. 😢 `criterion` requires functions to be named it seems; my first pass at this actually tried to use a closure to generate all of the benchmarking functions from command-line args, but I couldn't get it to work. I'm sure there's _some_ way; if you'd rather me investigate a final time before approving it, I wouldn't mind.

packages/fuel-indexer/src/executor.rs
```diff
@@ -428,6 +209,248 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
     }
 }
 
+pub async fn retrieve_blocks_from_node(
```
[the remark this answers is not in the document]
deekerno: Good point. 👍🏽

packages/fuel-indexer-benchmarks/benches/wasm/metering.rs
```diff
@@ -0,0 +1,7 @@
+// use criterion::Criterion;
```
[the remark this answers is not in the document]
deekerno: I'm down to remove it.

packages/fuel-indexer-benchmarks/benches/wasm/examples/fuel_explorer.rs
ra0x3: Let me clarify. We can't change this:

```rust
pub(crate) fn fuel_explorer_100_blocks_start_50000(c: &mut Criterion) {
    let f =
        create_wasm_indexer_benchmark(50000, 100, "fuel_explorer_100_blocks_start_50000");
    f(c, setup_fuel_explorer_manifest(), IndexerConfig::default())
}
```

To this:

```rust
pub(crate) fn fuel_explorer_100_blocks_start_50000(c: &mut Criterion) {
    let first_named_thing = 50000;
    let second_named_thing = 100;
    let f =
        create_wasm_indexer_benchmark(first_named_thing, second_named_thing, "fuel_explorer_100_blocks_start_50000");
    f(c, setup_fuel_explorer_manifest(), IndexerConfig::default())
}
```

Just trying to minimize "what do these numbers mean" as much as possible.

deekerno:
>     * What's the end result of this?
>       * Is this something we include on CI?
>       * Is this a report that we can save and compare over time?

The end result is an easy way for us to compare changes across branches. The way that I've been using this locally is running `cargo bench` on `develop`, then switching to whatever branch I'm working on, and running `cargo bench` again. `criterion` saves the last run's information in `target/criterion` and will compare the current run with the last one before showing the results to the user.

As far as CI goes, it's recommended that we _do not_ include this type of time-based benchmarking in a CI pipeline because providers can do all sorts of pausing, job-switching, etc. and that makes the timing unreliable at best and useless at worst. However, there are things like `lai` that measure instruction counts which would be unaffected by CI stuff.

ra0x3 (approved):

PR #1138, opened 2023-07-24
chore: switch to manual sqlx query logging

ra0x3:

### Description

- This will allow us to finally actually see queries in the logs 🤞🏽 
- We _will not_ be able to see queries that aren't `put_object` and `get_query`
- Since `queries` are stateless, if we wanted to add this verbosity to each query we'd either need to create some `QueryManager` or something else that can hold the state of the config

### Testing steps

- [ ] Run the explorer on beta-3 in verbose mode and see the queries

```bash
cargo run --bin fuel-indexer -- run --run-migrations --accept-sql-queries --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --auth-enabled --auth-strategy jwt --jwt-issuer FuelLabs --verbose
```

### Changelog

- chore: switch to manual sqlx query logging


deekerno (approved):
LGTM!

PR #1140, opened 2023-07-24
enhancement: implement many-to-many relationships

ra0x3:

### Description

- This PR adds support for the actual implementation of many-to-many relationships
- After this PR is merged, many-to-many relationships defined in GraphQL schema, will now have data persisted into the DB

### Approach

- I implemented this primarily by adding the metadata related to the many-to-many relationship, to the `Entity` trait
- Since many-to-many inserts don't require the full `FtColumn::Object`, (because the fields of a many-to-many are _always_ just two `ID` fields), we can just pass a String query to the DB, as opposed to serialized `FtColumns`
  - Just FYI we _could_ alternatively still pass serialized FtColumns around - but again, without needing `FtColumn::Object`, this is kinda more work for nothing
- So when `save_many_to_many` is called, we just look at the `JOIN_METADATA` on the `Entity` and build the query based on each piece of metadata

### Testing steps

- [ ] Tests were updated so CI should pass
- [ ] Set your explorer `start_block` to something good like `725000`
- [ ] Rebuild WASM `bash scripts/utils/build_test_wasm_module.bash`
- [ ] Refresh DB `bash scripts/utils/refresh_test_db.bash`
- [ ] Try running the explorer on beta-3
```bash
cargo run --bin fuel-indexer -- run --run-migrations --accept-sql-queries --manifest examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml --fuel-node-host beta-3.fuel.network --fuel-node-port 80 --replace-indexer --auth-enabled --auth-strategy jwt --jwt-issuer FuelLabs
```
- [ ] Let the explorer run for a while
- [ ] Now try to inspect some many-to-many relationships in the DB

Join table
```sql
postgres=# select * from fuel_explorer.blocks_transactionidfragments limit 5;
      block_id       | transactionidfragment_id
---------------------+--------------------------
 4134974272678539573 |      4049969933482275382
 4134974272678539573 |      7305795610431598898
 4134974272678539573 |      3846973899826934116
 4134974272678539573 |      7377288952369524789
 4134974272678539573 |      7378131371496596577
```

Another join table
```sql
postgres=# select * from fuel_explorer.transactions_inputs limit 5;
    transaction_id    |       input_id
----------------------+----------------------
 12448931662499174308 |  7439866743899315930
  3089334239409520063 |  9335741207491190050
 18202162423980436043 | 18183823380237920257
  1409094539828266049 | 11861295837450341937
 16801737969603378815 | 16273975443049326803
(5 rows)
```

One more...

```sql
postgres=# select * from fuel_explorer.transactions_outputs limit 5 ;
    transaction_id    |      output_id
----------------------+---------------------
 12448931662499174308 | 5911275356368460277
 12448931662499174308 |  606830226975614421
  3089334239409520063 | 2608708248133720063
  3089334239409520063 |  606830226975614421
 18202162423980436043 |   90196747415819419
(5 rows)
```

### Changelog

- enhancement: implement many-to-many relationships


packages/fuel-indexer-schema/src/join.rs
```diff
@@ -0,0 +1,110 @@
...
+        &self.0
+    }
+
+    /// Whether or not there are actual records to insert for this query.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Create a new `RawQuery` from the given metadata and columns.
+    pub fn from_metadata(metadata: &JoinMetadata<'_>, columns: &[FtColumn]) -> Self {
```
deekerno, line 50: I think we should make it clear that this is for `JoinMetadata` or rename `RawQuery` to something that makes it known that this is pretty much only for the parent-child join table stuff (as far as I can tell).

deekerno (approved):

PR #1146, opened 2023-07-26
Update --replace-indexer and add --remove-data

### LINKED ISSUE #1127: Update `forc index deploy` with option to not nuke old data
- If you run the service with `--replace-indexer`, everything related to the previous indexer with the same `namespace` and `identifier` gets deleted
- This is cool but makes debugging hard
  - Old blocks aren't as important as new blocks
  - If we wanna see how the indexer really performs, we want to use it on new blocks
  - However, with `--replace-indexer` you can't redeploy indexers without always having to start at the beginning
- We should add some flag `--keep-data` to `forc index deploy` that gets sent to the web endpoint
  - This flag would indicate that the user _does not_ want to delete data associated with their index
- This allows for redeploying resumable indexers

lostman:
### Description

Fixes #1127.

This PR updates how `--replace-indexer` option works.

1. `--replace-indexer` has to be enabled in `fuel-indexer`.
2. `--replace-indexer` has to be specified in `forc index deploy`.
3. `--remove-data` has been added to `forc index deploy`.

`forc index deploy --replace-indexer --remove data` removes the old indexer and all indexed data and replaces it with the new indexer.

`forc index deploy --replace-indexer` replaces the old indexer but keeps the indexed data.

### Testing steps

Start the `fuel-indexer`:

```
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80
```

Deploy and redeploy an indexer using various combinations:

```
cargo run --release -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer --target-dir `pwd` --replace-indexer --remove-data
```

(when running `forc index deploy` this way the `--release` flag may be required or tokio runtime error may appear).

Change the schema and attempt to deploy. It should result in:
```
▸▹▹▹▹ 🚀 Deploying...
{
  "details": "Conflict: Indexer(fuel.explorer)'s schema has changed. Use --remove-data to replace the indexer.",
  "success": "false"
}
```

### Changelog

* When replacing an indexer, we wait until the old indexer stops. This is necessary to prevent race conditions—when the indexer has been requested to stop but is still in the middle of execution and may write to the database. If we didn't do this, the old data tables may be deleted, new tables created, and then the old indexer writes some data to the new tables.
* Add a notification mechanism that is triggered then the indexer stops execution
* Add an option to be notified when `StopRequest` completes. This is required for orderly completion of indexer replacement.

### DIFF
--- a/packages/fuel-indexer-api-server/src/uses.rs
+++ b/packages/fuel-indexer-api-server/src/uses.rs
@@ -375,7 +375,7 @@ pub(crate) async fn register_indexer_assets(
                                     .to_string();
                                 GraphQLSchema::new(content)
                             };
-                            // If re are replacing an indexer and keeping its
+                            // If we are replacing an indexer and keeping its
                             // data, its schema already exists.
                             if replace_indexer && !remove_data {
                                 assets.push(result);


packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -272,28 +325,34 @@ pub(crate) async fn register_indexer_assets(
...
-                                    assets.push(result);
-                                }
-                                Err(e) => {
-                                    let _res =
-                                        queries::revert_transaction(&mut conn).await?;
-                                    return Err(e.into());
+                            // If re are replacing an indexer and keeping its
+                            // data, its schema already exists.
+                            if replace_indexer && !remove_data {
+                                assets.push(result);
```
deekerno, line 333: I may be understanding this wrong, but if we're not replacing the schema, can we get away with not doing a DB call at all since the schema is already present in the database?

packages/fuel-indexer/src/executor.rs
```diff
@@ -126,6 +127,14 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
         let mut num_empty_block_reqs = 0;
 
         loop {
+            if kill_switch.load(Ordering::SeqCst) {
+                info!("Kill switch flipped, stopping Indexer({indexer_uid}). <('.')>");
+                if kill_confirm.send(()).is_err() {
+                    error!("Unable to notifty listeners that Indexer({indexer_uid}) has stopped.");
+                };
```
deekerno, line 134: In the changelog, it's stated that we wait until the old indexer stops before replacing the indexer. Does this actually cause the service to wait until it's done? It appears that this would just send a message on the oneshot channel after the logging before exiting the executor loop; wouldn't it still be possible for the confirmation to have been sent but the WASM execution to continue to save data to the database?

packages/fuel-indexer/src/executor.rs
```diff
@@ -76,6 +76,7 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
     manifest: &Manifest,
     mut executor: T,
     kill_switch: Arc<AtomicBool>,
+    kill_confirm: futures::channel::oneshot::Sender<()>,
```
deekerno, line 79: Should this be named `kill_confirm_trigger`?

deekerno (changes requested):
Left a few questions.

packages/fuel-indexer/src/service.rs
```diff
@@ -42,6 +42,10 @@ pub struct IndexerService {
 
     /// Killers used to stop the spawned indexers.
     killers: HashMap<String, Arc<AtomicBool>>,
+
+    /// For notifications that an indexer stopped execution after kill was
+    /// requested.
+    kill_confirms: HashMap<String, futures::channel::oneshot::Receiver<()>>,
```
ra0x3, line 48: +1 for the slightly more clarified name of `kill_confirm_triggers` (or something similar)

packages/fuel-indexer/src/service.rs
```diff
@@ -88,6 +93,7 @@ impl IndexerService {
                 &mut conn,
                 manifest.namespace(),
                 manifest.identifier(),
+                false,
```
ra0x3, line 96: Magic value `false` ? What does it mean?

ra0x3 (changes requested):
Nice work @lostman 

- Looks pretty fine overall.
- Left some feedback
- I think this might require tests for:
  - `--replace-indexer` disabled (I think we might already have this)
  - `--replace-indexer` enabled with `--replace`
  - `--replace-indexer` enabled with `--replace --remove-data`

packages/fuel-indexer-api-server/src/uses.rs
ra0x3: @lostman 
- If someone is uploading without `--remove-data` we need to verify that the schema in the DB (that we aren't changing) matches the schema that's being uploaded
  - Else we will have mismatched schemas on our hands

deekerno: > @deekerno Where is the DB call here?

Actually, I'm pretty sure that I originally misread the code in this section, so my original comment is moot.

deekerno (dismissed):

ra0x3 (changes requested):
@lostman Tiny comments this time. I'm testing this now. I agree that we should be able to get this merged today

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -211,41 +214,115 @@ pub(crate) async fn register_indexer_assets(
...
+
+        // Check that the schema has not changed.
+        for (asset_type, data) in asset_bytes.iter() {
+            if *asset_type == IndexerAssetType::Schema {
+                // If the indexer already exists
+                if let Ok(indexer_id) = indexer_id {
+                    // The schema must be the same. This query returns an asset
+                    // if the bytes match. If it returns None (and the indexer
+                    // exists), it means that its schema is different.
+                    if let None = queries::asset_already_exists(
```
ra0x3, line 261: @lostman Everything seems to work but this.

Repro steps:
- `forc index new`
- `forc index deploy`
- Add a new `type` to your schema
- Now `forc index deploy` again -- should fail due to mismatched schemas since you didn't include `--remove-data`
- For me, it seems to just pick up from where the last indexer started

ra0x3 (commented):
@lostman Only one thing not working during my manual tests

ra0x3 (approved):
@lostman Big win!

PR #1147, opened 2023-07-26
enhancement: make `NODE_BLOCK_PAGE_SIZE` configurable

### LINKED ISSUE #1128: Make `NODE_GRAPHQL_PAGE_SIZE` configurable from the service
- `NODE_GRAPHQL_PAGE_SIZE` determines how many blocks an indexer will try to fetch from the `fuel-graqphql-client`
- As we're running experiments on the explorer, we see that different page sizes have different tradeoffs
  - E.g., indexed blocks per second speed tradeoff vs making more calls to `.save()` (more blocks == more saves)
- Let's make this `NODE_GRAPHQL_PAGE_SIZE` option configurable from the service side so if we want to try different values, we can just update the start command (rather than having to update source code)
- This should also be added to the `config.yaml` manifest

deekerno:
Closes #1128.

### Description

- Allow for configuration of `NODE_BLOCK_PAGE_SIZE`
- Add flag to documentation
- Change playground `NODE_BLOCK_PAGE_SIZE` value to 25

### Testing steps

CI should pass.

#### Manual Testing

Run the explorer example without the `--node-block-page size` flag; the service should request 10 blocks at a time. Run the explorer example again, but this time with `--node-block-page-size [NUM]`; the service should request `NUM` blocks at at time.

### DIFF
--- a/deployment/charts/fuel-indexer/templates/fuel-indexer-deployment.yaml
+++ b/deployment/charts/fuel-indexer/templates/fuel-indexer-deployment.yaml
@@ -59,7 +59,7 @@ spec:
             - "--verbose"
             - "--replace-indexer"
             - "--accept-sql-queries"
-            - "--node-block-page-size"
+            - "--block-page-size"
             - "25"
           env:
           - name: POSTGRES_PASSWORD
--- a/deployment/charts/fuel-indexer/templates/fuel-indexer-deployment.yaml
+++ b/deployment/charts/fuel-indexer/templates/fuel-indexer-deployment.yaml
@@ -59,6 +59,8 @@ spec:
             - "--verbose"
             - "--replace-indexer"
             - "--accept-sql-queries"
+            - "--node-block-page-size"
+            - "25"
           env:
           - name: POSTGRES_PASSWORD
             valueFrom:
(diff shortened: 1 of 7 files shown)


deployment/charts/fuel-indexer/templates/fuel-indexer-deployment.yaml
```diff
@@ -59,6 +59,8 @@ spec:
             - "--verbose"
             - "--replace-indexer"
             - "--accept-sql-queries"
+            - "--node-block-page-size"
```
ra0x3, line 62:
```suggestion
            - "--block-page-size"
```

packages/fuel-indexer-lib/src/defaults.rs
```diff
@@ -91,8 +91,8 @@ pub const VERBOSE_LOGGING: bool = false;
 /// Verbose output for database operations.
 pub const VERBOSE_DB_LOGGING: &str = "false";
 
-/// Amount of blocks to return in a GraphQL page.
-pub const NODE_GRAPHQL_PAGE_SIZE: usize = 10;
+/// Amount of blocks to return in a request to a Fuel node.
+pub const NODE_BLOCK_PAGE_SIZE: usize = 10;
```
ra0x3, line 95: To be clear `NODE_BLOCK_PAGE_SIZE` is fine for internal, but for external maybe we can shorten to `--block-page-size`

ra0x3 (approved):
🥇

PR #1196, opened 2023-08-04
fix: ensure we always get schema version the same way

### LINKED ISSUE #1143: Enusure schema version in WASM lib and schema version when uploading, match
- I'm not sure exactly what this issue might be but just leaving a note
- We encountered this issue when migrating the playground from prod to dev
- Apparently there was an indexer (shoutbox) that had a different version in its WASM lib, than it had in the DB
 - Meaning `VERSION` in the WASM lib, was different from `schema_version` in `graph_registry_type_ids`
- How did this happen? Not sure yet
- However, we should add some validation when registering these indexers to ensure this doesn't happen and return a 400 or something if it happens

lostman:

### Description

Closes #1143.

When creating `WasmIndexExecutor`, we no longer get the schema version through the FFI from the WASM file. Instead, we always get it from the `Manifest` through the `GraphQLSchema` struct.

In addition, when `WasmIndexExecutor` is created, we check that the WASM module version matches the schema file's version.

### Testing steps

CI tests are sufficient.

`build-and-test-examples` already caught an error in an earlier version:
https://github.com/FuelLabs/fuel-indexer/actions/runs/5809617283/job/15748759298

### Changelog

* `WasmIndexExecutor::new` now takes a `schema_version` `String`.
* Check that the WASM version matches the schema version.

### DIFF
--- a/packages/fuel-indexer-api-server/src/uses.rs
+++ b/packages/fuel-indexer-api-server/src/uses.rs
@@ -261,9 +261,7 @@ pub(crate) async fn register_indexer_assets(
                             }
                         }
                         IndexerAssetType::Schema => {
-                            let schema = GraphQLSchema::new(
-                                String::from_utf8_lossy(&data).to_string(),
-                            );
+                            let schema = GraphQLSchema::from(data.to_vec());
 
                             match queries::register_indexer_asset(
                                 &mut conn,
--- a/packages/fuel-indexer-lib/src/graphql/mod.rs
+++ b/packages/fuel-indexer-lib/src/graphql/mod.rs
@@ -104,6 +104,12 @@ impl From<&GraphQLSchema> for Vec<u8> {
     }
 }
 
+impl From<Vec<u8>> for GraphQLSchema {
+    fn from(value: Vec<u8>) -> Self {
+        GraphQLSchema::new(String::from_utf8_lossy(&value).to_string())
+    }
+}
+
 impl std::fmt::Display for GraphQLSchema {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", self.schema)


packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -261,20 +261,21 @@ pub(crate) async fn register_indexer_assets(
...
+                            let schema = GraphQLSchema::new(
+                                String::from_utf8_lossy(&data).to_string(),
+                            );
+
                             match queries::register_indexer_asset(
                                 &mut conn,
                                 &namespace,
                                 &identifier,
-                                data.to_vec(),
+                                schema.to_string().as_bytes().to_vec(),
```
ra0x3: Can we make this a bit more idiomatic?

```rust
impl From<GraphQLSchema> for Vec<u8> {
  fn from(schema: GraphQLSchema) -> Self {
    // ...
  }
}
```

then 

```rust
&mut conn,
&namespace,
&identifier,
schema.into()
```

packages/fuel-indexer-lib/src/graphql/mod.rs
```diff
@@ -56,7 +56,11 @@ type IndexMetadataEntity @entity {
 
 /// Inject native entities into the GraphQL schema.
 fn inject_native_entities_into_schema(schema: &str) -> String {
-    format!("{}{}", schema, IndexMetadata::schema_fragment())
+    if !schema.contains("type IndexMetadataEntity") {
```
ra0x3, line 59: 👌🏽

packages/fuel-indexer/src/executor.rs
```diff
@@ -631,12 +632,18 @@ impl WasmIndexExecutor {
             return Err(IndexerError::MissingHandler);
         }
 
-        // FunctionEnvMut and SotreMut must be scoped because they can't be used
-        // across await
-        let version = {
+        // FunctionEnvMut and SotreMut must be scoped with {} because they can't
```
ra0x3:
```suggestion
        // FunctionEnvMut and StoreMut must be scoped because they can't
```
The term "scoped" implies `{}`

packages/fuel-indexer/src/executor.rs
```diff
@@ -631,12 +632,18 @@ impl WasmIndexExecutor {
...
+        // be used across await
+        {
             let mut env_mut = env.clone().into_mut(&mut store);
             let (data_mut, mut store_mut) = env_mut.data_and_store_mut();
 
+            let schema_version_from_wasm = ffi::get_version(&mut store_mut, &instance)?;
+
+            if schema_version_from_wasm != schema_version {
+                return Err(IndexerError::Unknown(format!("Schema version from WASM {schema_version_from_wasm} does not match schema version {schema_version}")));
```
ra0x3, line 644: 👌🏽

ra0x3 (changes requested):
@lostman Looks OK. Small pieces of feedback

ra0x3:
@lostman Also, this PR does not include any testing steps

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -261,20 +261,21 @@ pub(crate) async fn register_indexer_assets(
                             }
                         }
                         IndexerAssetType::Schema => {
+                            let schema = GraphQLSchema::new(
```
ra0x3: Sorry I missed this prior @lostman but again here

```rust
let schema = GraphQLSchema::from(&data)
```

With:

```rust
impl From<Vec<u8>> for GraphQLSchema 
```

ra0x3 (approved):

Issue #1191, opened 2023-08-03
Deploying indexer triggers `SIGTERM`


ra0x3:
### Repro test
- I cannot repro this locally
- But in order to repro this on the playground:
```bash
forc index new indexer-test --namespace fuellabs && cd indexer-test

forc index auth --url https://indexer-playground.swayswap.io

forc index deploy --auth $TOKEN --url https://indexer-playground.swayswap.io
```

### Logs
Logs after deploying indexer and waiting 5-10 mins
```text
023-08-03T17:47:13.088639Z  INFO fuel_indexer_postgres: 577: Registered Asset(Schema) with Version(2738d221cf1e926d28e62bc93604a96ec6f7c5093e766f45a4555ed06e437b7f) to Indexer(fuellabs.indexer_test).
2023-08-03T17:47:13.089355Z  INFO fuel_indexer_schema::db::manager: 35: Creating schema for Indexer(fuellabs.indexer_test) with Version(2738d221cf1e926d28e62bc93604a96ec6f7c5093e766f45a4555ed06e437b7f).
2023-08-03T17:47:14.188731Z  INFO fuel_indexer_postgres: 577: Registered Asset(Wasm) with Version(68c702b88c0a7b498601604062f87a4a503c7dfc020a0605797ed3377c2112cf) to Indexer(fuellabs.indexer_test).
2023-08-03T17:47:14.288097Z  INFO tower_http::trace::on_response: 254: finished processing request latency=1537253 μs status=200
2023-08-03T17:47:14.489948Z  INFO fuel_indexer::service: 404: Resuming Indexer(fuellabs.indexer_test) from block 1
2023-08-03T17:48:39.888000Z  INFO fuel_indexer::commands::run: 37: Received SIGTERM. Stopping services.
```
### Details
- The indexer never starts, the service completely freezes and is rendered unusable
- I suspect the culprit is https://github.com/FuelLabs/fuel-indexer/pull/1146
  - This is the only change that was made to indexer deployments
- Not even the healthcheck works, the service is brick'd 🧱 
- You can verify healthchecks stop [here](https://grafana.swayswap.io/d/oMe0biyVk/fuel-indexer-playground?orgId=1)

ra0x3:
Fixed in https://github.com/FuelLabs/fuel-indexer/pull/1199

PR #1197, opened 2023-08-04
fix: spawn a tokio::task for kill_confirm.await to fix indexer SIGTERM

lostman:
### Description

Spawn a separate `tokio::task` to await indexer kill confirmation and notify the listener.

This may have something to do with https://github.com/FuelLabs/fuel-indexer/issues/1191


ra0x3:
@lostman I think we can close this for now :)

PR #1202, opened 2023-08-04
enhancement/breaking: update directive usage

ra0x3:

### Checklist

### Description

- Make our directive usage more idiomatic GraphQL (similar to how TheGraph does things)

### Testing steps

- [ ] CI should pass
- [ ] Docs should pass

### Changelog

- enhancement/breaking: update directive usage


packages/fuel-indexer-lib/src/graphql/parser.rs
```diff
@@ -344,6 +344,17 @@ impl ParsedGraphQLSchema {
                         TypeKind::Object(o) => {
                             let obj_name = t.node.name.to_string();
 
+                            // Only parsed `TypeDefinition`s with the `@entity` directive.
```
ra0x3, line 347: This is the only actual change. The rest of the diff is just docs/tests.

deekerno (changes requested):
Are all entities required to have the `@entity` directive? Was this to simplify virtual entity marking?

In any case, I think that we should also add documentation for `@entity`.

PR #1211, opened 2023-08-07
Trigger rebuild when schema changes.

### LINKED ISSUE #1175: Trigger re-build when schema changes.
There is a common pitfall when working on an indexer: modifying only the schema does not trigger a rebuild of the WASM binary.

Example:

1. deploy an indexer:
```
cargo run -p forc-index -- deploy --path examples/hello-world/hello-indexer/ --target-dir `pwd` --replace-indexer --remove-data
```

2. modify the schema:
```
echo "type Ty { id: ID! }" >> examples/hello-world/hello-indexer/schema/hello_indexer.schema.graphql
```

3. redeploy:
```
cargo run -p forc-index -- deploy --path examples/hello-world/hello-indexer/ --target-dir `pwd` --replace-indexer --remove-data
```

This triggers an error loop on `fuel-indexer`:

```
2023-08-02T09:21:59.865028Z ERROR fuel_indexer::database: 143: TypeId(-8293274664658733968) not found in tables: {}.

Does the schema version in SchemaManager::new_schema match the schema version in Database::load_schema?

Do your WASM modules need to be rebuilt?
```

The common advice is to modify `lib.rs` after changing the schema. This is a bad developer experience.

We should find a way to trigger rebuild when only the schema changes.

#### Option 1: `include_str!`: include the schema file contents in the WASM binary

`include_str!(file)` adds the `file` as a dependency. Thus, if we include this in the output of the `indexer` macro: 

```
    let mut output = quote! {
        #namespace_tokens
        #identifer_tokens
        #version_tokens
        static SCHEMA: &str = include_str!(#path_str);
    };
```

then the indexer will be rebuilt when the schema changes.

#### Option 2: always `touch src/lib.rs`

We always (unless `--skip-build` is set) modify the `atime` and `mtime` of `src/lib.rs`. This will always trigger a rebuild, even if the schema, ' lib.rs`, or any dependency has not changed.

I think removing the schema pitfall is worth it.

lostman:
### Description

Closes #1175.

When the schema is modified, rebuild the WASM module.

### Testing steps

Create and deploy an indexer:

```
forc index new --namespace test test-indexer
cd test-indexer
forc index deploy
```

Change the schema and deploy again:

```
echo "type Type { id: ID! }" >> schema/test_indexer.schema.graphql
forc index deploy
```

Without this PR, the second `forc index deploy` would not rebuild the WASM module, and the deployed indexer would cause an error.

### Changelog

* add ensure_rebuild_if_schema_changed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -333,7 +333,7 @@ jobs:
   publish-docker-image:
     needs:
       - set-env-vars
-    if: needs.set-env-vars.outputs.IS_RELEASE == 'true' || needs.set-env-vars.outputs.IS_MASTER == 'true'
+    if: needs.set-env-vars.outputs.IS_RELEASE == 'true'
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/packages/fuel-indexer/src/service.rs
+++ b/packages/fuel-indexer/src/service.rs
@@ -381,6 +381,10 @@ async fn get_start_block(
             info!("{action} Indexer({}) from block {block}", manifest.uid());
             Ok(block)
         }
-        None => Ok(manifest.start_block().unwrap_or(1)),
+        None => {
+            let block = manifest.start_block().unwrap_or(1);
+            info!("Starting Indexer({}) from block {block}", manifest.uid());
+            Ok(block)
+        }
     }
 }


ra0x3:
@lostman 
- Commit title does not conform to the CONTRIBUTING guidelines
- Also your testing steps mention `forc index` which is the published plugin, when that should be `forc-index`
- Testing steps should also mention that you have to rebuild the plugin locally to test

ra0x3 (approved):
Works, thanks @lostman this has been a PITA for a long time. Glad to see it solved.

ra0x3:
Let's defer to @deekerno before merging

plugins/forc-index/src/ops/forc_index_build.rs
```diff
@@ -64,6 +64,8 @@ pub fn init(command: BuildCommand) -> anyhow::Result<()> {
         );
     }
 
+    let path = path.unwrap_or(".".into());
```
deekerno: Can we use `std::env::current_dir()` instead of `".".into()` as it's platform-agnostic?

plugins/forc-index/src/ops/forc_index_deploy.rs
```diff
@@ -44,30 +44,25 @@ pub async fn init(command: DeployCommand) -> anyhow::Result<()> {
...
-            .to_string(),
-    );
+    let manifest_schema_file = Path::new(&workspace_root)
+        .join(manifest.graphql_schema())
+        .to_str()
+        .unwrap()
+        .to_string();
 
-    manifest.set_module(target_dir.join(manifest.module()).into());
+    let manifest_modulue_file = workspace_root.join(manifest.module());
```
deekerno: s/`manifest_modulue_file`/`manifest_module_file`/g

deekerno (commented):
Looks pretty good to me! Left just two nits.

lostman:
@deekerno fixed the nits.

deekerno (approved):

Issue #1194, opened 2023-08-03
Optimize and condense E2E tests


deekerno:
Our E2E tests have a lot of duplicated logic; we should condense and optimize them in order to speed up CI time and dev feedback loops.

ra0x3:
Minor thing I recently noticed but maybe `http_client` can be bundled into `setup_*_test_components` - it's used very simlarly and would save us a few lines 🙆🏽‍♂️

Issue #1145, opened 2023-07-25
Integrate web server, GraphQL layer, and data layer


ra0x3:
- We ideally want to keep GraphQL usage on the web API
  - We're using GraphQL syntax to define our data models (regardless of the outcome of this issue) so it would be nice to stay consistent, and other projects at Fuel use GraphQL
- In order to avoid https://github.com/FuelLabs/fuel-indexer/issues/1066, we need to figure out some way to integrate our web server (axum), our GraphQL usage (async-graphql), and our data layer (sqlx)
- The Graph solved for this by just hand-rolling _everything_
  - https://github.com/graphprotocol/graph-node/blob/master/graphql
- We will most likely have to do the same
- Although we maybe won't be able to use their `graph-graphql` crate, we could most definitely take inspiration from how they're doing things

### Tasks
- [x] https://github.com/FuelLabs/fuel-indexer/issues/1325
- [ ] ~~https://github.com/FuelLabs/fuel-indexer/issues/1310~~
- [ ] https://github.com/FuelLabs/fuel-indexer/issues/931
- [ ] https://github.com/FuelLabs/fuel-indexer/issues/960
- [ ] https://github.com/FuelLabs/fuel-indexer/issues/858

ra0x3:
Steps
- [ ] https://github.com/FuelLabs/fuel-indexer/issues/846

Will add more steps as we find them

ra0x3:
@deekerno Just an FYI
- Keep in mind the other two smaller issues in this larger issue ☝🏼 
- I closed them out but we should still integrate them into whatever work you do.
- They are:
  - Usage of `camelCase` field definition names in the schema (e.g., `account_id` becomes `accountId`)
  - Using `snake_case` table names for type definitions that are `PascalCased` (e.g., `indexmetadataentity` becomes `index_metadata_entity`

PR #1217, opened 2023-08-07
refactor: swap ID from u64 to `SizedAsciiString<64>`

ra0x3:

### Checklist

### Description
- I noticed quite a bit of weirdness whilst working on this - so we won't be cutting 0.20 immediately until we add some more (relevant, non-redundant, useful) testing 
- PR does a few things
  - Does not optimize usage of `.clone()` https://github.com/FuelLabs/fuel-indexer/issues/1278
  - Refactors the project to use `SizedAsciiString<64>` IDs instead of `u64` IDs
    -  With `SizedAsciiString<64>` aliased as `UID`
  - Removes simple-wasm indexer
    - It wasn't testing anything we weren't already testing
  - Also includes more work from https://github.com/FuelLabs/fuel-indexer/issues/1194
    - We don't use any other DB anymore so removed "postgres" suffix from test modules 
    - Moves the fuel-indexer-test indexer from fuel-indexer-tests/components/indices/ to fuel-indexer-tests/indexers
      - simple-wasm too
    - Removes redundant tests (database) that have not detected much breakage in the past
    - Removes the graphql_schema test module
      - Our GraphQL layer is gonna get an entire makeover per https://github.com/FuelLabs/fuel-indexer/issues/1145
        - Grantees are using the SQL endpoint
   - Updates our trybuild tests (which have not been running on CI :/ )
     - Re-enables these on CI
   - Moves `IndexMetadataEntity` to `fuel_indexer_types`
     - And makes `fuel_indexer_lib` dependent on `fuel_indexer_types`
       - `fuel_indexer_types` is now the base crate (previously was `fuel_indexer_lib`)

### Testing steps

- As this was a very low-level/granular change, CI for everything should pass
- Start your indexer service on beta-4 `cargo run --bin fuel-indexer -- run --run-migrations --fuel-node-host beta-4.fuel.network --fuel-node-port 80`
- Deploy the default indexer `forc index new indexer-test --namespace fuellabs && cd indexer-test && forc index deploy`
- Let this index to around block 250K, you should have no warnings for constraint errors

### Changelog

- refactor: swap ID from u64 to `SizedAsciiString<64>`


packages/fuel-indexer/src/executor.rs
```diff
@@ -687,7 +688,13 @@ impl WasmIndexExecutor {
             let schema_version_from_wasm = ffi::get_version(&mut store_mut, &instance)?;
 
             if schema_version_from_wasm != schema_version {
-                return Err(IndexerError::Unknown(format!("Schema version from WASM {schema_version_from_wasm} does not match schema version {schema_version}")));
+                // Just an FYI this error below is not getting unwrapped into a panic...
```
ra0x3, line 691: FYI @lostman

packages/fuel-indexer-types/src/lib.rs
```diff
@@ -21,102 +24,20 @@ pub trait TypeId {
...
     pub use crate::receipt::*;
     pub use crate::scalar::*;
-    pub use crate::{TypeId, FUEL_TYPES_NAMESPACE};
+    pub use crate::{
+        type_id, Bech32Address, Bech32ContractId, Bits256, Identity, SizedAsciiString,
+        TypeId, FUEL_TYPES_NAMESPACE,
+    };
 }
 
-#[cfg(test)]
```
ra0x3, line 29: Not even sure we ever used these conversions

packages/fuel-indexer-tests/tests/graphql_server.rs
```diff
@@ -256,24 +259,6 @@ async fn test_filtering() {
 
     mock_request("/ping").await;
 
-    // ID selection
```
ra0x3, line 259: - This part of the test was giving problems
- Something tells me it's related to passing the `ID` in here manually where `ID` was simply 1 - however, now, ID is now calculated differently . 
- The fact that everything else works kinda lets me know its this specific test

deekerno: Can we create an issue to ensure that we re-add an ID selection test at some point?

packages/fuel-indexer-plugin/src/wasm.rs
```diff
@@ -75,15 +75,15 @@ pub trait Entity<'a>: Sized + PartialEq + Eq + std::fmt::Debug {
         }
     }
 
-    fn load(id: u64) -> Option<Self> {
+    fn load(id: UID) -> Option<Self> {
         unsafe {
-            let buf = id.to_le_bytes();
-            let mut buflen = (buf.len() as u32).to_le_bytes();
+            let buff = bincode::serialize(&id.to_string()).unwrap();
```
deekerno, line 80: I _think_ we can remove this `.to_string()` call here as `SizedAsciiString` provides an `impl` for `Serialize`: https://github.com/FuelLabs/fuels-rs/blob/df21541b679bb90a1ee3e54f34a67b7037b40be7/packages/fuels-core/src/types/core/sized_ascii_string.rs#L162-L169.

Additionally, can we replace the `.unwrap()` with an `.expect()`?

packages/fuel-indexer/src/ffi.rs
```diff
@@ -258,22 +284,30 @@ impl<'a> WasmArg<'a> {
         })
     }
 
+    pub fn store(&mut self) -> &mut Store {
```
deekerno, line 288: Could we add a doc comment for this new function?

deekerno (changes requested):
Running through your testing steps now; left a few questions/suggestions in the meantime.

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -316,11 +316,16 @@ pub(crate) async fn register_indexer_assets(
...
+        if let Err(e) = tx
+            .send(ServiceRequest::Reload(ReloadRequest {
+                namespace,
+                identifier,
+            }))
+            .await
+        {
+            error!("Failed to send ServiceRequest::Reload: {e:?}");
+            return Err(e.into());
+        }
```
lostman, line 328: I wonder whether there is a better way of handling this pattern.

The error returned by `?` is generally logged somewhere up. We have a couple of error messages that repeat this way.

The advantage of the `error!` macro is the source location attribution.

Perhaps there is some way to have both concise code and precise location.

packages/fuel-indexer-tests/indexers/fuel-indexer-test/fuel_indexer_test.yaml
```diff
@@ -1,6 +1,6 @@
 namespace: fuel_indexer_test
 fuel_client: ~
-graphql_schema: packages/fuel-indexer-tests/components/indices/fuel-indexer-test/schema/fuel_indexer_test.graphql
+graphql_schema: packages/fuel-indexer-tests/indexers/fuel-indexer-test/schema/fuel_indexer_test.graphql
```
lostman, line 3: `indexers` 👍

deekerno (approved):

packages/fuel-indexer-api-server/src/uses.rs
ra0x3: @lostman 
- I agree it seems a bit...repetitive? 
- I think the main reason I bubbled things up in this way (and in executor/service) was to be almost annoyingly explicit with the path of the error (where we can trace the error's path just via looking at the logs)
  - I'm open to better ways to accomplish this

PR #1225, opened 2023-08-08
enhancement: Update --replace-indexer and add --remove-data V2

lostman:
### Description

A redo of https://github.com/FuelLabs/fuel-indexer/pull/1146 but with the kill notification mechanism removed.

#### This PR adds an automatic CI test:

https://github.com/FuelLabs/fuel-indexer/pull/1225/files#diff-61710fa718a25e0e49156f4ac2682a97e9eb0d9fb5cb5316f09f924dde896982R307

#### Manual testing:

1. Start the `fuel-indexer`:

```
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-3.fuel.network --fuel-node-port 80
```

2. Deploy an indexer:

```
cargo run --release -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer
```

Expected result: indexer processing blocks.

3. Redeploy an indexer, and keep the data:

```
cargo run --release -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer --replace-indexer
```

Expected result: indexer is replaced, and it resumes from where `fuel-explorer` left off:

```
INFO fuel_indexer::service: 397: Resuming Indexer(fuellabs.explorer) from block 1760
```

4. Redeploy an indexer, and remove the data:

```
cargo run --release -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer --replace-indexer --remove-data
```

Expected result: the indexer restarts from block 1:

```
INFO fuel_indexer::service: 397: Resuming Indexer(fuellabs.explorer) from block 1
```

5. Modify the schema, and attempt to redeploy the indexer without removing the data:

```
echo "type Ty { id: ID! }" >> examples/fuel-explorer/fuel-explorer/schema/fuel_explorer.schema.graphql 
```

```
chmod +x ./scripts/utils/build_modules.bash && ./scripts/utils/build_modules.bash
```

```
cargo run --release -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer --replace-indexer
```

Expected result: an error:

```
▹▹▹▹ 🚀 Deploying...
{
  "details": "Conflict: Indexer(fuel.explorer)'s schema has changed. Use --remove-data to replace the indexer.",
  "success": "false"
}
```

6. Redeploy the modified indexer and remove old data:

```
cargo run --release -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer --replace-indexer
```

Expected result: indexer restarts from block 1.

```
INFO fuel_indexer::service: 397: Resuming Indexer(fuellabs.explorer) from block 1
```

### Changelog
- Instead of the kill notification mechanism and removing/restarting the indexer in a single request, in this PR, `forc index deploy` first issues the `remove` command and then deploys the indexer as a separate command.
- The indexer commits no data when the kill switch is flipped on—it reverts the transaction. This is done in `handle_events`, which now takes `Arc<AtomicBool>`. This allows the indexer to shutdown gracefully and prevents any data from entering the DB after the tables have been removed and recreated for the replacement indexer.

Thoughts?


packages/fuel-indexer-tests/tests/indexing_postgres.rs
```diff
@@ -595,3 +598,103 @@ async fn test_index_sway_enums() {
     assert_eq!(row.get::<&str, usize>(1), EXPECTED_CONTRACT_ID);
     assert_eq!(row.get::<BigDecimal, usize>(2).to_u64().unwrap(), 0);
 }
+
+#[actix_web::test]
+async fn test_replacing_an_indexer_and_keeping_or_removing_data() {
```
ra0x3: @lostman 
- This is a web server test (related to uploading/deploying indexers), not a test related to indexing data (triggering events, indexing the data)
- So we should move this to web_api_postgres.rs (I think that's the name of the module)

ra0x3:
Sorry for the conflicts @lostman. I'm reviewing this now. Aiming to merge tomorrow 👌🏽

lostman:
@ra0x3, no worries. I'll fix the conflicts in the meantime.

ra0x3:
@lostman 
- The testing steps (in both this PR and the linked PR) are unclear.
- You've added two options in this PR
- What is the expected result after running each option?

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
-            "assets": assets,
-        })));
+    Ok(assets)
+}
+
+// This function parses the `Multipart` struct set to the deploy indexer
+// endpoint. It extracts the `bool` value indicating whether to replace an
+// indexer if it already exists, and extracts the indexer assets: manifest,
+// schema, and the WASM module.
+async fn parse_register_indexer_multipart(
```
ra0x3, line 369: 👌🏽

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
+    while let Ok(Some(field)) = multipart.next_field().await {
+        let name = field.name().unwrap_or("").to_string();
+        let data = field.bytes().await.unwrap_or_default();
+        match name.as_str() {
+            "replace_indexer" => {
+                replace_indexer = std::str::from_utf8(&data.to_owned())
+                    .map_err(|e| ApiError::OtherError(e.to_string()))?
+                    .parse::<bool>()
+                    .map_err(|e| ApiError::OtherError(e.to_string()))?;
+                continue;
```
ra0x3: Do we need this `continue` ?

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
-                                    return Err(e.into());
-                                }
-                            }
-                        }
-                    }
-                }
-                Err(e) => {
-                    let _res = queries::revert_transaction(&mut conn).await?;
-                    return Err(e.into());
+#[allow(clippy::too_many_arguments)]
```
ra0x3, line 262: - Let's remove this `#[allow(clippy::too_many_arguments)]` and adjust the code so that we don't need it 
- `namespace`, `identifier`, and `pubkey` are all a part of `RegisteredIndexer`
- `register_indexer_assets_transaction` fetches the `indexer_id` but you could adjust this to fetch the `RegisteredIndexer` using `namespace` and `identifier` via `get_indexer`

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
+
+    // If the indexer already exists, check that the schema has not changed.
+    if let Ok(indexer_id) = indexer_id {
+        for (asset_type, data) in asset_bytes.iter() {
+            if *asset_type == IndexerAssetType::Schema {
+                // The schema must be the same. This query returns an asset
+                // if the bytes match. If it returns None (and the indexer
+                // exists), it means that its schema is different.
+                let schema = GraphQLSchema::from(data.to_vec());
+                if queries::asset_already_exists(
```
ra0x3, line 292: - Not really a comment for this PR per-se
- But I (just now) confirmed that `asset_already_exists`'s `digest` resolves to the same hash as `GraphQLSchema.version()` (even though we're using we're using slightly different means of calculating the hash, which we should update/fix in https://github.com/FuelLabs/fuel-indexer/issues/1227

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
+                    conn,
+                    &IndexerAssetType::Schema,
+                    &(&schema).into(),
+                    &indexer_id,
+                )
+                .await?
+                .is_none()
+                {
+                    return Err(ApiError::Http(HttpError::Conflict(format!(
```
ra0x3, line 301: 👌🏽

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
+                            "Indexer({namespace}.{identifier})'s schema has changed. Use --replace-indexer --remove-data to replace the indexer and the indexed data."
+                        ))));
                 }
             }
         }
+    }
 
-        queries::commit_transaction(&mut conn).await?;
+    if !config.replace_indexer && replace_indexer {
+        error!("Failed to replace Indexer({namespace}.{identifier}): replacing an indexer is not enabled.");
```
ra0x3, line 310: 👌🏽

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -208,132 +212,181 @@ pub(crate) async fn register_indexer_assets(
...
+                    conn,
+                    namespace,
+                    identifier,
+                    (&schema).into(),
+                    IndexerAssetType::Schema,
+                    Some(pubkey),
+                )
+                .await?;
+
+                let schema =
```
ra0x3: Don't we already have the `GraphQLSchema` on line 327?

packages/fuel-indexer-tests/tests/web_server.rs
```diff
@@ -296,3 +302,103 @@ async fn test_querying_sql_endpoint_when_sql_is_enabled_returns_error_for_non_su
 
     server.abort();
 }
+
+#[actix_web::test]
+async fn test_replacing_an_indexer_and_keeping_or_removing_data() {
```
ra0x3, line 307: Nice clean test 👌🏽

packages/fuel-indexer/src/database.rs
```diff
@@ -235,10 +235,7 @@ Do your WASM modules need to be rebuilt?
                 .entry(column.type_id)
                 .or_insert_with(|| table.to_string());
 
-            let columns = self
-                .schema
-                .entry(table.to_string())
-                .or_insert_with(Vec::new);
+            let columns = self.schema.entry(table.to_string()).or_default();
```
ra0x3, line 234: 👌🏽

plugins/forc-index/src/ops/forc_index_deploy.rs
```diff
@@ -41,6 +43,19 @@ pub async fn init(command: DeployCommand) -> anyhow::Result<()> {
         })?;
     }
 
+    // If we are replacing an indexer but not removing data, there is no need to
+    // issuer remove command. Ordinary reload is enough.
```
ra0x3:
```suggestion
    // issue a remove command. Ordinary reload is enough.
```

plugins/forc-index/src/ops/forc_index_deploy.rs
```diff
@@ -41,6 +43,19 @@ pub async fn init(command: DeployCommand) -> anyhow::Result<()> {
...
+    if replace_indexer && remove_data {
```
ra0x3, line 48: - This here is the reason I don't think we _really_ need the complexity of a `kill_switch` in `handle_events`.
- A three second wait here would be more than enough time for the tokio task to stop 
  - Just now testing on my machine on origin/develop - the explorer never takes more than 2 seconds to stop after calling `forc index remove`
  - Remember `forc index deploy` is literally a deployment (it posts assets, registers them in the DB, creates a new task, starts that task -- it does a lot)
    - An additional small delay will not disrupt any UX here (especially if it means less complexity elsewhere)

ra0x3 (changes requested):
@lostman 
- Looks mostly ok thanks for getting back to this
- The only part I'm skeptical about is the change to `handle_events`
  - There are not details in the description (or anywhere else) that prove how passing a `kill_switch` to `handle_events` helps here
    - IMO, what you added on line 48 of `forc_index_deploy.rs` removes the need for this complexity 
- This work produced an issue in production last time, so that's the reason the review is more strict this time

plugins/forc-index/src/ops/forc_index_deploy.rs
lostman: @ra0x3, It is possible for the remove and the deploy commands to execute, thus cleaning up the database and recreating the tables while the indexer is doing something. The indexer then finishes and commits the transaction, putting the data into newly created tables.

Why should we allow this possibility at all?

The `kill_switch` is already there in `run_executor.` All we need to do is to pass it to `handle_events` and abort the transaction. This gives us a clean and reliable behavior: no data is written to the DB after the kill switch has been flipped.

Currently, it can go either way—the indexer may write some data and then stop, or it may stop without writing any more data.

We could **also** add **early indexer termination**: since the kill switch is just an `AtomicBool,` we could check it each time `put_object` or `get_object` are called and terminate the (WASM) indexer then and there. That would be only slightly more complex but would have the benefit of clear and reliable behavior.

ra0x3: @lostman 👌🏽

packages/fuel-indexer-api-server/src/uses.rs
lostman: `get_indexer` still needs `namespace` and `identifier`, so we only save `pubkey` as an argument.

This would avoid the clippy warning, but it isn't worth doing (this is a rather dubious warning).

We should use the `pubkey` we have in the API request, instead of fetching it from the DB. They should be the same, but still—it feels off.

Moreover, `get_indexer` returns a `Result<Option>`, so the unwrapping code gets more complicated. All of this is minor, but definitely not worth doing to avoid a clippy warning.

ra0x3 (changes requested):
It works! 🏅 

Notes after manual testing

- Testing steps Fuel host needs to be updated to beta-4.fuel.network
- I noticed that I don't need to `--replace-indexer` when I `forc index deploy` in order for the indexer to be replaced
  - This PR should require that I include `--replace-indexer` if my intention is to replace the indexer
    - If I `forc index deploy` an existing indexer without `--replace-indexer` I should receive a 409 conflict error
    - **This is my only feedback from manual testing**
- When using `forc index deploy` with `--replace-indexer --remove-data`, my indexer is replaced and the data is removed 👍🏽 
  - However I do notice a lot of these logs
  - Log spam isn't the end of the world but might be a UX issue down the road (no biggie for now)

```
2023-08-29T14:02:35.598665Z ERROR fuel_indexer::database: 175: Failed to put_object: Database(PgDatabaseError { severity: Error, code: "25P02", message: "current transaction is aborted, commands ignored until end of transaction block", detail: None, hint: None, position: None, where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("postgres.c"), line: Some(1438), routine: Some("exec_parse_message") })
```

lostman:
@ra0x3, my latest commit addresses the missing check for `--replace-indexer` flag from the CLI.

As for the log spam, once @deekerno adds return codes, I plan to use the kill switch to do proper early termination in WASM. This way, we won't be getting any of these errors.

ra0x3 (approved):

ra0x3:
Given how long this has been outstanding, gonna go ahead and merge it so that it doesn't hang around any longer. Thanks @lostman

PR #1234, opened 2023-08-09
fix: clean up WasmArg immediately after call to WASM

lostman:
### Description

This PR moves the `WasmArg` cleanup immediately after the call to the WASM module.

It occurred to me that `handle_events` might return an error without cleanup–hence skipping dealloc.

This may be related to the memory leak.

### Testing steps

CI testing.

### Changelog

- move the `WasmArg` cleanup


lostman:
Closing in favor of https://github.com/FuelLabs/fuel-indexer/pull/1245

PR #1236, opened 2023-08-09
fix: update + explain executor logic

ra0x3:

### Checklist

### Description

- This PR clears a lot of the unnecessary stuff (that I added) to the executor, and also adds comments to explain the functionality, since there is a lot of business logic here
- PR also swaps the naming of `cursor` and `next_cursor` as they were confusing/reversed previously 
- We _always_ bump the cursor now, no more using `cursor.is_none()` to decide if we should move forward
  - [See comment here from Green](https://fuellabs.slack.com/archives/C01L1HUJSL8/p1691592348026359?thread_ts=1691591690.062229&cid=C01L1HUJSL8)
- I'm trying to figure out why indexers in our QA environment keep sleeping -- even though they're only on block 80,000 🤔 

### Testing steps

- [ ] The only actual logic change this PR implements is removing all the logic related to constraints/DB errors and retrying
- [ ] Everything CI/tests/examples/service should work as is.

### Changelog

- fix: update + explain executor logic

### DIFF
--- a/packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
+++ b/packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
@@ -364,11 +364,8 @@ async fn test_can_index_metadata_when_indexer_macro_is_called_postgres() {
             .await
             .unwrap();
 
-    assert_eq!(
-        row.get::<BigDecimal, usize>(0).to_u64().unwrap(),
-        11557898405596845040
-    );
-    assert_eq!(row.get::<BigDecimal, usize>(1).to_u64().unwrap(), 0);
+    assert!(row.get::<BigDecimal, usize>(1).to_u64().unwrap() > 0);
+    assert_eq!(row.get::<BigDecimal, usize>(2).to_u64().unwrap(), 1);
 }
 
 #[actix_web::test]
--- a/packages/fuel-indexer/src/executor.rs
+++ b/packages/fuel-indexer/src/executor.rs
@@ -1,3 +1,4 @@
+/// Abstractions for indexer task execution.
 use crate::{
     database::Database, ffi, queries::ClientExt, IndexerConfig, IndexerError,
     IndexerResult,
@@ -30,21 +31,24 @@ use std::{
     str::FromStr,
     sync::atomic::{AtomicBool, Ordering},
 };
-use thiserror::Error;
 use tokio::{
     task::{spawn_blocking, JoinHandle},
     time::{sleep, Duration},
 };
 use tracing::{debug, error, info, warn};
 use wasmer::{
-    imports, CompilerConfig, Cranelift, FunctionEnv, Instance, Memory, Module,
-    RuntimeError, Store, TypedFunction,
+    imports, CompilerConfig, Cranelift, FunctionEnv, Instance, Memory, Module, Store,
+    TypedFunction,
 };
 use wasmer_middlewares::metering::MeteringPoints;
 
+/// Source of the indexer's execution.
 #[derive(Debug, Clone)]
 pub enum ExecutorSource {
+    /// The executor was created from a manifest file.
     Manifest,
+
+    /// The executor was created from indexer bytes stored in the database.
     Registry(Vec<u8>),
 }
 
@@ -57,9 +61,9 @@ impl AsRef<[u8]> for ExecutorSource {
     }
 }
 
-impl ExecutorSource {
-    pub fn to_vec(self) -> Vec<u8> {
-        match self {
+impl From<ExecutorSource> for Vec<u8> {
+    fn from(source: ExecutorSource) -> Self {
+        match source {
             ExecutorSource::Manifest => vec![],
             ExecutorSource::Registry(bytes) => bytes,
         }
@@ -479,6 +483,9 @@ pub async fn retrieve_blocks_from_node(
     Ok((block_info, cursor, has_next_page))
 }
 
+/// Executors are responsible for the actual indexing of data.
+///
+/// Executors can either be WASM modules or native Rust functions.
 #[async_trait]
 pub trait Executor
 where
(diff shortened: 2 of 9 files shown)


packages/fuel-indexer-tests/tests/e2e/indexing_postgres.rs
```diff
@@ -364,11 +364,8 @@ async fn test_can_index_metadata_when_indexer_macro_is_called_postgres() {
             .await
             .unwrap();
 
-    assert_eq!(
```
ra0x3, line 367: `IndexMetadataEntity` is now created via `::new()` so the ID is no longer deterministic (since it includes `time`)

packages/fuel-indexer/src/service.rs
```diff
@@ -134,7 +134,7 @@ impl IndexerService {
         .await?;
 
         let mut items = vec![
-            (IndexerAssetType::Wasm, exec_source.to_vec()),
+            (IndexerAssetType::Wasm, Vec::<u8>::from(exec_source)),
```
deekerno: Just a suggestion: Perhaps we can give `items` a type and then use `Into` on each of the bits as necessary?

```
let mut items: Vec<(IndexerAssetType, Vec<u8>)> = vec![
    (IndexerAssetType::Wasm, exec_source.into())),
    ...
];
```

packages/fuel-indexer/src/executor.rs
```diff
@@ -564,6 +602,7 @@ impl<F> Executor for NativeIndexExecutor<F>
 where
     F: Future<Output = IndexerResult<()>> + Send,
 {
+    /// Handle events for  native executor.
```
deekerno, line 605: Remove extra space here.

packages/fuel-indexer/src/executor.rs
```diff
@@ -501,13 +531,21 @@ unsafe impl<F: Future<Output = IndexerResult<()>> + Send> Send
 {
 }
 
+/// Native executors differ from WASM executors in that they are not sandboxed - they are
+/// merely a set of native Rust functions that directly from the process running the indexer
+/// service.
```
deekerno: Consider:
> Native executors differ from WASM executors in that they are not sandboxed; they are merely a set of native Rust functions that _(run/execute/are spawned)_ directly from the indexer service process.

Mainly, it seems there's a verb missing here.

packages/fuel-indexer/src/executor.rs
```diff
@@ -156,94 +190,84 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
...
 
 /// Retrieve blocks from a client node.
 ///
-// This was abstracted out of `run_executor` in order to allow for
-// use in the benchmarking suite to give consistent timings.
+/// This was abstracted out of `run_executor` in order to allow for use in the benchmarking suite
+/// to give consistent timings.
+///
+/// If there is an issue fetching blocks, we return an empty cursor and an empty list of blocks, then
+/// run_executor will determine whether or not we should sleep for a bit and try again.
```
deekerno, line 246: This can be an internal comment as the reasoning probably doesn't need to be included in the public rustdocs, just my two cents.

packages/fuel-indexer/src/executor.rs
```diff
@@ -79,75 +83,105 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
...
         let mut retry_count = 0;
 
         // If we're testing or running on CI, we don't want indexers to run forever. But in production
-        // let the index operators decide if they want to stop idle indexers. Maybe we can eventually
-        // make this MAX_EMPTY_BLOCK_REQUESTS value configurable
+        // let the indexer service operator decide if they want to stop idle indexers.
+        //
+        // Maybe we can eventually make this MAX_EMPTY_BLOCK_REQUESTS value configurable
+        //
+        // Also note that this count considers _consecutive_ empty block requests.
```
deekerno: Perhaps we can rename `retry_count` and `MAX_EMPTY_BLOCK_REQUESTS` to `consecutive_retries` and `MAX_CONSECUTIVE_EMPTY_BLOCK_RESPONSES`, respectively? It'll make it more self-documenting and it also communicates that we're logging empty block _responses_ to our requests.

deekerno (commented):
Looks good! Left some comments.

packages/fuel-indexer/src/service.rs
ra0x3: Me like this idea

packages/fuel-indexer/src/executor.rs
ra0x3: I don't believe there's any whitespace here? This is a doc comment for line 606

packages/fuel-indexer/src/executor.rs
ra0x3: @deekerno 
- So previously I've agreed that these should be internal comments.
- However, after looking over other projects (namely tokio and sqlx) I see they have a _ton_ of rustdoc comments
  - They have tons of rustdoc comments that we would definitely consider "internal comments"
  - Which makes sense, so devs reading on crates.io don't have to open the source code to see the maintainer's important comments
- So yes, in the past I've agreed that these were internal (and have made such changes to my PRs)
- However, now I'm gonna push back a bit ;) and say we can be liberal with the rustdoc comments (given that other high visibility projects do the same)

packages/fuel-indexer/src/executor.rs
deekerno: May be a bit of weirdness on my end then. 👍🏽

packages/fuel-indexer/src/executor.rs
deekerno: I'm always down to look like the popular projects. 😄

deekerno (dismissed):

lostman (dismissed):

deekerno (approved):

PR #1072, opened 2023-07-10
Update wasmer from v2.3.0 to v4.0.0

lostman:
### Description

Update `wasmer` from `v2.3.0` to `v4.0.0`.

### Testing steps

Please provide the _exact_ testing steps for the reviewer(s) if this PR requires testing.

This PR also changes how `load_schema` works and how it is called. It is now a private function called from `Database::new`. Previously it would fetch the namespace and identifier from the manifest for the native executor and through FFI from the WASM executor. The latter is unnecessary. We can use the namespace and identifier from the manifest in both cases. This cleans up our code a bit.

https://github.com/FuelLabs/fuel-indexer/pull/1072/commits/82bec0dbdb80f63105a583c783f41de8290c69c0

### Changelog

* update wasmer to v4.0.0
* clean up `Database` creation and initialization

PR #1245, opened 2023-08-10
enhancement: reimplement WasmArg Drop

lostman:
### Description

`Drop` for `WasmArg` was removed in #1072. It did not occur to me then to store `MutexGuard<'_, Store>` (instead of a full `Mutex`) in `WasmArg`.

Reimplementing `Drop` guarantees cleanup.

### Testing steps

CI testing.

### Changelog

* reimplement `Drop` for `WasmArg` to ensure cleanup.


ra0x3:
@lostman 
- The description is a bit unclear as to what this PR is and does, could you add a bit more context?
- Also, given that these changes are a bit of a hot path, it would be nice to see the before/after result of this PR to confirm that the changes work as expected

lostman:
@ra0x3, in #1072, I removed `Drop` implementation for `WasmArg` because at the time I did not see how do to it with the changes required for `wasmer` update.

So, the cleanup (dealloc) had to be done manually, and I put it here:
https://github.com/FuelLabs/fuel-indexer/pull/1245/files#diff-fa722c66d2896c684e5a26aa2c56b7ec1734888609fbb64ecbc7dfd6a420c4ffL849

However, this isn't the right place for it.

The call to WASM is here:
https://github.com/FuelLabs/fuel-indexer/pull/1245/files#diff-fa722c66d2896c684e5a26aa2c56b7ec1734888609fbb64ecbc7dfd6a420c4ffL830

And before the cleanup (dealloc), there is a possibility of the function returning with an error:
https://github.com/FuelLabs/fuel-indexer/pull/1245/files#diff-fa722c66d2896c684e5a26aa2c56b7ec1734888609fbb64ecbc7dfd6a420c4ffL835

Thus, cleanup wouldn't happen.

After I spotted the mistake, I opened a PR to fix it. And I have also realized that I could implement `Drop` after all.

This PR does that.

So,

1. a bug is fixed when cleanup might be skipped
2. `Drop` is reimplemented to restore the guarantee that cleanup always happens.

deekerno:
@lostman: If you don't mind, I'd like to do some memory profiling on my side before approving this. I imagine that this probably would probably lead to more reasonable memory allocation over time and may have been a contributing factor to our memory issues that we saw a few days ago.

lostman:
@deekerno, I don't mind. I recommend merging it sooner rather than later because it fixes a bug.

The type of errors that would trigger this is rare (e.g., WASM trap), so it shouldn't significantly contribute to the memory issues we've seen.

Running out of metering points would trigger it too, but with our generous default allocation, it shouldn't be happening either.

Still, if you find that it did contribute, please let me know 🙏

deekerno (approved):
Decided to do my memory profiling at a later time as it's not necessary to block this bug fix when I can just check out the commits on my own. 👌🏽

PR #1270, opened 2023-08-14
ci: optimize workflow for general development PRs

### LINKED ISSUE #1252: Optimize CI workflow for general development PRs
I have some thoughts for adjustments to general development PRs; generally, the CI workflow for a development PR would be as follows:

_First group_
- set-env-vars
- docs-tests
- Cargo.toml check
- `cargo fmt --check` on the workspace

_Second group_
- `cargo clippy --locked --no-default-features` on the workspace
- `cargo clippy --locked --all-targets --all-features` on the workspace
- Unit tests
- E2E and integration tests
-- This would include the optimized tests from #1194 and any new tests added as part of #1230.
- Build default indexer
- Build and test examples

This would speed up development PRs considerably as we'd be cutting about 24 jobs from our workflow by not building each package with both sets of feature flags.

Reasoning:
- [`cargo clippy` runs `cargo check`](https://github.com/rust-lang/rust-clippy/blob/c154754b74577906c5d55d57f7daeff02d6a33e7/src/main.rs#L59), which is [`cargo build` without the actual code generation process](https://doc.rust-lang.org/cargo/commands/cargo-check.html#description). I don't think we've had an error that was build-time only in a very, very long time (please correct me if I'm wrong).
-- If we're skittish about not having a build job that actually catches errors in code generation, I think we can move them to the version bump PR type as we would still be able to catch any failed builds before the publishing step.
- The E2E and integration tests generally have to compile the project in order to properly test things from end-to-end so we should be able to catch any failed builds at this step.

@ra0x3, @lostman: Thoughts?

deekerno:
_Note: Pointing this at [deekerno/1194-e2e-test-optimization](https://github.com/FuelLabs/fuel-indexer/tree/deekerno/1194-e2e-test-optimization) until that's merged._

Closes #1252.

### Description

Optimizes the workflow for general development PRs:
- First stage runs the doc tests, `cargo-toml-fmt-check` and `cargo-fmt-check`
- Second stage runs `cargo clippy --locked --all-targets --all-features`, `cargo clippy --locked --no-default-features`, unit tests, E2E tests, default indexer build, and build/test of examples.

### Testing Steps

CI should pass.

### Notes
Old workflow:
image

New workflow:
image


ra0x3:
@deekerno Main thing I care about is CI time, but I don't see that result? How much faster (if at all) are these changes? Eye-balling our actions, it actually looks like this new check takes longer? Old checks were around 17m, this check around 25m ?

deekerno:
> @deekerno Main thing I care about is CI time, but I don't see that result? How much faster (if at all) are these changes? Eye-balling our actions, it actually looks like this new check takes longer? Old checks were around 17m, this check around 25m ?

@ra0x3 With no cache, they do seem to be about the same, which is surprising. I'm down to close this as we know that our current flow works with no unintended side effects.

deekerno:
Gonna close this for now as we really just want faster times.

PR #1275, opened 2023-08-16
fix: block height conversion

lostman:
### Description

`block_height` is now `u32`. Attempting to get `BigDecimal` results in a silent error, and `last_block_height_for_indexer` always returns `1`. This PR fixes it. 

### Testing steps

Run `forc-index`:
```
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-4.fuel.network --fuel-node-port 80 --replace-indexer --manifest examples/hello-world/hello-indexer/hello_indexer.manifest.yaml 
```
And restart:
```
cargo run -p fuel-indexer -- run --run-migrations --fuel-node-host beta-4.fuel.network --fuel-node-port 80
```

On `develop,` `hello_indexer` starts from `1`. With this fix, it resumes where it left off.

### Changelog

* fix `block_height` conversion: from `BigDecimal` to `i32`.

### DIFF
--- a/packages/fuel-indexer/src/service.rs
+++ b/packages/fuel-indexer/src/service.rs
@@ -357,7 +357,7 @@ async fn create_service_task(
 }
 
 /// Determine the starting block for this indexer.
-async fn get_start_block(
+pub async fn get_start_block(
     conn: &mut IndexerConnection,
     manifest: &Manifest,
 ) -> Result<u32, IndexerError> {
--- a/packages/fuel-indexer-database/postgres/src/lib.rs
+++ b/packages/fuel-indexer-database/postgres/src/lib.rs
@@ -6,7 +6,7 @@ use fuel_indexer_lib::utils::sha256_digest;
 use sqlx::{
     pool::PoolConnection,
     postgres::PgRow,
-    types::{BigDecimal, JsonValue},
+    types::JsonValue,
     Postgres, Row,
 };
 use std::str::FromStr;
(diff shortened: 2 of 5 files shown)


ra0x3:
@lostman 
- Thanks for finding this 🙏🏽 
- Just an FYI that this (and all PRs going forward) needs a test. 
  - cc @deekerno on this ☝🏽 
- Really tired of seeing the same functionality break over and over.

ra0x3:
Additionally, I think I also see another bug related to start block. Give me a few mins while I try to confirm

ra0x3 (approved):

packages/fuel-indexer/src/service.rs
```diff
@@ -357,7 +357,7 @@ async fn create_service_task(
 }
 
 /// Determine the starting block for this indexer.
-async fn get_start_block(
+pub async fn get_start_block(
```
deekerno, line 360: Does this need to be public? I know we're using it in the test, but if we're only using it in one place, I'd rather just use `queries::last_block_height_for_indexer` and assert the value.

deekerno (approved):
LGTM! Left one question.

packages/fuel-indexer/src/service.rs
ra0x3: @deekerno I saw you ✅ so I merged. We can address this later if you want (I agree we should only make things `pub` if necessary)

PR #1281, opened 2023-08-18
enhancement: refactor GraphQL schema parser

lostman:
### Description

This PR refactors the `ParsedGraphQLSchema::new()` function and removes unnecessary content from the struct.

All parsing logic is moved to a separate struct, `SchemaDecoder`.

`ServiceDocument` and raw `GraphQLSchema` are removed from `ParsedGraphQLSchema`. `ServiceDocument` was only used in the code generator, and we only used the type definitions. We don't need both since we already have these in the parsed content—in `type_defs`. As for `GraphQLSchema`, we only need the version, not the raw schema `String`.

### Testing steps

CI testing is sufficient.

### Changelog

* Move all parsing code out of `ParsedGraphQLSchema::new()` and into a separate struct.
* Remove `ast: ServiceDocument`
* Remove `schema: GraphQLSchema`
* Add `version: String` (the version from `GraphQLSchema`)


ra0x3:
@lostman 
- I haven't looked at any of the file changes since you haven't pinged this for review, but just a couple of notes/points
- We will most likely be caching this `ParsedGraphQLSchema` (inside of `IndexerSchema`) when web requests come in, so we have a vested interest in not cloning `TypeDefinition`s if we don't have to - and keeping this `ParsedGraphQLSchema` as small as possible
  - Obviously in the previous implementation there is _a lot_ of cloning of these `TypeDefinition`s
- We'll also want to be able to Serialize/Deserialize this ParsedGraphQLSchema

lostman:
@ra0x3, if we want to keep it as small as possible, we can remove the following:
```rust
ast: ServiceDocument
```
It is used in only one place:
```rust
    for definition in schema.ast().definitions.iter() {
        if let Some(def) = process_definition(&schema, definition) {
            output = quote! {
                #output
                #def
            };
        }
    }
```

And `process_definition` produces an output only for `Type`s

```rust
    match definition {
        TypeSystemDefinition::Type(def) => process_type_def(schema, &def.node),
        TypeSystemDefinition::Schema(_def) => None,
        def => {
            panic!("Unhandled definition type: {def:?}");
        }
    }
```

If we want to throw any errors, like the above panic in the case of a `Directive,` we can do that at the parser level.

We could instead refer to `type_defs` directly:

```rust
    for (_, type_definition) in schema.type_defs().iter() {
        if let Some(def) = process_type_def(&schema, type_definition) {
        }
    }
```

> Serialize/Deserialize

This poses a problem. The types we use from `async-graphql-parser` (e.g., `TypeDefinition`) are not serializable. We would have to write a serializer by hand.

Do we **really** need to serialize this struct? Loading the schema from the DB and parsing it shouldn't be a bottleneck. And for in-memory cache, we shouldn't need to serialize them.

One more thing, in addition to the `ServiceDocument`, we keep the raw schema `String` inside the `ParsedGraphQL` struct. It isn't used anywhere—only the version. So we should only keep the version and the parsed content. I did this in this PR, and I think it is the way to go, but I can always undo this change if needed.

ra0x3 (approved):
I think this does look cleaner. Will defer to @deekerno

deekerno (approved):
LGTM!

PR #1289, opened 2023-08-22
fix: move `client.chain_info()` call outside of loop

deekerno:
### Description

`client.chain_info()` is used to get the necessary chain information to identify a transaction. When doing the beta-4 upgrade, I didn't realize that I could have placed the method call outside of the loop. This ensures that the call only happens once instead of every transaction.

### Testing steps

CI should pass.

#### Manual testing

There should be a significant speed up in block retrieval and processing in comparison to `develop`.


lostman (approved):
Huge difference. 👍 👍 👍

PR #1290, opened 2023-08-22
ci: optimize workflow for general development PRs

### LINKED ISSUE #1252: Optimize CI workflow for general development PRs

deekerno:
Closes #1252.

### Description

Optimizes the workflow for general development PRs:
- First stage runs the doc tests, `cargo-toml-fmt-check` and `cargo-fmt-check`
- Second stage runs `cargo clippy --locked --all-targets --all-features`, `cargo clippy --locked --no-default-features`, unit tests, E2E tests, default indexer build, and build/test of examples.

### Testing Steps

CI should pass.

### Notes
Old workflow:
image

New workflow:
image

Compare the times between a fresh run of this PR (CI #4456) and other similar runs (CI #4454, CI #4453); there is a slight improvement in run time for un-cached runs -- https://github.com/FuelLabs/fuel-indexer/actions


ra0x3:
@deekerno There's a conflict but this is good to go once that's fixed ✅

ra0x3 (approved):

PR #1294, opened 2023-08-23
fix: update `MessageOut` data parsing

deekerno:
### Description

Due to [a change in how the `data` field in a `MessageOut` receipt is formed](https://github.com/FuelLabs/fuel-vm/commit/18a2e4a13094448068e378baf3b89e448ec356d6), our type ID and data parsing did not cover each case. This PR adjusts the parsing to check for a `None` value as well as an empty byte array as well.

### Testing steps

On `develop`: Clear database, build any indexer (`fuel-explorer` is probably best) and run indexer on internal beta-4 test network. You should see a failure at or near block 364788.

On this branch: Repeat above steps. You should see no failure.

### Notes

A change in an upstream dependency has led to the `getrandom` crate being included in our dependency graph for `fuel-indexer-utils`. Pinning `fuel-{crypto, types, vm, etc.}` to `v0.35.3` seems to have solved the issue for now until we can work with the other teams to rectify the issue. Also, the `forc-index-tests` CI job has been disabled until we put out a new patch as the public version will not have the pinned dependencies until a new release is published.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -270,7 +270,8 @@ jobs:
         run: cargo test --locked --all-targets --features trybuild -p
           fuel-indexer-tests
   forc-index-tests:
-    if: needs.set-env-vars.outputs.IS_RELEASE_OR_MASTER_OR_DEVELOP_OR_SEMVER != 'true'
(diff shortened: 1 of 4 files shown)


lostman (approved):
Confirming that it fixes the issue. 👍 .

PR #1297, opened 2023-08-25
enhancement: store BlockData in the DB and ensure the indexers don't miss blocks

lostman:
### Description

This PR adds `index_block_data` table in which we store serialized `BlockData` structs. We can use these to get a new indexer up to speed without fetching blocks from the client again.

The table contains a trigger that ensures we have a full range of blocks—no `block_height` may be missing.

Additionally, this PR adds the same trigger to each indexer's `indexmetadataentity` table. Again, this ensures that an indexer must process each block in sequence. If the indexer fails to process a block or processes blocks out of order, the trigger will raise an exception (and the indexer's transaction will fail).

### Note

The trigger ensures the indexer can't process blocks out of order or miss any blocks. However, it doesn't stop the indexer from running—it simply doesn't progress. To stop the indexer, we need the WASM error codes https://github.com/FuelLabs/fuel-indexer/pull/1337 and return an error (and trigger an early exit!) when `put_object` fails.

### Testing steps

### Changelog


packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -681,7 +711,7 @@ pub async fn last_block_height_for_indexer(
     Ok(row
         .try_get::<i32, usize>(0)
         .map(|id| id.to_u32().expect("Bad block height."))
-        .unwrap_or_else(|_e| 1))
+        .unwrap_or(0))
```
lostman, line 714: I think we need to return zero here, for consistency.

If an indexer processed blocks 1, 2, and 3, the result would be `3`.

If an indexer processed block 1, the result would be `1`.

If an indexer processed no blocks, we shouldn't return `1` (as we currently do) but `0`.

I made changes to `get_start_block` to account for this.

packages/fuel-indexer/src/service.rs
```diff
@@ -387,7 +387,8 @@ pub async fn get_start_block(
             .await?;
             let start = manifest.start_block().unwrap_or(last);
             let block = if *resumable {
-                std::cmp::max(start, last)
+                // if the last processed block is N, we want to resume from N+1
+                std::cmp::max(start, last + 1)
```
lostman, line 448: If the indexer processed no blocks and `last_block_height_for_indexer` returned 0, we bump it to 1—the first new block it needs to process.

Similarly, if the indexer processed blocks 1, 2, 3, and `last_block_height_for_indexer` returned 3, we bump it to 4—the first new block.

ra0x3:
@lostman Is this PR good to close?

lostman:
Deprecated in favor of https://github.com/FuelLabs/fuel-indexer/pull/1369

PR #1302, opened 2023-08-27
enhancement: Added message to compiler error when on Apple M1

### LINKED ISSUE #889: Inform User about possible missing environment vars
Sometimes a user is presented with this message, even after installing `llvm`.

```
error: unable to create target: 'No available targets are compatible with triple "wasm32-unknown-unknown"'
```
If they were to run:
`clang --version`
They might see:
```
Apple clang version 13.0.0 (clang-1300.0.29.30)
Target: arm64-apple-darwin21.3.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```
Whereas we want it to be:
```
Homebrew clang version 16.0.2
Target: arm64-apple-darwin21.3.0
Thread model: posix
InstalledDir: /opt/homebrew/opt/llvm/bin
```
The diff here being the `InstalledDir`, clang does not always link to llvm even after its installed. In this case what the user needs to do is add:
```
export LIBCLANG_PATH="/opt/homebrew/opt/llvm/lib"
export LDFLAGS="-L/opt/homebrew/opt/llvm/lib"
export CPPFLAGS="-I/opt/homebrew/opt/llvm/include"
```
This reminder *has* been added to our docs, but many users may miss this. It would be great if we could do the following:
1. Check whether a user is on an M1 system
2. If so, check for an updated clang binary

NOTE: these are all read operations, we don't want to write anything to the users system.
If the users `InstalledDir` does not point to llvm then we want to link them to our docs that mentions the possible missing vars. 

There could be a better way of presenting this info to them in the terminal, but this is the general gist.

dadepo:
Currently opening as draft:
- To get feedback on the wording of the error message.
- @ra0x3 You left a comment [here](https://github.com/FuelLabs/fuel-indexer/issues/889#issuecomment-1543790207) stating:

> I might even go as far to say we maybe should read the output of a failed forc index build and look for wasm-related output in order to present this message their as well

Not sure I totally understand the requirement here and whether the changes in this PR already satisfies this.

### Description

When compilation fails and host is Apple M1, print out error message Informing user about possible missing environment vars.

Closes #889

Currently the message printed is

```

▪▪▪▪▪ ❌ Build failed.
For Apple Silicon macOS users the preinstalled llvm has limited WASM targets,
please install a binary with better support from Homebrew (brew install llvm)
and configuring rustc with necessary environment variables:
            
export LIBCLANG_PATH='/opt/homebrew/opt/llvm/lib'
export LDFLAGS='-L/opt/homebrew/opt/llvm/lib'
export CPPFLAGS='-I/opt/homebrew/opt/llvm/include'                                                                                                                                   Error: ❌ Failed to build index.
```

### Changelog

- Added a private `error_message` function that, on compilation error, includes information about setting env variables if compilation fails on Apple M1s.

### DIFF
--- a/plugins/forc-index/src/ops/forc_index_build.rs
+++ b/plugins/forc-index/src/ops/forc_index_build.rs
@@ -3,12 +3,14 @@ use fuel_indexer_lib::manifest::{Manifest, Module};
 use indicatif::{ProgressBar, ProgressStyle};
 use serde::Deserialize;
 use std::{
+    env,
     fs::File,
     io::{Read, Write},
     path::Path,
     process::{Command, Stdio},
     time::Duration,
 };
+use std::collections::HashSet;
 use tracing::info;
 
 #[derive(Deserialize)]
@@ -165,18 +167,18 @@ pub fn init(command: BuildCommand) -> anyhow::Result<()> {
                         if s.success() {
                             pb.finish_with_message("✅ Build succeeded.");
                         } else {
-                            pb.finish_with_message("❌ Build failed.");
+                            pb.finish_with_message(error_message());
                             anyhow::bail!("❌ Failed to build index.");
                         }
                     }
                     Err(e) => {
-                        pb.finish_with_message("❌ Build failed.");
+                        pb.finish_with_message(error_message());
                         anyhow::bail!("❌ Failed to determine process exit status: {e}.",);
                     }
                 }
             }
             Err(e) => {
-                pb.finish_with_message("❌ Build failed.");
+                pb.finish_with_message(error_message());
                 anyhow::bail!(format!("❌ Error: {e}",));
             }
         }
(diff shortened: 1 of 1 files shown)


ra0x3:
@dadepo 
- I think this is fair
- However, I do think this message should only be logged when the `-v --verbose` flag is included

dadepo:
> @dadepo
> 
> * I think this is fair
> * However, I do think this message should only be logged when the `-v --verbose` flag is included

@ra0x3 Updated to as suggested. I only extended the error message for one of the 3 error paths, as I was not sure if the other two can be due to the missing env variables.

ra0x3:
@dadepo 
- Sorry CI is failing here :( still trying to get this release issue resolved
- In the meantime, this PR looks ok w/ me and I'll defer to @lostman or @deekerno for any additional feedback
- We will get this merged ASAP :)

plugins/forc-index/src/ops/forc_index_build.rs
```diff
@@ -226,3 +228,40 @@ pub fn init(command: BuildCommand) -> anyhow::Result<()> {
...
+            extra_msg.push_str("\nexport CPPFLAGS='-I/opt/homebrew/opt/llvm/include'");
+        }
+
+        if !extra_msg.is_empty() {
+            extra_msg.insert_str(
+                0,
+                r#"
+For Apple Silicon macOS users the preinstalled llvm has limited WASM targets,
+please install a binary with better support from Homebrew (brew install llvm)
+and configuring rustc with necessary environment variables:
```
lostman:
```suggestion
For Apple Silicon macOS users, the preinstalled llvm has limited WASM targets.
Please install a binary with better support from Homebrew (brew install llvm)
and configure rustc with the necessary environment variables:
```

Minor grammar corrections.

lostman (changes requested):
Looks good. Just requested minor corrections.

dadepo:
@lostman Thanks. I have accepted your suggestions.

ra0x3 (approved):

lostman (approved):

deekerno (approved):

PR #1313, opened 2023-08-30
fix: update MessageOut data parsing due to API change

deekerno:
### Description

Due to [a change in how the `data` field in a `MessageOut` receipt is formed](https://github.com/FuelLabs/fuel-vm/commit/18a2e4a13094448068e378baf3b89e448ec356d6), our type ID and data parsing did not cover each case. This PR adjusts the parsing to check for a `None` value as well as an empty byte array as well.

### Testing steps

On `develop`: Clear database, build any indexer (`fuel-explorer` is probably best) and run indexer on internal beta-4 test network. You should see a failure at or near block 364788.

On this branch: Repeat above steps. You should see no failure.


ra0x3 (approved):
utACK 🏅

PR #1315, opened 2023-08-30
fix: fix overly broad tuple type check

deekerno:
### Description

While debugging an issue for a grantee, I noticed that `is_tuple_type()` was causing the unit type (`()`) to not be included as part of the hash table keeps track of types in a contract ABI. After fixing that, I noticed that we needed to add a unit type check to `is_nondecodable_type()` so that we do not create a decoder for it.

### Testing steps

On develop:
1. `cargo build -p forc-index`
2. Create a new indexer.
3. Change the `fuel-indexer-utils` dependency to use the local package (make sure you're still on `develop`).
4. Save the following reproduction ABI and adjust the indexer manifest accordingly:
```
{
  "types": [
    {
      "typeId": 0,
      "type": "()",
      "components": [],
      "typeParameters": null
    },
    {
      "typeId": 1,
      "type": "enum ExampleEnum",
      "components": [
        {
          "name": "A",
          "type": 0,
          "typeArguments": null
        }
      ],
      "typeParameters": null
    }
  ],
  "functions": [
    {
      "inputs": [
        {
          "name": "e",
          "type": 1,
          "typeArguments": null
        }
      ],
      "name": "example_fn",
      "output": {
        "name": "",
        "type": 0,
        "typeArguments": null
      },
      "attributes": null
    }
  ],
  "loggedTypes": [],
  "messagesTypes": [],
  "configurables": []
}
```
5. `cargo build`

The build should fail with the error:
`message: Could not derive TypeApplication param types.: InvalidData("type id 0 not found in type lookup")`.

On this branch:

1. Switch to this branch and ensure that you have the changes from this PR present locally.
2. Run `cargo build` again.

The indexer should successfully build.

### Changelog

- https://github.com/FuelLabs/fuel-indexer/commit/2c87dd4d991660acf5d9afc243d5f7606130e2b8 - fix type check
- https://github.com/FuelLabs/fuel-indexer/commit/1e1900ecd3c8ca52bb9d5bf17ed2f43b07edbe01 - upgrade test contracts to Sway v0.45.0

### Notes

The ABI included above was generated for the following Sway contract using Sway v0.45:
```
contract;

pub enum ExampleEnum {
    A: (),
}

abi BugReproduction {
    fn example_fn(e: ExampleEnum);
}

impl BugReproduction for Contract {
    fn example_fn(e: ExampleEnum) {
        match e {
            ExampleEnum::A => {},
        }
    }
}
```


packages/fuel-indexer-macros/src/helpers.rs
```diff
@@ -93,6 +104,10 @@ pub fn is_non_decodable_type(typ: &TypeDeclaration) -> bool {
         return true;
```
ra0x3, line 104: Nit:
```rust
pub fn is_ignored_type(typ: &TypeDeclaration) -> bool {
  is_tuple_type(typ) || is_unit_type(typ)
}
```

ra0x3 (approved):
- Looks Ok thanks @deekerno 
- Left a small comment regarding style (not a blocker for merge)
- We can cut a patch release for this if necessary - but beta-3 is only gonna last for a few more days anyway

packages/fuel-indexer-macros/src/helpers.rs
deekerno: @ra0x3 We can't add the unit type check to the ignored type as it would lead to the same error coming back; see https://github.com/FuelLabs/fuel-indexer/blob/develop/packages/fuel-indexer-macros/src/indexer.rs#L65-L74:
```
    let abi_types_tyid = abi_types
        .iter()
        .filter(|typ| {
            if is_ignored_type(typ) {
                return false;
            }
            true
        })
        .map(|typ| (typ.type_id, typ.clone()))
        .collect::<HashMap<usize, TypeDeclaration>>();
```

PR #1318, opened 2023-08-31
chore: add graphql schema validation

### LINKED ISSUE #1046: Add much more GraphQL schema validation
- We have a `GraphQLSchemaValidator` in `fuel_indexer_lib::graphql::validator`
- This object is responsible for making sure users aren't writing GraphQL schema that will throw the indexer in a weird state (whether at compile time or runtime)
- We need to add _a lot_ more rules to this validator, since currently, users can do some wacky stuff that puts the service into weird states

Some rules I can think of off the top of my head are:
- [x] No nested lists in schema
- [x] ~~No nested foreign key references~~
- [x] All `id` field names must be of `ID`
- [x] Virtual types (tagged with a `@virtual` directive) can't have `id: ID` fields
- [x] Can't use reserved names
- [x] ~~Can't have `TypeDefinition` with too long of name~~
- [x] ~~Can't have `FieldDefinition` with too long of a name~~
- [x] Can't use unique directive `@unique` on a foreign key field
- [x] Can't use `@unique` directive on any `TypeDefinition` that is already `@virtual`
- [x] ~~If using `::new()` apparently can't include a `u64` foreign key as the first arg?~~
- [x] ~~Within the context of list types, that any `Vec` that could possibly be empty is typed as `[Foo!]` or `[Foo]` (where the outer list is optional)~~
- [x] Can't have a TypeDefinition with many-to-many relationships that exceed `MAX_MANY_TO_MANY_RELATIONSHIPS = 10`
  - Defined in this PR https://github.com/FuelLabs/fuel-indexer/pull/1140
- [x] Can't have a `FieldDefinition` that is a many-to-many relationship, where the FK is explicit (i.e., the FK points to anything other than an `ID` type on the child TypeDefintion)
 

(We can add to this list as we think of more things)

Additional context:
- These rules should also be clearly document in the book

ra0x3:

### Checklist

### Description

- PR adds tests (and updates parser functionality) to ensure users GraphQL schema can't get into weird states

### Testing steps

- [ ] Several tests were added so CI should pass

### Changelog

- chore: add graphql schema validation

### DIFF
--- a/packages/fuel-indexer-lib/src/graphql/parser.rs
+++ b/packages/fuel-indexer-lib/src/graphql/parser.rs
@@ -600,6 +600,8 @@ impl SchemaDecoder {
         node: TypeDefinition,
         u: &UnionType,
     ) {
+        GraphQLSchemaValidator::check_disallowed_graphql_typedef_name(&union_name);
+
         self.parsed_graphql_schema
             .parsed_typedef_names
             .insert(union_name.clone());
@@ -611,7 +613,7 @@ impl SchemaDecoder {
             .union_names
             .insert(union_name.clone());
 
-        GraphQLSchemaValidator::check_derived_union_is_well_formed(
+        GraphQLSchemaValidator::check_derived_union_virtuality_is_well_formed(
             &node,
             &mut self.parsed_graphql_schema.virtual_type_names,
         );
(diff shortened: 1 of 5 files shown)


deekerno (approved):
LGTM. 👌🏽

PR #1327, opened 2023-09-04
chore: fuel-indexer: refactor run, service, and executor to take advantage of tokio::task::JoinSet

lostman:
### Description

This PR cleans up parts of `IndexerService`, `Executor`, and `fuel-indexer`'s `run.rs`.

There is a slight change of semantics due to the replacement of `FuturesUnordered` with `JoinSet`: `JoinSet` aborts all tasks when it is dropped. This appeared in `web-api` tests that dropped the `IndexerService` object before the tests' completion. Tasks added to `FuturesUnordered` would've been running anyway (since `task::spawn` detaches the future). Tasks spawned into the task set of `IndexerService` are aborted. This required a small change to the tests to keep the `IndexerService` alive.

### Testing steps

CI tests should pass.

### Changelog

* Replace `FuturesUnordered` with `tokio::task::JoinSet`
* Clean up code in `service.rs` and `executor.rs`: move the kill switch into `Executor` implementations and expose through the trait interface; expose the manifest the same way. This allows cleaning up some functions that now only need the executor and no other parameters.
* Simplify fuel-indexer's `run.rs` to use the `JoinSet` for all subsystems (node, web-api, signal-handler, etc.).
* `WasmIndexExecutor::new()` takes WASM bytes and not `ExecSource`: we were passing `ExecSource::Manifest` and returning `ExecSource::Registry(wasm_bytes)`. Instead, the WASM bytes are loaded before `new()` is called, and everything is simpler this way.
* `IndexerService::run` removes completed tasks from the `JoinSet` and runs the service loop: `loop { tokio::select! ... }`. This way, we don't need two separate tasks for it.

### Measuring resource usage:

Changes in this PR have no adverse effect on the resource usage:

Screenshot 2023-09-08 at 12 55 33 PM

### DIFF
--- a/packages/fuel-indexer/src/commands/run.rs
+++ b/packages/fuel-indexer/src/commands/run.rs
@@ -130,7 +130,12 @@ pub async fn exec(args: IndexerArgs) -> anyhow::Result<()> {
         }
     }
 
-    subsystems.spawn(service.run());
+    subsystems.spawn(async {
+        let result = service.run().await;
+        if let Err(e) = result {
+            tracing::error!("Indexer Service failed: {e}");
+        }
+    });
 
     #[cfg(feature = "api-server")]
     subsystems.spawn({
--- a/packages/fuel-indexer/src/commands/run.rs
+++ b/packages/fuel-indexer/src/commands/run.rs
@@ -168,7 +168,7 @@ pub async fn exec(args: IndexerArgs) -> anyhow::Result<()> {
     // Each subsystem runs its own loop, and we require all subsystems for the
     // Indexer service to operate correctly. If any of the subsystems stops
     // running, the entire Indexer Service exits.
-    if let Some(_) = subsystems.join_next().await {
+    if subsystems.join_next().await.is_some() {
         subsystems.shutdown().await;
     }
 
--- a/packages/fuel-indexer/src/commands/run.rs
+++ b/packages/fuel-indexer/src/commands/run.rs
@@ -8,55 +8,47 @@ use fuel_indexer_lib::{
 };
 use tokio::signal::unix::{signal, Signal, SignalKind};
 use tokio::sync::mpsc::channel;
-use tokio_util::sync::CancellationToken;
 use tracing::info;
 
 #[cfg(feature = "api-server")]
 use fuel_indexer_api_server::api::WebApi;
 
-// Returns a CancellationToken which will be notified when a shutdown signal
-// have been reveived.
-async fn shutdown_signal_handler() -> anyhow::Result<CancellationToken> {
-    let cancel_token = tokio_util::sync::CancellationToken::new();
-
+// Returns a future which completes when a shutdown signal has been reveived.
+fn shutdown_signal_handler() -> std::io::Result<impl futures::Future<Output = ()>> {
     let mut sighup: Signal = signal(SignalKind::hangup())?;
     let mut sigterm: Signal = signal(SignalKind::terminate())?;
     let mut sigint: Signal = signal(SignalKind::interrupt())?;
 
-    tokio::spawn({
-        let cancel_token = cancel_token.clone();
-
-        async move {
-            #[cfg(unix)]
-            {
-                tokio::select! {
-                    _ = sighup.recv() => {
-                        info!("Received SIGHUP. Stopping services.");
-                    }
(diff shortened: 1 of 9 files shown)


packages/fuel-indexer/src/commands/run.rs
```diff
@@ -8,55 +8,47 @@ use fuel_indexer_lib::{
...
 
 #[cfg(feature = "api-server")]
 use fuel_indexer_api_server::api::WebApi;
 
-// Returns a CancellationToken which will be notified when a shutdown signal
-// have been reveived.
-async fn shutdown_signal_handler() -> anyhow::Result<CancellationToken> {
-    let cancel_token = tokio_util::sync::CancellationToken::new();
-
+// Returns a future which completes when a shutdown signal has been reveived.
```
deekerno: s/reveived/received

packages/fuel-indexer/src/commands/run.rs
```diff
@@ -156,28 +157,18 @@ pub async fn exec(args: IndexerArgs) -> anyhow::Result<()> {
...
             };
-            tokio::spawn(async move {
-                let node = FuelService::new_node(config).await.unwrap();
-                Some(node)
-            })
-        } else {
-            tokio::spawn(futures::future::ready(None))
+            subsystems.spawn(async move {
+                if let Err(e) = FuelService::new_node(config).await {
+                    tracing::error!("Fuen Node failed: {e}");
```
deekerno: s/Fuen/Fuel

packages/fuel-indexer/src/commands/run.rs
```diff
@@ -156,28 +157,18 @@ pub async fn exec(args: IndexerArgs) -> anyhow::Result<()> {
...
-            let _ = tokio::join!(service_handle, node_handle, web_handle);
-            cancel_token.cancel();
-        }
-    });
-
-    cancel_token.cancelled().await;
+    // Each subsystem runs its own loop, and we require all subsystems for the
+    // Indexer service to operate correctly. If any of the subsystems stops
+    // running, the entire Indexer Service exits.
+    subsystems.join_next().await;
```
deekerno: If a service stops, does this actually cause the others to exit? Or are we relying on the fact that an error will be returned in the respective components and waiting for `join_next()` to catch it?

I don't mind this, but what do you think about something a bit more explicit?:
```
if let Some(_res) = subsystems.join_next().await {
    subsystems.shutdown().await;
}
```

packages/fuel-indexer/src/executor.rs
```diff
@@ -742,76 +740,21 @@ impl WasmIndexExecutor {
...
-                    }
-                }
+        match WasmIndexExecutor::new(config, manifest, wasm_bytes, pool, schema_version)
+            .await
+        {
+            Ok(executor) => Ok(executor),
+            Err(e) => {
+                error!(
+                    "Could not instantiate WasmIndexExecutor({uid}) from ExecutorSource::Manifest: {e:?}."
+                );
```
deekerno: We're no longer instantiating from an `ExecutorSource` here, so this error message should be adjusted.

packages/fuel-indexer/src/service.rs
```diff
@@ -247,129 +253,108 @@ impl IndexerService {
...
-            },
-            Err(e) => {
-                debug!("No service request to handle: {e:?}.");
-                sleep(Duration::from_secs(defaults::IDLE_SERVICE_WAIT_SECS)).await;
             }
         }
     }
+
+    // Spawn and register a tokio::task running the Executor loop, as well as
+    // the kill switch and the abort handle.ß
```
deekerno: Looks like there's an errant `ß` here.

deekerno (changes requested):
This looks pretty good! I like the cleanup. I left a few comments to be addressed before approval.

packages/fuel-indexer/src/commands/run.rs
lostman: I like your suggestion.

If a subsystem stops, the others don't exit automatically. I was relying here on this `await` holding the `main` running. After this `await`, all that is left is the database cleanup and the `main` exits. Shutting things down explicitly is cleaner.

deekerno (approved):
🆗

Issue #1328, opened 2023-09-04
missing generics for enum `std::option::Option` during `forc index build`


chlenc:
# missing generics for enum `std::option::Option` during `forc index build`

Hi guys, I'm trying to use indexer in my swaylend contract on beta4 
Swaylend market contract deployed here:`0x2679d41c2981bcd6ab69d66f0ae656077efecac2c2dd729f88f5a6457ce1d5f1`
Swaylend code is here:
https://github.com/compolabs/sway-lend/blob/beta-4/contracts/market/src/main.sw

## The problem
I've created a brand new indexer using `forc index new swaylend-indexer  --namespace composabilitylabs `
Changed a bit `src/lib.rs, added abi and changed `swaylend_indexer.manifest.yaml`
You can find the swaylend indexer here
https://github.com/compolabs/sway-lend/tree/beta-4/services/swaylend-indexer

But during build, it fails with next error
```
>>> forc index build                           
▹▹▸▹▹ ⏰ Building...                                                                                                                                           Compiling swaylend_indexer v0.0.0 (/Users/alexey/projects/fuel/sway-lend/services/swaylend-indexer)
error[E0107]: missing generics for enum `std::option::Option`
   --> src/lib.rs:4:1
    |
4   | #[indexer(manifest = "swaylend_indexer.manifest.yaml")]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 1 generic argument
    |
note: enum defined here, with 1 generic parameter: `T`
   --> /Users/alexey/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:563:10
    |
563 | pub enum Option<T> {
    |          ^^^^^^ -
    = note: this error originates in the attribute macro `indexer` (in Nightly builds, run with -Z macro-backtrace for more info)
help: add missing generic argument
    |
4   | #[indexer(manifest = "swaylend_indexer.manifest.yaml")]<T>
    |                                                        +++

For more information about this error, try `rustc --explain E0107`.
error: could not compile `swaylend_indexer` (lib) due to previous error
▪▪▪▪▪ ❌ Build failed.                                                                                                                                      Error: ❌ Failed to build index.
```

![Uploading image.png…]()

## Toolchain
```
Default host: aarch64-apple-darwin
fuelup home: /Users/alexey/.fuelup

installed toolchains
[...]

PR #1332, opened 2023-09-05
enhancement: add support for sway generics

### LINKED ISSUE #503: Add ABI support for `Vec`
- Blocked by https://github.com/FuelLabs/fuel-indexer/issues/423
- Sway supports `Vec` now, so we need to do the same
- The info below is from an example given to me by @mohammadfawaz 
- Will need to add `Vec` support to our macro suite
- This `Vec` usage need to go in our test index and will require both Postgres & Sqlite indexing e2e tests

### Contract
```sway
abi MyContract {
    fn foo(v: Vec<u64>);
}

impl MyContract for Contract {
    fn foo(v: Vec<u64>) {}
}
```

### ABI
```json
{
  "types": [
    {
      "typeId": 0,
      "type": "()",
      "components": [],
      "typeParameters": null
    },
    {
      "typeId": 1,
      "type": "generic T",
      "components": null,
      "typeParameters": null
    },
    {
      "typeId": 2,
      "type": "raw untyped ptr",
      "components": null,
      "typeParameters": null
    },
    {
      "typeId": 3,
      "type": "struct RawVec",
      "components": [
        {
          "name": "ptr",
          "type": 2,
          "typeArguments": null
        },
        {
          "name": "cap",
          "type": 5,
          "typeArguments": null
        }
      ],
      "typeParameters": [
        1
      ]
    },
    {
      "typeId": 4,
      "type": "struct Vec",
      "components": [
        {
          "name": "buf",
          "type": 3,
          "typeArguments": [
            {
              "name": "",
              "type": 1,
              "typeArguments": null
            }
          ]
        },
        {
          "name": "len",
          "type": 5,
          "typeArguments": null
        }
      ],
      "typeParameters": [
        1
      ]
    },
    {
      "typeId": 5,
      "type": "u64",
      "components": null,
      "typeParameters": null
    }
  ],
  "functions": [
    {
      "inputs": [
        {
          "name": "v",
          "type": 4,
          "typeArguments": [
            {
              "name": "",
              "type": 5,
              "typeArguments": null
            }
          ]
        }
      ],
      "name": "foo",
      "output": {
        "name": "",
        "type": 0,
        "typeArguments": null
      },
      "attributes": null
    }
  ],
  "loggedTypes": [],
  "messagesTypes": [],
  "configurables": []
}
```

A further note from @mohammadfawaz :

> As you can see, Vec is basically a struct but contains a raw pointer which points to the actual data. Now I’m not sure what the indexer needs to do in this case, but the SDK basically interpret Vec as a special type and when passing a vector argument, it converts the Rust’s Vec to script data and then creates a Sway Vec with the raw pointer pointing to that raw data

### LINKED ISSUE #1334: Support `Option` types in ABI JSON
- This is per https://github.com/FuelLabs/fuel-indexer/issues/1328
- We should make sure we support `Option` types in ABI JSON
- We'd just map these ABI `TypeDeclaration`s to Rust tokens, same as we do with everything else
- Might be able to re-use some things from https://github.com/FuelLabs/fuel-indexer/pull/1332

ra0x3:

### Checklist

### Description

PR does a few things:
  - "Adds" support for Vec types
    - Adds in quotes because according to @segfault-magnet Sway is still not ready (October-ish estimated date)
    - However, when Sway makes that update, we won't really have anything to do after this PR is merged because all the heavy lifting is here
      - All we have to do when the time comes is add a test for `Vec` types
 - Adds support for Option types
 - Refactors the `#[indexer]` macro codegen to be more clear and support arbitrary generic types
 - Adds a trybuild test for unsupported  types used in indexer handler function args
   - We already had a test to ensure unsupported types don't break the ABI part of the codegen
- ⚠️ I have confirmed Spark's ABI in their issue https://github.com/FuelLabs/fuel-indexer/issues/1328 builds with the work from this PR
  - Not saying there might not be another issue that comes up

### Testing steps

- [ ] An indexing test was added for the new Option type so CI should pass

### Changelog

- enhancement: add support for sway generics


ra0x3:
CC @deekerno and @lostman since this updated `ColumnType`

@AlicanC When you're OK with this PR, feel free to merge 👌🏽

packages/fuel-indexer-macros/src/helpers.rs
```diff
@@ -711,3 +745,360 @@ pub fn can_derive_id(field_set: &HashSet<String>, field_name: &str) -> bool {
...
+
+    result.first().expect("No inner type found.")
+}
+
+/// Derive the inner ident names for collections types.
+///
+/// Given a `GenericType`, and ABI JSON metadata, derive the inner `TypeDefinition`s associated with the given
+/// generic `TypeDefinition`.
+///
+/// So this function will parsed all function inputs/outputs and log types, find all generics (e.g., `Vec<T>`, `Option<U>`),
```
lostman:
```suggestion
/// So this function will parse all function inputs/outputs and log types, find all generics (e.g., `Vec<T>`, `Option<U>`),
```

lostman (dismissed):
👍

lostman (approved):

deekerno (approved):

Issue #1150, opened 2023-07-27
Don't allow the service to miss blocks


ra0x3:
- Using the explorer dashboard I can see we're missing a few block when indexing for the explorer
  - [Grafana link](https://graphana-beta-1.fuel.network/d/a266b7cc-4872-4882-87b7-2c5930f3dbb5/fuel-explorer?orgId=1)
- Obviously the explorer should not be missing _any_ blocks
- However, it's also safe to assume we're going to have some 0.01% margin of error
- We need to define some way to index the blocks we missed

Additional context
- The quick win here is probably to just create a solution for the explorer only
  - The better/longer solution would be to define a solution that will work for _any_ indexer

ra0x3:
[Comment to keep in mind](https://fuellabs.slack.com/archives/C04BQKYTJQJ/p1690473162835619?thread_ts=1690467674.855509&cid=C04BQKYTJQJ)

ra0x3:
- Indexing blocks should _always_ be sequential - `Block(1), Block(2), Block(3), ... Block(n)`
- If the indexer is ever at a state where `$LAST_INDEXED_BLOCK_HEIGHT - $CURRENT_BLOCK_HEIGHT != 1` the the service should pause/wait until the issue is remediated
- The service should _not_ continue to index blocks if things are out of order
- I.e., we should detect whether we are `$IN_SYNC` or `$OUT_OF_SYNC` before even calling `handle_event` in the executor
- [How TheGraph allows indexers to opt-in to this (somewhat similar)](https://thegraph.com/docs/en/developing/creating-a-subgraph/#non-fatal-errors)
  - `syncrhonize: true/false` (or something similar) might be a good config option to add to the manifest

ra0x3:
- Just an FYI here that we are most likely going to have to devise a mechanism to allow the `fuel-indexer` (the service) to store blocks, outside of the context of any individual indexer
- We will use this cache of blocks in order to determine whether any given indexer is in or out of sync
- We have to store this cache of blocks on the service level, because `Block`s and `Transaction`s are not required for indexers
- This ☝🏽 is slightly related to https://github.com/FuelLabs/fuel-indexer/issues/990, just without the concept of migrating/backfilling (although those would side benefits we'd get for free)

PR #1337, opened 2023-09-07
enhancement: WASM error codes, early exit on error or kill switch

### LINKED ISSUE #1156: Add error codes to WASM executors
- Without error codes we have literally no way of knowing why these executors are having issues 
- We need to use error codes the similar to how we use `thiserror` error enums
- Every panic/unwrap/expect should be associated with an error code
- Example error:

```text
2023-07-28T16:01:57.023060Z ERROR fuel_indexer::executor: 821: WasmIndexExecutor WASM execution failed: RuntimeError { source: Wasm { pc: 4540778896, backtrace:    0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
```

lostman:
### Description

UPDATE: in addition to adding error codes to host functions as described below, this PR adds another host function:

```rust
fn early_exit(err_code: WasmIndexerError) -> !
```

Using this function we can remove more `unwrap` or `expect` calls.

Building on https://github.com/FuelLabs/fuel-indexer/pull/1293.

This PR follows the example in `wasmer` repository:

https://github.com/wasmerio/wasmer/blob/master/examples/early_exit.rs#L60

```
    fn early_exit() -> Result<(), ExitCode> {
        // This is where it happens.
        Err(ExitCode(1))
    }
```

Note that the return value is `Result<(), ExitCode>` while the function is imported in WASM as

https://github.com/wasmerio/wasmer/blob/master/examples/early_exit.rs#L41

```
  (type $early_exit_t (func (param) (result)))
```

That is, as if it were `fn early_exit();`.

This PR does the same for our FFI functions. For instance:

```rust
fn put_object(
    mut env: FunctionEnvMut<IndexEnv>,
    type_id: i64,
    ptr: u32,
    len: u32,
) -> Result<(), WasmIndexerError> { }
```

While their types—as declared to the WASM module which will use them—doesn't mention the `Result` value:

https://github.com/FuelLabs/fuel-indexer/pull/1337/files#diff-447e55cb0fae7d02ac8fd8f7de5a521cbf60d3f2422b5166c39faeb9f995bea6R21-R29

The one difference between how things are handled in `early_exit` example versus this PR is getting the error code out:

https://github.com/wasmerio/wasmer/blob/master/examples/early_exit.rs#L93

When I tried the same method:
```rust
Err(e) => match e.downcast::<ExitCode>()
```

I got no error out of it. However, the error was printed as `RuntimeError(User(EarlyExit))`, so the error was handled correctly. (`User` indicates that a user-defined error triggered a WASM trap).

So, I ended up handling the error this way:

https://github.com/FuelLabs/fuel-indexer/pull/1337/files#diff-fa722c66d2896c684e5a26aa2c56b7ec1734888609fbb64ecbc7dfd6a420c4ffR933

```rust
            } else {
                if let Some(e) = e
                    .source()
                    .and_then(|e| e.downcast_ref::<WasmIndexerError>())
                {
                    error!("Indexer({uid}) WASM execution failed: {e}.");
                } else {
                    error!("Indexer({uid}) WASM execution failed: {e:?}.");
                };
                self.db.lock().await.revert_transaction().await?;
                return Err(IndexerError::from(e));
            }
```

I am unsure why I couldn't `downcast` the error the same way as was shown in the example.

### Testing steps

Manual testing:

1. Start an indexer.

```
cargo run -p fuel-indexer -- run --fuel-node-host beta-4.fuel.network --fuel-node-port 80 --replace-indexer --manifest examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml
```

2. Redeploy:

```
cargo run -p forc-index -- deploy --path examples/fuel-explorer/fuel-explorer --replace-indexer
```

Expected output:

```
2023-09-07T09:12:49.663539Z  INFO fuel_indexer::service: 399: Resuming Indexer(fuellabs.explorer) from block 1240
2023-09-07T09:12:54.321165Z  INFO fuel_indexer::database: 206: Database loading schema for Indexer(fuellabs.explorer) with Version(143093c6bbe3a8937f4fb71514cbe6799266c44398866dbbd23e12b977bc1641).
2023-09-07T09:12:54.346555Z  INFO fuel_indexer::executor: 110: Indexer(fuellabs.explorer) subscribing to Fuel node at beta-4.fuel.network:80
2023-09-07T09:12:54.346681Z  WARN fuel_indexer::executor: 117: No end_block specified in manifest. Indexer will run forever.
2023-09-07T09:12:54.346717Z  INFO fuel_indexer::service: 335: Indexer(fuellabs.explorer) was replaced. Stopping previous version of Indexer(fuellabs.explorer).
2023-09-07T09:12:54.347376Z ERROR fuel_indexer::executor: 941: Indexer(fuellabs.explorer) WASM execution failed: Kill switch has been triggered.
2023-09-07T09:12:54.347819Z  INFO fuel_indexer::executor: 199: Kill switch flipped, stopping Indexer(fuellabs.explorer). <('.')>
```

In particular, this line indicates that an early termination happened due to the kill switch:

```
2023-09-07T09:12:54.347376Z ERROR fuel_indexer::executor: 941: Indexer(fuellabs.explorer) WASM execution failed: Kill switch has been triggered.
```

### Changelog

* FFI functions can access the kill switch indicator and exit early when the kill switch has been triggered.
* FFI functions can exit early on error, including `sqlx` database operation errors.
* Add WASM error codes (from @deekerno's PR)
* Get an error code from the call to the indexer's WASM module.


ra0x3:
@lostman 
- What's the difference between this and https://github.com/FuelLabs/fuel-indexer/pull/1293 ?
- Asking because this was ping'd for review but is based on `develop` (but description says it's a follow up)

lostman:
> @lostman What's the difference between this and #1293 ?

@deekerno's PR was a starting point. This PR adds a few things. From changelog:

FFI functions can access the kill switch indicator and exit early when the kill switch has been triggered. (added in this PR)
FFI functions can exit early on error, including sqlx database operation errors. (added in this PR)
Add WASM error codes (from @deekerno's PR)
Get an error code from the call to the indexer's WASM module. (added in this PR)

The most significant difference is how the error codes are returned:

https://github.com/FuelLabs/fuel-indexer/pull/1337/commits/e74645e0d418db7834e8a271ad50edcaaf055624#diff-447e55cb0fae7d02ac8fd8f7de5a521cbf60d3f2422b5166c39faeb9f995bea6L27-L28

In @deekerno's PR, the result is `u32` error code. As I explain in the description, in this PR, I follow the `early_exit` example—the function signature, as declared in the FFI interface mentions no `Result` type. `wasmer` terminates the execution immediately if the host function returns an error this way (which is exactly what we want here).

This way, even `get_object`, which returns `u32`, can be terminated early:

https://github.com/FuelLabs/fuel-indexer/pull/1337/files#diff-25a226eeb0da9150a8af6c4267e5745445aab8a84cbe942fa13d8a7756246c11R119

It now returns `Result<u32, WasmIndexerError>`. However, in the FFI interface, it is still `-> *mut u8`.

ra0x3:
@lostman 
- Reading over your testing steps, I see the expected output for this PR, however, what is the result as of today on `develop`? Trying to compare what we have now (prior to this PR) to what we will have after this is merged

packages/fuel-indexer/src/database.rs
```diff
@@ -64,29 +64,28 @@ impl Database {
         let conn = self.pool.acquire().await?;
         self.stashed = Some(conn);
         debug!("Connection stashed as: {:?}", self.stashed);
-        let conn = self.stashed.as_mut().expect(
-            "No stashed connection for start transaction. Was a transaction started?",
-        );
+        let conn = self.stashed.as_mut().ok_or(crate::IndexerError::Unknown(
```
deekerno: Since we're returning an error due to the absence of a stashed transaction, can we add a variant to `IndexerError` (e.g. `NoStashedConnection`) and use it instead of the `Unknown` variant in each of the places where we attempt to retrieve a stashed connection?

ra0x3 (commented):
@lostman 

- Following your testing steps I get the following output:
```text
2023-09-11T16:02:51.282146Z  WARN fuel_indexer::executor: 117: No end_block specified in manifest. Indexer will run forever.
2023-09-11T16:02:51.282175Z  INFO fuel_indexer::service: 335: Indexer(fuellabs.explorer) was replaced. Stopping previous version of Indexer(fuellabs.explorer).
2023-09-11T16:02:51.282689Z ERROR fuel_indexer::executor: 937: Indexer(fuellabs.explorer) WASM execution failed: Kill switch has been triggered.
2023-09-11T16:02:51.283092Z  INFO fuel_indexer::executor: 195: Kill switch flipped, stopping Indexer(fuellabs.explorer). <('.')>
```

UX Questions:
- Why am I seeing an `error!` about something failing? 
  - My re-deployment worked so ideally I shouldn't see any errors (as this can be confusing)
    - Maybe we can `warn!` any additional details?
- The last log message: "kill switch flipped" implies that the previous indexer was stopped (ok), but the message stops there, there's no additional info to suggest that my re-deployment worked 

Still reviewing the PR. But the logs are confusing me as to what's actually happening (from a user perspective)

packages/fuel-indexer/src/ffi.rs
```diff
@@ -70,8 +70,23 @@ fn get_object_id(mem: &MemoryView, ptr: u32, len: u32) -> FFIResult<String> {
...
     let (idx_env, store) = env.data_and_store_mut();
+
+    if idx_env
+        .kill_switch
+        .load(std::sync::atomic::Ordering::SeqCst)
+    {
+        // If the kill switch has been flipped, returning an error will cause an
+        // early termination of WASM execution.
+        return Err(WasmIndexerError::KillSwitch);
+    }
```
deekerno, line 88: This is definitely fine, but I can't help but wonder if there's a way to de-dupe this code and perhaps have a wrapper function that checks the kill switch before running the desired operation. 🤔

packages/fuel-indexer-macros/src/decoder.rs
```diff
@@ -868,7 +868,7 @@ impl From<ObjectDecoder> for TokenStream {
                                         Self::TYPE_ID,
                                         self.to_row(),
                                         serialize(&self.to_row())
-                                    ).await;
+                                    ).await.unwrap();
```
ra0x3: - No `unwrap()` unless you can prove it'll never panic
- Let's fail with a helpful error message here

packages/fuel-indexer-macros/src/decoder.rs
```diff
@@ -833,7 +833,7 @@ impl From<ObjectDecoder> for TokenStream {
                                             .map(|query| query.to_string())
                                             .collect::<Vec<_>>();
 
-                                        d.lock().await.put_many_to_many_record(queries).await;
+                                        d.lock().await.put_many_to_many_record(queries).await.unwrap();
```
ra0x3: - No `unwrap()` unless you can prove it'll never panic
- Let's fail with a helpful error message here

deekerno (changes requested):
Thanks for picking this up, @lostman! I left a few comments on the code itself.

Also, in addition to your manual testing, we should figure out a way to test the error codes themselves. Part of the reason my PR was still a WIP was that I hadn't figured out a good way to cause the errors (and that I was busy with other things 😅).

packages/fuel-indexer-plugin/src/wasm.rs
```diff
@@ -17,11 +18,14 @@ pub use sha2::{Digest, Sha256};
 pub use std::collections::{HashMap, HashSet};
 
 extern "C" {
-    // TODO: error codes? or just panic and let the runtime handle it?
+    // TODO: How do we want to return an error code for
```
ra0x3: - Instead of a `TODO` with explanations, the preference is for creating a Github issue, and leaving the issue link in the `// TODO: http://github.com/some-link`
- Helps keep track of when issues were found (and track them)
- Move the details from this `TODO` to the Github issue
- [Example](https://github.com/FuelLabs/fuel-indexer/blob/develop/packages/fuel-indexer/src/executor.rs#L85)

packages/fuel-indexer-plugin/src/wasm.rs
```diff
@@ -49,18 +53,28 @@ impl Logger {
     }
 }
 
+/// Trait for a type entity.
```
ra0x3, line 56: 👌🏽

lostman:
@ra0x3, agreed. I will add a special-case for the kill switch. That is, after all, the expected behavior.

packages/fuel-indexer-plugin/src/wasm.rs
```diff
@@ -71,40 +85,64 @@ pub trait Entity<'a>: Sized + PartialEq + Eq + std::fmt::Debug {
...
-            unsafe { ff_put_many_to_many_record(bytes.as_ptr(), bytes.len() as u32) }
+            unsafe {
+                ff_put_many_to_many_record(bytes.as_ptr(), bytes.len() as u32);
+            }
         }
     }
 
+    /// Loads a record given a UID.
     fn load(id: UID) -> Option<Self> {
+        Self::load_unsafe(id).unwrap()
```
ra0x3: We know what this `.unwrap()` is, but let's still use something more helpful here and leave a helpful error message

packages/fuel-indexer/src/executor.rs
```diff
@@ -917,7 +930,14 @@ impl Executor for WasmIndexExecutor {
                 self.db.lock().await.revert_transaction().await?;
                 return Err(IndexerError::RunTimeLimitExceededError);
             } else {
-                error!("WasmIndexExecutor({uid}) WASM execution failed: {e:?}.");
+                if let Some(e) = e
+                    .source()
+                    .and_then(|e| e.downcast_ref::<WasmIndexerError>())
+                {
+                    error!("Indexer({uid}) WASM execution failed: {e}.");
```
ra0x3: - Will `{e}` here instead of `{e:?}` swallow the extra debug info in `e` ?
- `e` will usually give the simple `&str` of the error, but `e:?` will sometimes include more info

ra0x3 (changes requested):
- Did a first pass
- For changes like this it would give us more confidence if there was an associated test to prove the functionality works (we already don't have many `service` or `executor` tests as is)
  - Tests also make the review easier

packages/fuel-indexer/src/executor.rs
lostman: In this case, `{e}` prints the `Display` message `"Kill switch has been triggered"` while `{e:?}` simply prints `KillSwitch` (and similarly for other variants).

packages/fuel-indexer/src/ffi.rs
lostman: I don't think it is possible to instantiate `HostFunction` with a closure, so I can't see how to make a wrapper.

lostman:
I added a `WasmIndexExecutor` test for exit codes. It is a bit crude but covers the basics.

packages/fuel-indexer-database/src/lib.rs
```diff
@@ -28,6 +28,8 @@ pub enum IndexerDatabaseError {
     Unknown,
     #[error("You don't own this indexer.")]
     NotYourIndexer,
+    #[error("No table mapping exists for TypeId({0:?})")]
```
ra0x3, line 31: ❤️

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -180,6 +180,12 @@ pub(crate) async fn remove_indexer(
     }))
     .await?;
 
+    // We have early termination on a kill switch. Yet, it is still possible for
```
ra0x3, line 183: 👌🏽

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+    UnableToFetchLogString,
+    KillSwitch,
+    DatabaseError,
+    GeneralError,
+}
+
+impl From<i32> for WasmIndexerError {
+    fn from(value: i32) -> Self {
+        match value {
+            0 => unreachable!("WasmIndexerError index starts at 1"),
```
ra0x3: This seems odd. Why would this not be a `GeneralError` ?

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+            _ => Self::GeneralError,
+        }
+    }
+}
+
+impl std::fmt::Display for WasmIndexerError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::SerializationError => {
+                write!(f, "Failed to serialize")
```
ra0x3:
```suggestion
                write!(f, "Failed to serialize object.")
```

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+            }
+            Self::DeserializationError => {
+                write!(f, "Failed to deserialize")
```
ra0x3:
```suggestion
                write!(f, "Failed to deserialize object.")
```

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+                write!(f, "Failed to save list")
+            }
+            Self::PutObjectError => {
+                write!(f, "Failed to save object")
+            }
+            Self::UninitializedMemory => {
+                write!(f, "Failed to create MemoryView for indexer")
+            }
+            Self::UnableToFetchLogString => {
+                write!(f, "Failed to fetch log string")
```
ra0x3: It's unclear what this implies? Is there a more simple or straightforward way to say this?

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+            }
+            Self::KillSwitch => {
+                write!(f, "Kill switch has been triggered")
```
ra0x3: It's not clear what "Kill switch" means in this case.
```suggestion
                write!(f, "Indexer kill switch has been triggered. Indexer will halt.")
```

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+            }
+            Self::DatabaseError => {
+                write!(f, "Failed performing a database operation")
```
ra0x3:
```suggestion
                write!(f, "Database operation failed.")
```

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+            }
+            Self::GeneralError => write!(f, "A WASM error occurred"),
```
ra0x3:
```suggestion
            Self::GeneralError => write!(f, "Some unspecified WASM error occurred."),
```

packages/fuel-indexer-lib/src/lib.rs
```diff
@@ -36,6 +36,70 @@ impl ExecutionSource {
...
+        }
+    }
+}
+
+impl std::error::Error for WasmIndexerError {}
```
ra0x3: Any reason not to `#[derive(thiserror::Error)]` for  `WasmIndexerError` ?

packages/fuel-indexer-macros/src/decoder.rs
```diff
@@ -833,7 +833,7 @@ impl From<ObjectDecoder> for TokenStream {
                                             .map(|query| query.to_string())
                                             .collect::<Vec<_>>();
 
-                                        d.lock().await.put_many_to_many_record(queries).await;
+                                        d.lock().await.put_many_to_many_record(queries).await.expect(&format!("Entity::save_many_to_many for {} failed.", stringify!(#ident)));
```
ra0x3, line 836: - I think here we should be logging the error messages (and `.expect`) as close to the actual culprit call as possible
- In this case the culprit of this error would be `fuel_indexer::database::Database:: put_many_to_many_record`
  - If the error is logged/handled there, we don't need to handle it here

packages/fuel-indexer-macros/src/decoder.rs
```diff
@@ -868,7 +871,7 @@ impl From<ObjectDecoder> for TokenStream {
                                         Self::TYPE_ID,
                                         self.to_row(),
                                         serialize(&self.to_row())
-                                    ).await;
+                                    ).await.expect(&format!("Entity::save for {} failed.", stringify!(#ident)));
```
ra0x3, line 874: Same comment here:

- I think here we should be logging the error messages (and `.expect`) as close to the actual culprit call as possible
- In this case the culprit of this error would be `fuel_indexer::database::Database:: put_object`
  - If the error is logged/handled there, we don't need to handle it here

packages/fuel-indexer-tests/tests/service.rs
```diff
@@ -85,3 +88,125 @@ async fn test_wasm_executor_can_meter_execution() {
         ),
     }
 }
+
+#[tokio::test]
+async fn test_wasm_executor_exit_codes() {
```
ra0x3: ❤️

ra0x3 (changes requested):
Left some more feedback

- @lostman I also wouldn't rush this PR
- We do have a deadline, but this work should be able to make it in _comfortably_ before that date, with plenty of time to spare
- This is an extremely sensitive PR, so let's not rush it

packages/fuel-indexer-macros/src/decoder.rs
lostman: The error is handled in `put_many_to_many_record`. In WASM, it is a host function that returns `Result<>`, which triggers early termination.

The code here is for native execution. This `expect` achieves an equivalent behavior.

packages/fuel-indexer-lib/src/lib.rs
lostman: `impl std::error::Error for WasmIndexerError {}` was in the `early_exit` example in `wasmer`.

packages/fuel-indexer-lib/src/lib.rs
lostman: I changed this. Error codes are arbitrary, so that we can start at `0`. Left `GeneralError` as a catch-all for the `From<u32>` implementation.

lostman:
> Left some more feedback
> 
> * @lostman I also wouldn't rush this PR
> * We do have a deadline, but this work should be able to make it in _comfortably_ before that date, with plenty of time to spare
> * This is an extremely sensitive PR, so let's not rush it

I'd rather merge this now (very soon) and have it battle-tested for a couple of weeks than merge it just before the deadline and have no time to resolve any issues that may arise.

I need this to finish #1150 and again, I'd rather test both for a couple of weeks before the deadline is upon us.

lostman:
@ra0x3, @deekerno,

I added an `early_exit` function and tried using it in `handler_block_wasm`:

https://github.com/FuelLabs/fuel-indexer/commit/adee0f8e08e2a404dbb067173cfc25f1cc0cda8d#diff-7121bb4841992a7a257dbb97c0e60f371c59f2845b4413d5413b7adf779feb5aR22

and some `trybuild` tests failed to link, complaining about missing `ff_early_exit` at the linking stage:

https://github.com/FuelLabs/fuel-indexer/actions/runs/6171631718/job/16750296869#step:13:106

All integration tests succeeded, and I could compile and deploy indexers.

Any idea what could be going on?

It would've been a nice use-case for this function.

ra0x3:
> Any idea what could be going on?

@lostman 

- In the trybuild tests, we manually add the missing FF symbols
- [Example](https://github.com/FuelLabs/fuel-indexer/blob/develop/packages/fuel-indexer-tests/trybuild/pass_if_unsupported_types_are_used.rs#L4)
- Let me know if this helps

lostman:
@ra0x3, yes, it helps. Thanks!

Just pushed this:
https://github.com/FuelLabs/fuel-indexer/pull/1337/commits/46507e8ef0082f6d0becdff4a51bbea7cfb9a2cf

And trybuild tests pass 😄 (locally; waiting on CI)

ra0x3:
@lostman Is this mean to be re-reviewed? I do remember leaving a previous review, but I see I'm ping'd for review again. But I also see a lot of the feedback was not implemented (and as well no reason was provided as to why).

lostman:
@ra0x3, yes, it is meant to be re-reviewed. I can't re-trigger the review request since it is already re-requested.

I implemented your feedback. What is missing?

ra0x3:
@lostman The review comments are unresolved, and they aren't labeled as "outdated". For any review comments where you implement the feedback, be sure to resolve the conversation (so it doesn't clutter the review, and look as if it's unresolved). And for anything you didn't implement, just leave a responding comment as to why you chose to to implement the feedback.

packages/fuel-indexer-macros/src/decoder.rs
lostman: The error is handled in `put_object`. In WASM, it is a host function that returns Result<>, which triggers early termination.

The code here is for native execution. This expect achieves an equivalent behavior.

packages/fuel-indexer-lib/src/lib.rs
lostman: This variant was unused so I removed it.

lostman:
Apologies. GitHub UI tricked me a little:

Screenshot 2023-09-15 at 4 31 14 PM

Completely missed these comments. 😂

ra0x3 (dismissed):

ra0x3:
@lostman CI failing

deekerno (approved):

ra0x3 (approved):

PR #1340, opened 2023-09-11
chore: remove indexer asset versioning logic

### LINKED ISSUE #1227: Remove logic related to indexer asset "versions"
- We originally used to `version` because we were allowing users to upload multiple instances of the same indexer
- However we've since replaced that with the `--replace-indexer` option, so `version` largely goes unused
- We should remove all references to this `version` stuff in the database, and in the web server use cases

lostman:
### Description

Closes #1227.

This PR removes the vestiges of indexer asset versions.

### Testing steps

CI tests should pass.

### Changelog

* remove appearances of indexer asset version
* rename `laset_asset_for_indexer` query to `indexer_asset`
* rename `laset_assets_for_indexer` query to `indexer_assets`

### DIFF
--- a/packages/fuel-indexer-database/database-types/src/lib.rs
+++ b/packages/fuel-indexer-database/database-types/src/lib.rs
@@ -566,9 +566,6 @@ pub struct IndexerAsset {
     /// Database ID of the indexer.
     pub index_id: i64,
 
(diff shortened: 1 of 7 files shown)


ra0x3 (approved):
Thanks for getting to this @lostman 

utACK

PR #1372, opened 2023-09-22
enhancement: Removed report_metrics from indexer manifest

### LINKED ISSUE #1347: Remove `report_metrics` from indexer manifest
- This flag was added at a time when we thought we'd report indexer metrics of some kind
- This can be removed, as we're not going that route

dadepo:
### Description

Removed report_metrics from indexer manifest

Closes #1347


ra0x3 (approved):
Thanks @dadepo

PR #1381, opened 2023-09-26
enhancement: add predicate support

ra0x3:
Thanks for opening a PR with the Fuel Indexer project. Please review the **Checklist** below and ensure you've completed all of the necessary steps to make this PR review as painless as possible.

### Checklist
- [ ] Ensure your top-level commit message is in line with our [contributor guidelines](./CONTRIBUTING.md).
- [ ] Please add proper labels.
- [ ] If there is an issue associated with this PR, please link the issue (right-hand sidebar)
- [ ] If there is not an issue associated with this PR, add this PR to the "Fuel Indexer" project (right-hand sidebar)
- [ ] Please allow Codeowners at least 24 hours to do a first-pass review.
- [ ] Please add thoroughly detailed testing steps below.
- [ ] Please keep your Changelog message short and sweet.

### Description

Please include a short description of what this PR does.

### Testing steps

Please provide the _exact_ testing steps for the reviewer(s) if this PR requires testing.

### Changelog

Please add neat Changelog info here, according to our [Contributor's Guide](./CONTRIBUTING.md).

Issue #1350, opened 2023-09-13
`forc index build` should preserve comments in manifest


ra0x3:
- Currently, if you `forc index build` with comments in your manifest, these comments will be removed after your manifest gets updated and re-written
- We should preserve whatever comments the manifest already had
- The only change to the manifest should be the module path

PR #1392, opened 2023-10-02
chore: update examples and qa script

### LINKED ISSUE #1069: Create `forc index` command for starting native indexers
- Native execution builds binaries, not libs
- Since these binaries are invoked using `IndexerArgs` -- which aren't exactly the same set of args as args passed to plugin commands like `forc index {build,deploy}` -- native execution doesn't support things like `--target-dir` in commands
  - Which means, you can't run a native indexer and pass the target dir at which the binary is located
- In order to allow native execution to feel like the rest of `forc index` we need a command to wrap `cargo run`
  - Maybe `forc index run` ?
- This new command would support things like `--target-dir` (to be specific)
- This is currently blocking native execution CI testing (without having to hack things)

ra0x3:

### Checklist

### Description

- PR does a few things:
  - Updates examples
    - Our "Hello-world" example has always been too complex for a hello-world program
    - This PR changes the previous "hello world" to be "greetings" and "hello world native" to be "greetings native"
    - I also created a new default indexer `forc index new` and called it "hello world" (since this is our _actual_ hello world example)
    - Also went through all examples/docs and updated all Docker components to ensure they work
    - Dockerized `greetings-native` example (all examples are now dockerized)
  - Adds new `forc index run-native` command to run native indexers

### Testing steps

- QA run report is attached below

#### Ensure new command works

```
# start local fuel node
cargo run -p greetings-fuel-client --bin greetings-fuel-client

# run native indexer
cargo run --bin forc-index -- run-native --path examples/greetings-native/greetings-native-indexer -- --run-migrations

# Trigger data 
cargo run -p greetings-data --bin greetings-data -- --host 0.0.0.0:4000

# Verify data in DB
```

#### Ensure examples work

- Stop any local Postgres

##### Hello world

```
# Go to example
cd examples/hello-world

# Spin up containers
docker compose up 

# Deploy
forc index deploy --path hello-world --url http://0.0.0.0:29987

# Trigger data
<< NO DATA TRIGGERED BECAUSE IT CONNECTS TO BETA-4 >> 

# Verify data in DB
```

##### Greetings

```
# Go to example
cd examples/greetings

# Spin up containers
docker compose up 

# Deploy
forc index deploy --path greetings-indexer --url http://0.0.0.0:29987

# Trigger data
cargo run -p greetings-data --bin greetings-data -- --host 0.0.0.0:4000

# Verify data in DB
```

##### Greetings Native

```
# Go to example
cd examples/greetings-native

# Spin up containers
docker compose up 

# Deploy
<< NO DEPLOY BECAUSE THIS IS NATIVE EXECUTION >>

# Trigger data
cargo run -p greetings-data --bin greetings-data -- --host 0.0.0.0:4000

# Verify data in DB
```

### Changelog

- chore: update examples and qa script


packages/fuel-indexer-benchmarks/src/bin/qa.rs
```diff
@@ -520,6 +566,29 @@ async fn main() {
     let manifest = Manifest::from_file(&mani_path).unwrap();
     let mut stats = StatManager::new();
 
+    // for i in num_additional_indexers {
```
lostman: Maybe it would be better to delete this code instead of leaving it commented out?

lostman (dismissed):

ra0x3:
### QA report
[Rashads-Air.lan-2023-09-28.indexer-qa.txt](https://github.com/FuelLabs/fuel-indexer/files/12785985/Rashads-Air.lan-2023-09-28.indexer-qa.txt)

packages/fuel-indexer-tests/scripts/test-examples.bash
```diff
@@ -6,61 +6,83 @@
...
+sed -i "s|^graphql_schema: examples/greetings-native/greetings-native-indexer/schema/greetings_native_indexer\.schema\.graphql|graphql_schema: $PWD/greetings-native-indexer/schema/greetings_native_indexer.schema.graphql|" ./greetings-native-indexer/greetings_native_indexer.manifest.yaml
+
+# Start indexer in the background
+forc-index run-native --path ./greetings-native-indexer -- --run-migrations --postgres-password my-secret
+sleep 60
 
-cargo build -p hello_indexer_native --locked --release
+check_service_status
+
+# cargo run -p greetings-data --bin greetings-data
```
ra0x3, line 78: - There's something odd happening on CI here where starting this indexer sort've blocks the rest of CI - even if the command is launched in the background. 
- I want to continue to look into this but don't want to block the PR
- You can confirm that running these steps locally works

packages/fuel-indexer-tests/scripts/test-examples.bash
```diff
@@ -6,61 +6,85 @@
...
-sleep 2
+forc_index_kill
 
 # ********************************
-# examples/hello-indexer-native
+# examples/greetings-native
 # ********************************
 
-# TODO: https://github.com/FuelLabs/fuel-indexer/issues/1069
+# cd ../greetings-native
```
ra0x3, line 65: - There's something odd happening on CI here where starting this indexer sort've blocks the rest of CI - even if the command is launched in the background. 
- I want to continue to look into this but don't want to block the PR
- However, you can confirm that running these steps locally works

deekerno:
@ra0x3 I'm reviewing this now; while I'm doing that, can you share the reasoning behind adding a separate command -- `forc index run-native` -- for running native indexers? It seems to me that we could keep it as `forc index run --native`.

ra0x3:
> @ra0x3 I'm reviewing this now...

@deekerno 
- We have `forc index start` but I don't believe we have a `forc index run` command?
- I originally called it `forc index run` but felt that might be confusing so I explicitly named it `run-native`

deekerno:
> > @ra0x3 I'm reviewing this now...
> 
> 
> 
> @deekerno 
> 
> - We have `forc index start` but I don't believe we have a `forc index run` command?
> 
> - I originally called it `forc index run` but felt that might be confusing so I explicitly named it `run-native`
> 
> 

Oops, you're totally right. And in that case, I think "run-native" is fine. I'm just thinking that we can keep it as a separate operation but maybe invoke it as `start --native` to keep the number of commands low.

ra0x3:
> > > @ra0x3 I'm reviewing this now...
> > 
> > 
> > @deekerno
> > 
> > * We have `forc index start` but I don't believe we have a `forc index run` command?
> > * I originally called it `forc index run` but felt that might be confusing so I explicitly named it `run-native`
> 
> Oops, you're totally right. And in that case, I think "run-native" is fine. I'm just thinking that we can keep it as a separate operation but maybe invoke it as `start --native` to keep the number of commands low.

@deekerno 
- So keep in mind `fuel-indexer` has a subcommand `run` and that doesn't apply to native indexers (so a different entrypoint is needed)
  - We'd have to do some voodoo where we compile the `run` command into native indexers
- Also an FYI (cc @lostman) that as I'm working on #1381 I think we're reaching the limits of what we can do with `if (is_native) { this } else { that }` 😅 
  - Hopefully after the leadership offsite we'll have more direction with where the project is going
  - I'm trying to keep the native stuff to the point where we can yank all `if $is_native` code in a single easy PR if we decide we should just nuke it (which is inching closer)

deekerno (dismissed):

examples/hello-world/hello-world/hello_world.manifest.yaml
```diff
@@ -0,0 +1,13 @@
+---
+namespace: fuellabs
+identifier: hello_world_2
```
lostman: Why make the default `hello_world_2` instead of `hello_world`?

packages/fuel-indexer-tests/tests/commands.rs
```diff
@@ -1,6 +1,10 @@
 use duct::cmd;
 use fuel_indexer_lib::config::IndexerConfig;
 
+const FUEL_INDEXER: &str = "./../../target/release/fuel-indexer";
```
lostman, line 4: We could run all of these through `cargo run --bin PKG` to make this fully automatic. These are better than before—no need to manipulate the path—but still require pre-building the binaries before running tests.

lostman (dismissed):

examples/hello-world/hello-world/hello_world.manifest.yaml
ra0x3: Doh, good catch. This manifest was changed after a `forc index deploy` .

We really need #1350 😅 . Will fix

deekerno (approved):

lostman (approved):

PR #1396, opened 2023-10-05
enhancement: generate GraphQL schema from JSON ABI

### LINKED ISSUE #846: Auto-generate GraphQL schema using contract ABI
- When you create a new indexer project with `forc index new` your schema, lib module, and manifest are auto-generated
- As of today, a user has to open their GraphQL schema file to add even their basic schema
- It would be cool if the GraphQL schema could be initially populated automagically
  - After all, we know the types from looking at contract ABI

Additional context
- This would be something that a user opts into
  - Some type of flag `--auto-graphql-types` (or some such) passed to `forc index new`
    - This flag means the user wants to just copy the types _exactly_ from the ABI, into their GraphQL equivalents
- This won't be possible (without some hacks) until we separate ABI structs and GraphQL structs into different namespaces
  - Currently, if you have a `Foo` struct in your Sway contract and a `Foo` in your GraphQL schema - the types will clash
  - Namespacing is cool, but also will require users to use types such as `my_abi::Foo` and `my_graphql::Foo` instead of just `Foo` (due to the naming collision).

lostman:
### Description

Closes #846.

This PR adds a `--json-abi` flag to `forc index new` which facilitates automatic generation of the GraphQL schema from the given JSON ABI.

### Testing steps

Create a new indexer:

```
cargo run -p forc-index -- new --json-abi ./packages/fuel-indexer-tests/trybuild/abi/DAO-contract-abi.json dao-indexer
```

Copy the JSON ABI:

```
cp ./packages/fuel-indexer-tests/trybuild/abi/DAO-contract-abi.json /tmp/dao-indexer
```

Set the `abi` in the `Manifest`

```
abi: DAO-contract-abi.json
```

Change `lib.rs` to refer to types from the generated schema:

```
extern crate alloc;
use fuel_indexer_utils::prelude::*;

#[indexer(manifest = "dao_indexer.manifest.yaml")]
pub mod dao_indexer_index_mod {

    fn dao_indexer_handler(ie: InitializeEvent) {
        info!("author: {:?}", ie.author);
        let iee = InitializeEventEntity::new(ie.author, ie.token);
        iee.get_or_create();
    }
}
```

Verify that it builds:

```
forc index build
```

### Changelog

* Add `generate_schema()`, which generates a GraphQL schema from JSON ABI.
* Add documentation.
* Move some functions from `packages/fuel-indexer-macros/src/helpers.rs` to `packages/fuel-indexer-lib/src/helpers.rs` because they are now used in both places.


docs/src/SUMMARY.md
```diff
@@ -21,6 +21,7 @@
   - [Scalars](./designing-a-schema/scalars.md)
   - [Directives](./designing-a-schema/directives.md)
   - [Relationships](./designing-a-schema/relationships.md)
+- [Generating a Schema](./generating-a-schema/index.md)
```
ra0x3, line 24: This should go int the `Designing a schema` section maybe under a section called `Automatic Schema Generation` (or something that makes it clear what this section is about).

docs/src/generating-a-schema/index.md
```diff
@@ -0,0 +1,155 @@
+# Automatically generating GraphQL schema from JSON ABI
+
+`forc index new` supports automatically generating GraphQL schema from a contract JSON ABI.
```
ra0x3, line 3: - I think this has to be re-written. 
- @lostman We spent _weeks_ on the docs, combing through every single line making sure they read properly and that all examples are informative and work
  - It sucked, it wasn't fun, but the docs are _so_ much better now
  - Let's keep this trend going with the docs in this PR.
  - Ideally the docs should:
    1. Show me the tiniest Sway smart contract with various simple types (struct, enum, etc)
    2. Show me how to use `forc index new --json-abi`
    3. Show me the resultant schema that gets produced from that command ☝🏼 
  - I think we have an example of this in the docs already [here](https://docs.fuel.network/docs/indexer/indexing-custom-types/) (again, I want to be very clear that I realize this is a PITA 😅 )

packages/fuel-indexer-lib/src/graphql/constants.rs
```diff
@@ -61,4 +61,25 @@ lazy_static! {
         "Option<Identity>",
         "Option<Json>",
     ]);
+
+    pub static ref ABI_TYPE_MAP: HashMap<&'static str, &'static str> = HashMap::from_iter([
```
ra0x3, line 66: Maybe we can get a rustdoc explaining exactly what these are? The name `ABI_TYPE_MAP` is, on its own, pretty generic/uninformative

packages/fuel-indexer-lib/src/graphql/schema_gen.rs
```diff
@@ -0,0 +1,149 @@
+use crate::{
```
ra0x3: Is there any particular reason this is in `fuel-indexer-lib` and not `fuel-indexer-macros` ? All of the codgen is 95% in `fuel-indexer-macros` (might be 100% but I'd have to double check).

packages/fuel-indexer-lib/src/helpers.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Whether a `TypeDeclaration` is tuple type
+pub fn is_tuple_type(typ: &TypeDeclaration) -> bool {
+    let mut type_field_chars = typ.type_field.chars();
+    type_field_chars.next().is_some_and(|c| c == '(')
+        && type_field_chars.next().is_some_and(|c| c != ')')
+}
+
+/// Whether a `TypeDeclaration` is a unit type
+pub fn is_unit_type(typ: &TypeDeclaration) -> bool {
```
ra0x3: @lostman 
- Ah ok, yea, as I mentioned in the "Is there any particular reason..." comment...
  - _All_ of this codegen logic needs to be moved to `fuel-indexer-macros` (we're duplicating a ton of logic here)
  
So maybe something like:
- `fuel_indexer_macros::schema_gen`
  - This is where you put all the logic related to auto-creating this schema
    - And you would just pull in `fuel_indexer_macros::helpers` instead of duplicating this stuff
    - We already have a `fuel_indexer_macro_utils` crate
       - This was previously being used for indexer-internal proc macros (as opposed to public macros)
       - You could potentially move that `#[metrics]` macro out of `fuel_indexer_macro_utils` and into `fuel_indexer_macros` (so now fuel_indexer_macros would export `#[metrics]` and `#[indexer`]
       - _Then_ that would allow you to put your macro logic in `fuel_indexer_macro_utils` (after you make it a non-proc-macro crate)
       - Finally, this would allow you to import the macro functions into `forc_index::new` via `use fuel_indexer_macro_utils::schema_gen::*`
         - I think this solves a lot of problems but let me know what you think

packages/fuel-indexer-tests/tests/snapshots/integration_tests__commands__forc_index_new_help_output.snap
```diff
@@ -13,6 +13,7 @@ ARGS:
 OPTIONS:
         --absolute-paths           Resolve indexer asset filepaths using absolute paths.
     -h, --help                     Print help information
+        --json-abi <JSON_ABI>      Path to JSON ABI for automatic schema generation.
```
ra0x3, line 16: 👌🏼

ra0x3 (changes requested):
@lostman 
- Left some feedback
- Mainly about restructuring this a bit, and using `fuel_indexer_macro_utils`

packages/fuel-indexer-lib/src/helpers.rs
lostman: The issue is that `fuel-indexer-macros` is a `proc-macro` crate and can only export macros.

So, `fuel_indexer_macros::schema_gen` doesn't work.

> we're duplicating a ton of logic here

Not duplicating. I moved the code from `fuel_indexer_macros::helpers`.

> Finally, this would allow you to import the macro functions into forc_index::new via use fuel_indexer_macro_utils::schema_gen::*

This is fine. I can move this logic there and make it non-proc-macro.

packages/fuel-indexer-lib/src/graphql/schema_gen.rs
lostman: `fuel-indexer-macros` is a `proc-macro` crate, so I can't use it from there.

packages/fuel-indexer-lib/src/helpers.rs
ra0x3: @lostman I mentioned in my previous comment

> Then that would allow you to put your macro logic in fuel_indexer_macro_utils (after you make it a non-proc-macro crate)

PR #1400, opened 2023-10-06
enhancement: compare fuel-indexer and forc-index versions

### LINKED ISSUE #1377: Return `405` on`forc index deploy` unless `forc-index` version matches `fuel-indexer` version
- Indexers compiled with different `fuel-indexer` and `forc-index` versions always present issues
  - Because we can have non-trivial changes between even patch versions
- We should add a `VERSION` to the indexer module that indicates what version of `forc-index` it was compiled with 
- When deployed via `forc-index deploy` if the compiled `VERSION` does not match the version of the service being used, should return a 405
- We might remove this of feature flag it in the future, but for now this tends to always be an issue

lostman:
### Description

Closes #1377.

This PR embeds `TOOLCHAIN_VERSION` in the indexer's WASM module and adds a check when the indexer is deploys which only permits the indexers whose `TOOLCHAIN_VERSION` matches the `fuel-indexer` version.

### Testing steps

CI tests should pass.

Manual testing:

1. Start `fuel-indexer`:

```
cargo run -p fuel-indexer -- run --fuel-node-host beta-4.fuel.network --fuel-node-port 80 --replace-indexer
```

2. Create and deploy an indexer using an older version of `forc-index`:

```
forc-index --version
forc index 0.20.11
```

```
forc-index new /tmp/test-indexer
forc-index deploy --path /tmp/some-indexer
```

```
▹▸▹▹▹ ⏰ Building indexer...
Finished release [optimized] target(s) in 0.10s
▪▪▪▪▪ ✅ Build succeeded.
Deploying indexer...
▸▹▹▹▹ 🚀 Deploying...
{
  "details": "WASM module toolchain version `unknown` does not match fuel-indexer version `0.21.1`",
  "success": "false"
}
```

Additionally, run `fuel-indexer` with `--disable-toolchain-version-check` to be able to deploy the above indexer anyway.

3. Deploy one of the indexers in the repository (hence, using the same toolchain version):

```
cargo run --bin forc-index -- deploy --path examples/fuel-explorer/fuel-explorer/
```

```
▹▹▸▹▹ ⏰
Building indexer...
    Finished release [optimized] target(s) in 0.40s
▪▪▪▪▪ ✅ Build succeeded.
Deploying indexer...
▪▪▪▪▪ ✅ Successfully deployed indexer.
```

### Changelog

* embed `TOOLCHAIN_VERSION` in WASM module
* check that the embedded version matches `fuel-indexer`'s version
* add a flag to disable this behavior


packages/fuel-indexer-api-server/src/ffi.rs
```diff
@@ -0,0 +1,75 @@
...
+        "ff_put_many_to_many_record".to_string(),
+        Function::new_typed(&mut store, |_: i32, _: i32| {}),
+    );
+    exports.insert(
+        "ff_log_data".to_string(),
+        Function::new_typed(&mut store, |_: i32, _: i32, _: i32| {}),
+    );
+
+    let mut imports = imports! {};
+    wasmer::Imports::register_namespace(&mut imports, "env", exports);
```
ra0x3, line 35: Magic variable here with `"env"` ? (It's actually used in `executor.rs` like this too)

packages/fuel-indexer-api-server/src/uses.rs
```diff
@@ -390,12 +402,23 @@ async fn parse_register_indexer_multipart(
...
+                if asset_type == IndexerAssetType::Wasm {
+                    toolchain_version =
+                        crate::ffi::check_wasm_toolchain_version(data.clone().into())
+                            .map_err(|e| {
+                                tracing::warn!(
+                                    "Failed to get WASM module toolchain version: {e}"
+                                );
+                                e
+                            })
+                            .unwrap_or("unknown".to_string());
```
ra0x3:
```suggestion
                            .unwrap_or(toolchain_version);
```

packages/fuel-indexer-api-server/src/ffi.rs
```diff
@@ -0,0 +1,75 @@
...
+
+/// Extract the `TOOLCHAIN_VERSION` string from a WASM module. This function
+/// creates a `wasmer::Instance` in order to do this.
+pub(crate) fn check_wasm_toolchain_version(data: Vec<u8>) -> anyhow::Result<String> {
+    let mut store = wasmer::Store::default();
+
+    let module = wasmer::Module::new(&store, data.clone())?;
+
+    let mut exports = Exports::new();
+    exports.insert(
```
ra0x3, line 13: I guess `wasmer` requires us to add these exports even though they're not actually used?

ra0x3 (dismissed):
utACK. Will defer to @deekerno

packages/fuel-indexer-api-server/src/ffi.rs
lostman: Yes, these are required or the module instantiation fails.

packages/fuel-indexer-api-server/src/ffi.rs
lostman: Multiple namespaces, multiple memories... I've yet see any worthwhile use of these. So the magic variable for namespace is `"env"` and magic variable for memory is `"memory"`. 🤷‍♂️

ra0x3 (approved):
Again deferring to @deekerno

deekerno (approved):

PR #1411, opened 2023-10-12
enhancement: support for Sway array types

### LINKED ISSUE #1384: Investigate support for Array `[_; n]` Sway types
- Might require a bit of investigation
- Whilst working on https://github.com/FuelLabs/fuel-indexer/pull/1341 I noticed that we currently don't support array-like `[_ ; 5]`
- Not sure if we should, but  needs a further look

ra0x3:

### Checklist

### Description

- PR adds support for indexing array types from Sway contracts
  - Array types are primitives (rust primitives) so no decoding work to be done

### Testing steps

- [ ] CI should pass (because an existing test was used for this work)

### Changelog

- enhancement: support for Sway array types


lostman (dismissed):

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -117,6 +117,40 @@ fn process_fn_items(
...
+            if is_array_type(typ) {
+                let inner = typ.components.as_ref().unwrap().first().unwrap().type_id;
+                let size = typ
+                    .type_field
+                    .split(' ')
+                    .last()
+                    .unwrap()
+                    .trim_end_matches(']')
+                    .parse::<usize>()
+                    .unwrap();
```
deekerno: Can we switch these unwraps for expects? Seems that the other additions have expects, so we should be consistent.

deekerno (dismissed):
LGTM, left one non-blocking (but important) comment.

PR #1412, opened 2023-10-13
enhancement: forc index status: report indexer status and errors

### LINKED ISSUE #1316: enhancement: forc index status: report errors and indexed blocks
### Report indexed blocks

The only way to observe that an indexer is working is to look at the data in the database.

Or, what I do practically all the time, add this to an indexer:

```rust
info!("Block #{}", block_data.height);
```

It works fine, but every time I switch a branch, I must stash/apply this. 😢 

We need a more convenient way to check that "it's working."

We can do that by making `forc index status` gather information about indexed blocks for each indexer (`MAX(block_height)` on the `indexmetadataentity` table) and display it.

### Save and display status/errors

If an indexer stops for any reason, we should save the error and display it in the output of `forc index status`.

This may include WASM error (and with WASM error codes PR, we'll get more useful errors: https://github.com/FuelLabs/fuel-indexer/pull/1293) or perhaps some database error that causes the indexer to miss blocks.

This PR https://github.com/FuelLabs/fuel-indexer/pull/1297 introduces a trigger on `indexmetadataentity` to ensure no blocks are missing (more info here: https://github.com/FuelLabs/fuel-indexer/issues/1150). It will stop the indexer from progressing. However, the only way to check what happened is to look through the logs:

```
Aug 31 10:50:10.335 ERROR fuel_indexer::database: Failed to put_object: Database(PgDatabaseError { severity: Error, code: "P0001", message: "fuel_indexer_test_index1.indexmetadataentity attempted to insert value with block_height = 5 while last processed block_height = 3. block_height values must be consecutive.", detail: None, hint: None, position: None, where: Some("PL/pgSQL function ensure_block_height_consecutive() line 8 at RAISE"), schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("pl_exec.c"), line: Some(3907), routine: Some("exec_stmt_raise") })
```

We can save this (and other) error messages in the database and include them in `forc index status`.

Example:

```
─ fuellabs
   └─ explorer
      • id: 78
      • created_at: 2023-08-28 10:13:28.617545 UTC
      • pubkey: None
      • indexed blocks: 12345
      • status: stopped with error: fuellabs_explorer.indexmetadataentity: attempted to insert value with block_height = 5 while last processed block_height = 3. block_height values must be consecutive.
```
Or simply:
```
      • status: active
```
If everything is well.

lostman:
### Description

Closes #1316.

This PR adds information about active indexers to `forc index status`.

### Testing steps

Start Fuel Indexer Service:
```
cargo run -p fuel-indexer -- run --fuel-node-host beta-4.fuel.network --fuel-node-port 80 --replace-indexer
```

Modify `fuel-explorer` to trigger a panic:

```
    fn index_block(block_data: BlockData) {
        let _x: Option<usize> = None.unwrap();
        ...
    }
```

Deploy:
```
forc-index deploy --path examples/fuel-explorer/fuel-explorer
```

Deploy another indexer:
```
forc-index deploy --path examples/hello-world/hello-world/
```

Check the status

```
```

Screenshot 2023-10-17 at 5 09 32 PM

`fuel-explorer`'s status should be an `error` while the other indexer's status should be `running`, and the status message should show an increasing block count.

### Changelog

* Add additional information to `forc index status`
* Add `IndexerStatus` and `IndexerStatusKind` types
* Create `index_status` table to hold the new information
* Add `humantime` formatting of `created at`, e.g. `(23 min 30s ago)`


packages/fuel-indexer-database/postgres/src/lib.rs
```diff
@@ -972,3 +977,57 @@ pub async fn remove_ensure_block_height_consecutive_trigger(
 
     Ok(())
 }
+
+/// Set the status of a registered indexers to be displayed by `forc index status`.
```
deekerno: s/indexers/indexer

packages/fuel-indexer-database/src/queries.rs
```diff
@@ -439,3 +445,28 @@ pub async fn remove_ensure_block_height_consecutive_trigger(
         }
     }
 }
+
+/// Set the status of a registered indexers to be displayed by `forc index status`.
```
deekerno: s/indexers/indexer

deekerno (dismissed):
This is excellent. Thanks @lostman!

packages/fuel-indexer-database/postgres/migrations/20231012190730_indexer_status.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS index_status (
+   id bigserial primary key,
```
ra0x3: @lostman Should this not be a foreign key to `index_registry`?

ra0x3 (commented):
@lostman 
- This mostly looks OK
- However, I'm personally not sure how this works
- Does this work for a single indexer? For a single public key? This works for all indexers under a given public key? What if I want to check the status of a single indexer?
  - If these have simple answers, I don't think we need docs
  - If these answers are a bit more invovled, maybe we'll have to add a `## Usage` section in ` docs > src > forc-index > status.md`

packages/fuel-indexer-database/postgres/migrations/20231012190730_indexer_status.up.sql
lostman: > Does this work for a single indexer? For a single public key? This works for all indexers under a given public key? What if I want to check the status of a single indexer?

If `fuel-indexer` does not require authentication, the output is all indexers with all the keys.

If `fuel-indexer` does require authentication, the output is the indexers for one's key.

There is no way to select a single indexer. It could be added easily. Something like:

```
forc index status fuellabs.explorer
```

I can add it to this PR if you'd like

ra0x3: @lostman No this is fine as is :)

ra0x3: @lostman Also, I see you commented under this particular comment about the foreign key 😆 . So before I resolve this thread, was there a reason `index_status` doesn't use a foreign key pointing to `index_registry` ?

packages/fuel-indexer-database/postgres/migrations/20231012190730_indexer_status.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS index_status (
+   indexer_id bigint,
+   namespace varchar(32) not null,
```
ra0x3: @lostman If we're keeping track of indexer info in `index_registry` and using a foreign key for `index_status` do we need to also keep the namespace and identifier here as well?

ra0x3 (changes requested):

packages/fuel-indexer-database/postgres/migrations/20231012190730_indexer_status.up.sql
lostman: Nope, we don't! Changed it in the latest commit.

ra0x3 (dismissed):

deekerno (dismissed):

ra0x3 (approved):

deekerno (approved):

PR #1415, opened 2023-10-16
chore: forc-index-build preserve yaml comments

### LINKED ISSUE #1350: `forc index build` should preserve comments in manifest

ra0x3:

### Checklist
- [x] Ensure your top-level commit message is in line with our [contributor guidelines](./CONTRIBUTING.md).
- [x] Please add proper labels.
- [x] If there is an issue associated with this PR, please link the issue (right-hand sidebar)
- [x] If there is not an issue associated with this PR, add this PR to the "Fuel Indexer" project (right-hand sidebar)
- [x] Please add thoroughly detailed testing steps below.

### Description

- PR makes it such that comments will remain in your manifest after `forc index build`
- `serde_yaml` obviously does not (and will not) handle comments
- The idea in this PR is to just add a small manifest proxy `RawManifest` to cache comments on read, and write the comments back on write
- This was actually not easy 😅 

### Testing steps

- [ ] Build plugin binary locally `cargo build -p forc-index --release`
- [ ] Create a new project using local binary `/path/to/fuel-indexer/target/release/forc-index new fool --namespace fuel`
- [ ] Build it `/path/to/fuel-indexer/target/release/forc-index build`
- [ ] You should notice that the comments remain in the same place after `forc index build`

### Changelog

- chore: forc-index-build preserve yaml comments

### DIFF
--- a/examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml
+++ b/examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml
@@ -32,7 +32,6 @@ graphql_schema:  examples/fuel-explorer/fuel-explorer/schema/fuel_explorer.schem
 # of the indexer.
 # Important: At this time, wasm is the preferred method of execution.
 module:
-
   wasm: target/wasm32-unknown-unknown/release/fuel_explorer.wasm
 
 # The resumable field contains a boolean that specifies whether or not the indexer should, synchronise
--- a/examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml
+++ b/examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml
@@ -32,6 +32,7 @@ graphql_schema:  examples/fuel-explorer/fuel-explorer/schema/fuel_explorer.schem
 # of the indexer.
 # Important: At this time, wasm is the preferred method of execution.
 module:
+
   wasm: target/wasm32-unknown-unknown/release/fuel_explorer.wasm
 
 # The resumable field contains a boolean that specifies whether or not the indexer should, synchronise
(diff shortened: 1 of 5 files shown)


examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml
```diff
@@ -32,6 +32,7 @@ graphql_schema:  examples/fuel-explorer/fuel-explorer/schema/fuel_explorer.schem
 # of the indexer.
 # Important: At this time, wasm is the preferred method of execution.
 module:
+
```
lostman: Why add an empty line here? Is this something the comment-preserving code is forcing us to do?

ra0x3: @lostman 
- Well the alternative to simply joining each line in the manifest on a newline, is - finding the particular lines that (we think) shouldn't have new lines above them (e.g., `    wasm: ~`) and removing the newline in-between this line and its precursor line
  - Nothing forcing us to _not_ do this, I just thought it was a little much for something small/subjective

lostman: I see. I think that's fine. What I expected is that nested items would not be separated by empty lines:
```yaml
toplevel:
  nested: 123

another-top-level:
  another-nested: 123
```

ra0x3: @lostman Let me push up a change

PR #1425, opened 2023-10-20
enhancement: full anyhow::Result<()> support in indexers

### LINKED ISSUE #1409: Add UserError WasmIndexerError code and extract an error message from WASM
### Description

Similarly to what was done in https://github.com/FuelLabs/fuel-indexer/issues/1402 with `Panic` and `panic_hook`, we should be able to change the indexer's handler functions' types to `-> anyhow::Result<()>`, and generate a wrapper which converts the `Err` variant into `WasmIndexerError::UserError`, while storing the message in a `static mut String` to be extracted and logged by `fuel-indexer`.

This would give the users another method of reporting errors, using `?`. However, due to error location attribution, panicking may be preferable.

It could also be helpful for https://github.com/FuelLabs/fuel-indexer/issues/1316.

lostman:
### Description

Closes #1409.

This PR changes the `indexer` macro to allow writing functions with a `Result` type and adds a mechanism to extract the error message from the WASM module. This is done the same way as with panic hook—by storing the message in a `static mut` `String`.

The error message automatically includes the block height as well as the name of the failing handler. Multiple errors from multiple handlers can be reported at once.

### Testing steps

Deploy an indexer that returns an `Err`:

```rust
use anyhow::Context;

#[indexer(manifest = "examples/fuel-explorer/fuel-explorer/fuel_explorer.manifest.yaml")]
pub mod explorer_index {

    // Returns Result
    fn index_block_3(_block_data: BlockData) -> Result<(), String> {
        Err("Bleep Bleep. Error. Bleep.".to_string())
    }

    // Or anyhow::Result
    fn index_block_2(block_data: BlockData) -> anyhow::Result<()> {
        index_block_2_inner(&block_data).with_context(|| {
            format!("failed at height {}", block_data.height)
        })
    }

    // Returns ()
    fn index_block(block_data: BlockData) {
        // stays the same
    }
}
```

Check the status:

```
```

It should display the error, and the context:

```
✅ Successfully fetched service health:

client status: OK
database status: OK
uptime: 2m 36s

Indexers:

─ fuellabs
   ├─ explorer
   |  • id: 107
   |  • created at: 2023-10-19 19:23:02.325661 UTC (4days 50m 11s ago)
   |  • status: error
   |  • status message:
   |      At height 1:
   |          index_block_3 failed with an error: Bleep Bleep. Error. Bleep.
   |          index_block_2 failed with an error: failed at height 1
   |
   └─ hello_world
      • id: 99
      • created at: 2023-10-17 16:06:58.930780 UTC (6days 4h 6m 14s ago)
      • status: running
      • status message:
          Indexed 940840 blocks
```

### Changelog

* Change the codegen to allow the handler function to either return `()` or `Result` type


deekerno (commented):
I personally like this change. However, I am a bit concerned with _requiring_ people to use a `Result`; most of that concern is from whether we can expect someone to know how to use it. Is there any way that a user can still do something like a regular `fn index_block(...)` without having the handler return a `Result`?

ra0x3 (changes requested):
- I'll echo @deekerno a bit here
- `anyhow::Result` should be changed to just `Result` in all cases where possible
- And since all handler functions are returning `Result<()>` this seems a bit...odd - as it's only used to report errors that might have happened
  - Not saying that's not important, but the UX seems a bit off

ra0x3 (dismissed):
utACK 🔥 Will defer to @deekerno

ra0x3 (approved):
- utACK 
- Deferring to @deekerno 
- Thanks @lostman

deekerno (approved):

Issue #1424, opened 2023-10-20
Unable to index starting from block 4000000


Walter-Phillips:
# Issue

Thanks for opening an issue on the Fuel Indexer project.

_*If you're an internal FuelLabs contributor, please feel free to remove this template when creating issues, as this template is meant for external users and contributors_.

## Prerequisites

- [x] Can you reproduce the problem?
- [x] Do you know which version of the indexer you're using?
- [x] Did you already check for an answer in previous [Discussions](https://github.com/FuelLabs/fuel-indexer/discussions)?

### Description

- When I put a start block in the indexer manifest file that is 4000000 or greater I get a wasm error
> Detailed information helps us narrow down any potential causes.

### Steps to Reproduce

Please list the steps required to reproduce your issue in detail.

1. Change the manifest.yaml to start at some block, I've been using 4500000
2. Deploy the indexer (I'm using some absolute paths you'll have to change, sorry)
3. After a few seconds you get this error:
![Screenshot 2023-10-19 at 4 49 20 PM](https://github.com/FuelLabs/fuel-indexer/assets/34078012/29c4e348-9d74-432f-97d2-c1c8b847a946)
This runs 10 times then fail like the message says 

**Expected behavior:** Able to index from any block

**Actual behavior:** Can only index from blocks less than 4000000

### Versions

Please list what version of indexer components you're using. Latest is currently <img src="https://img.shields.io/crates/v/fuel-indexer?label=latest" />.

- `fuel-indexer`: 0.21.0
- `forc-index`: 0.21.0

The staging branch should have everything you will need

Walter-Phillips:
The staging branch will have all the latest changes you would need

deekerno:
@Walter-Phillips What network are you on? And just to be sure, you can't **start** indexing at blocks greater than 4,000,000, but you're still able to index blocks greater than that, right?

Walter-Phillips:
Hey @deekerno we're on beta-4. I haven't let the indexer run long enough to get to blocks past 4,000,000 from block 1, so I'm not sure.

deekerno:
Can you start indexing from a number lower than 4,000,000 but still close to it (e.g. 3.9M, 3.5M, etc.) and see if it indexes past that point?

Walter-Phillips:
@deekerno, doing that now and will let you know how it goes.🫡

Walter-Phillips:
[...]

PR #1435, opened 2023-10-25
enhancement: extract better error messages from indexers

### LINKED ISSUE #1434: enhancement: improve error messages from indexers
### Description

Taking inspiration from  https://github.com/FuelLabs/fuel-indexer/issues/1424#issuecomment-1775866262, and building on https://github.com/FuelLabs/fuel-indexer/pull/1425, we can extract more information from the codegen for reporting errors.

lostman:
### Description

Closes #1434.

This PR improves the error messages from indexers. Block height, transaction ID, the name of decoded struct, failing handlers, etc., are automatically included in the error output.

This is implemented using `anyhow` with the error propagating upwards, accumulating more context.

Since we now use `anyhow` internally, it is also added to indexer's `Cargo.toml` (existing users will have to update manually).

Since #1425, users can also use `Result` type in the handler functions, so having anyhow already included should prove useful (the handlers can return `()`, or `Result`).

```

✅ Successfully fetched service health:

client status: OK
database status: OK
uptime: 44m 38s

Indexers:

┌─ fuellabs
|  ├─ explorer
|  |  • id: 107
|  |  • created at: 2023-10-19 19:23:02.325661 UTC (5days 16h 15m 35s ago)
|  |  • status: error
|  |  • status message:
|  |      Failed processing Block #128821
|  |
|  |      Caused by:
|  |          Failed executing index_block_3()
|  |
|  |          Caused by:
|  |              Another error.
|  |
|  |          Failed executing index_block_2()
|  |
|  |          Caused by:
|  |              Bleep blop Error. Bleep.
|  |
|  └─ hello_world
|     • id: 99
|     • created at: 2023-10-17 16:06:58.930780 UTC (7days 19h 31m 39s ago)
|     • status: running
|     • status message:
|         Indexed 1401240 blocks
|
└─ other
   └─ other_indexer
      • id: 108
      • created at: 2023-10-24 14:26:00.129558 UTC (21h 12m 37s ago)
      • status: error
      • status message:
          Failed processing Block #4000231

          Caused by:
              0: Failed processing Transaction 1d0fb5d1bee5534a1f904d9ff511eae80f219538a1d389728779106f2cf550b4
              1: [codegen] Failed decoding struct SomeStruct
              2: Invalid data: tried to read 64 bytes from response but only had 32 remaining!
```

### Testing steps

Same testing steps as #1425.

The following handlers can be added to `fuel-explorer`:

```rust
    fn index_block_3(_block_data: BlockData) -> anyhow::Result<()> {
        anyhow::bail!("Another error.".to_string())
    }

    fn index_block_2(_block_data: BlockData) -> anyhow::Result<()> {
        anyhow::bail!("Bleep blop Error. Bleep.".to_string())
    }
```

And then, after building and deploying, the `status` output should be as above.

The `other_indexer` is an example of failing codegen, which is harder to reproduce but can be done based on:
https://github.com/FuelLabs/fuel-indexer/issues/1424

### Changelog


examples/fuel-explorer/fuel-explorer/Cargo.toml
```diff
@@ -9,6 +9,7 @@ rust-version = "1.73.0"
 crate-type = ['cdylib']
 
 [dependencies]
+anyhow = "1"
```
ra0x3: @lostman 

- So when importing most external crates, we import those through the plugin, so that we don't have to add a ton of dependencies to the indexer itself
- So `anyhow` would be added to `fuel_indexer_plugin/lib.rs` (as well as to that prelude)
- `fuel_indexer_utils` automatically pulls in `fuel_indexer_plugin::prelude` - so if you add `anyhow` to the plugin, it will automatically be available in the indexers 👌🏼 
  - As an example reference, look at how we do this with `bincode`

ra0x3 (changes requested):
@lostman 
- The look of it is fine, however needs to be organized a bit differently

plugins/forc-index/src/defaults.rs
```diff
@@ -28,6 +28,7 @@ rust-version = "1.73.0"
 crate-type = ['cdylib']
 
 [dependencies]
+anyhow = "1"
```
ra0x3: @lostman We'll want to remove this as well

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -696,217 +702,231 @@ fn process_fn_items(
...
-                                        &Return {
-                                            contract_id: ContractId::from(<[u8; 32]>::from(id)),
-                                            val,
-                                            pc,
-                                            is
-                                        }
-                                    );
-                                    decoder.decode_type(ty_id, data);
+            use anyhow::Context;
+            use fuel_indexer_utils::plugin::types::fuel::TransactionData;
```
ra0x3: All fuel/explorer types are available via the plugin under the `fuel` namespace
```suggestion
            use fuel::TransactionData;
```

ra0x3 (commented):
@lostman Could you also include a QA report with this - for maybe 30K blocks each run - given how much of a critical path this PR touches?

lostman:
[Maciejs-MBP.Home-2023-10-25.indexer-qa.txt](https://github.com/FuelLabs/fuel-indexer/files/13172777/Maciejs-MBP.Home-2023-10-25.indexer-qa.txt)

@ra0x3 ⬆️

packages/fuel-indexer-macros/src/indexer.rs
```diff
@@ -696,217 +702,231 @@ fn process_fn_items(
...
+                decoder.dispatch()?;
 
                 let metadata = IndexMetadataEntity::new(block.time as u64, block.header.height, block.id);
                 metadata.save();
+                Ok(())
+            };
+
+            for block in blocks {
+                let block_height = block.header.height;
+                process_block(block).with_context(|| format!("Failed processing Block #{}", block_height))?;
```
deekerno, line 929: Shouldn't this come before we save the metadata? Otherwise, we have no guarantee that we actually processed a block prior to incrementing the metadata table.

deekerno (changes requested):

packages/fuel-indexer-macros/src/indexer.rs
lostman: Well caught. I inadvertently changed the control flow when moving the code into closures.

lostman: Or did I?

On `develop`:

```rust
                decoder.dispatch(block.header.height);

                let metadata = IndexMetadataEntity::new(block.time as u64, block.header.height, block.id);
                metadata.save();
```

here:

```rust
                decoder.dispatch()?;

                let metadata = IndexMetadataEntity::new(block.time as u64, block.header.height, block.id);
                metadata.save();
                Ok(())
```

The only difference is that the whole chunk is now inside the `process_block` closure.

deekerno: Ah yes, you're completely right. That's what I get for using the side-by-side diff view.

deekerno (approved):

ra0x3 (approved):

PR #1441, opened 2023-10-31
enhancement: add option for delaying requests to Fuel node

deekerno:
### Description

Adds a configuration flag to delay requests to a Fuel node. This is useful if a Fuel node makes use of rate limiting.

### Testing steps

Run any of the examples with the `--verbose` flag and an arbitrary value for the `--node-request-delay`. You should see that the service waits for the desired amount of time before processing and committing another group of blocks to the database.

### DIFF
--- a/packages/fuel-indexer-lib/src/config/cli.rs
+++ b/packages/fuel-indexer-lib/src/config/cli.rs
@@ -206,12 +206,12 @@ pub struct IndexerArgs {
     )]
     pub disable_toolchain_version_check: bool,
 
-    /// If present, the Fuel indexer service will wait for this particular duration between requests to a Fuel node.
+    /// Make the service wait for the given duration between block requests to a Fuel client.
     #[clap(
         long,
-        help = "If present, the Fuel indexer service will wait for this particular duration between requests to a Fuel node"
+        help = "Make the service wait for the given duration between block requests to a Fuel client."
     )]
-    pub node_request_delay: Option<u64>,
+    pub client_request_delay: Option<u64>,
 }
 
 #[derive(Debug, Parser, Clone)]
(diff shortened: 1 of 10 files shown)


ra0x3 (commented):
@deekerno Looks fine thanks for doing this. My only feedback is that this "node" term is confusing (we have `fuel-node` but it's been past time to change that). I believe the proper term for this is "fuel client" so maybe `client-request-delay` ?

packages/fuel-indexer-lib/src/config/cli.rs
```diff
@@ -205,6 +205,13 @@ pub struct IndexerArgs {
         help = "By default, Fuel Indexer will only accept WASM indexer modules compiled with the same toolchain version as the version of Fuel Indexer."
     )]
     pub disable_toolchain_version_check: bool,
+
+    /// If present, the Fuel indexer service will wait for this particular duration between requests to a Fuel node.
+    #[clap(
+        long,
+        help = "If present, the Fuel indexer service will wait for this particular duration between requests to a Fuel node"
```
ra0x3:
```suggestion
        help = "Force the indexer executor to wait for the specified duration between block requests to a Fuel client"
```

packages/fuel-indexer/src/executor.rs
```diff
@@ -301,6 +302,11 @@ pub fn run_executor<T: 'static + Executor + Send + Sync>(
 
             // Since we had successful call, we reset the retry count.
             consecutive_retries = 0;
+
+            if let Some(delay) = node_request_delay {
+                sleep(Duration::from_secs(delay)).await;
+                continue;
```
lostman:
```suggestion
```

Is this `continue` needed here? It is the end of the loop.

lostman (dismissed):

deekerno:
@ra0x3 Works for me, I'll make the  requisite changes.

ra0x3 (approved):

lostman (approved):

PR #1446, opened 2023-11-02
enhancement: implement find on entity

### LINKED ISSUE #1437: Implement `.find()` on `Entity`
### Context
- This is one of several pieces of work that will be a part of https://github.com/FuelLabs/fuel-indexer/issues/1417
- The use case is as a user, I want to find a single instance of an `Entity`
- Looking at how Prisma does this, I think they have something we can replicate with minimal complexity 
  - https://prisma.brendonovich.dev/reading-data/find

### What we currently have
- We currently `impl Entity` in our `fuel_indexer_macros::decoder` module as follows:

```rust
#[derive(Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
pub struct #ident {
    #struct_fields
}

impl<'a> Entity<'a> for #ident {
    const TYPE_ID: i64 = #type_id;
    const JOIN_METADATA: Option<[Option<JoinMetadata<'a>>; MAX_FOREIGN_KEY_LIST_FIELDS]> = #join_metadata;

    fn from_row(mut vec: Vec<FtColumn>) -> Self {
        #field_extractors
        Self {
            #from_row
        }
    }

    fn to_row(&self) -> Vec<FtColumn> {
        vec![
            #to_row
        ]
    }

}
```

### What we want
- We would update this ☝🏼 implementation to the following

```rust
#[derive(Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
pub struct #ident {
    #struct_fields
}

impl<'a> Entity<'a> for #ident {
    const TYPE_ID: i64 = #type_id;
    const JOIN_METADATA: Option<[Option<JoinMetadata<'a>>; MAX_FOREIGN_KEY_LIST_FIELDS]> = #join_metadata;

    #const_fields

    fn from_row(mut vec: Vec<FtColumn>) -> Self {
        #field_extractors
        Self {
            #from_row
        }
    }

    fn to_row(&self) -> Vec<FtColumn> {
        vec![
            #to_row
        ]
    }

}
```
- Explanation:
  - `#const_fields` would be a series of tokens for each field on the `Entity` as a `const`
 
So with the following schema
```graphql
type Foo @entity {
  id: ID!
  account: Address!
  name: String!
}
```
You'd get the following `const` fields on the `Entity`
```rust

trait SQLFragment {
  fn equals<T: Sized>(val: T) -> String;
}

struct FooId;

impl FooId {
  const name: &str = "id";
}

impl SQLFragment for FooId {
  fn equals<T: Sized>(val: T) -> String {
      format!("{} = '{}'", Self::name, val);
  }
}

struct FooAccount;

impl FooAccount {
  const name: &str = "account";
}

struct FooName;

impl FooName {
  const name: &str = "name";
}

impl<'a> Entity<'a> for Foo {
   const id: String = FooId;
   const account: Address =  FooAddress;
   const name: &str = FooName;

   // .. all the other code ...
}
```

This ☝🏼 would allow us to build a `.find()` method as follows:

```rust 
impl<'a> Entity<'a> for Foo {

   // .. all the other code ...

   fn find(fragments: Vec<impl SQLFragment>) -> Option<Self> {
      let where_clause = fragments.join(" AND ");
      let query = format!("SELECT * FROM {} WHERE {} LIMIT 1", self.name, where_clause);
      let buff = bincode::serialize(&query);
      let mut bufflen = (buff.len() as u32).to_le_bytes();
      let ptr = ff_arbitrary_single_select(Self::TYPE_ID, buff.as_ptr(), bufflen.as_mut_ptr());

       match ptr {
            Some(p) => {
                   let len = u32::from_le_bytes(bufflen) as usize;
                   let bytes = Vec::from_raw_parts(ptr, len, len).unwrap();
                   let data = deserialize(&bytes).unwrap();
                   Some(Self::from_row(data));
            }
            None => None,
       }
   }
}
```

- The idea is that `Entity::field::equals(value)` returns `String` fragments that can be aggregated into an arbitrary single `SELECT` query and easily passed to the DB via the FFI
- So the end result in your indexer handler would look like:

```rust
extern crate alloc;
use fuel_indexer_utils::prelude::*;

mod indexer_mod {
    fn handle_burn_receipt(order: OrderEntity) {
        // single find parameter
        let found_order = OrderEntity.find(vec![OrderEntity::amount::equals(5)])?; 

        // multiple find parameters
        let found_order = OrderEntity.find(vec![OrderEntity::amount::equals(5), OrderEntity::address::equals("0x00001")])?; 
    }
}
```

### Future work

- This would also allow us to update the `SQLFragment` trait to support other things such as 

```rust
trait SQLFragment {
  fn equals<T: Sized>(val: T) -> String;
  fn gt<T: Sized>(val: T) -> String;
  fn lt<T: Sized>(val: T) -> String;
  // .. so on and so forth ..
}
```

lostman:
### Description

Closes https://github.com/FuelLabs/fuel-indexer/issues/1437.

Example usage:

```rust
// Test searching for a string field
let f = FindEntity::find(
    FindEntity::string_value()
        .gt("f".to_string())
        .order_by_asc(FindEntity::value()),
)
```

This PR extends the codegen to generate field selectors for an `@entity`, e.g.:

```rust
1460   │     pub fn id() -> Field<ChainInfo, ID> {
1461   │         Field::new("id".to_string())
1462   │     }
1463   │     pub fn base_chain_height() -> Field<ChainInfo, U32> {
1464   │         Field::new("base_chain_height".to_string())
1465   │     }
1466   │     pub fn name() -> Field<ChainInfo, String> {
1467   │         Field::new("name".to_string())
1468   │     }
1469   │     pub fn peer_count() -> Field<ChainInfo, I32> {
1470   │         Field::new("peer_count".to_string())
1471   │     }
1472   │     pub fn latest_block() -> Field<ChainInfo, UID> {
1473   │         Field::new("latest_block".to_string())
1474   │     }
1475   │     pub fn consensus_parameters() -> Field<ChainInfo, UID> {
1476   │         Field::new("consensus_parameters".to_string())
1477   │     }
```

`Field<T>` structs implement methods such as `eq`, `gt`, `lt`, etc., which return a `Filter<T>`, which, in turn, can be combined using `and` and `or` to produce a mode complex `Filter<T>`. Finally, a `Filter<T>` can be ordered ascending or descending, turning it into a `QueryFragment<T>`.

The `find` function on `Entity` takes a `Filter` or `QueryFragment` and fetches the corresponding object from the database, if any exists.

### Testing steps

CI testing. This PR includes an integration test for added functionality.

### Changelog

* Add `find(query: impl Into<QueryFragment<Self>>) -> Option<Self>` to `Entity`.
* Add `ff_single_select` to FFI, and `single_select` to the indexer.
* Add `fuel_indexer_plugin::find` module, which contains the code for generating SQL query fragments.
* Decoder now generates `field_selectors`, e.g., `pub fn base_chain_height() -> Field<ChainInfo, U32>` which are used to construct queries


packages/fuel-indexer-plugin/src/find.rs
```diff
@@ -0,0 +1,298 @@
...
+/// WHERE {filter} {order_by}` statement that is assembled by the indexer to
+/// fetch an object from the database. The table name is not available to the
+/// plugin and thus only a part of the statment is generated there. The indexer
+/// maps the TYPE_ID to the tale name and assemles the full statemnt.
+pub struct QueryFragment<T> {
+    filter: Filter<T>,
+    order_by: Option<sql::OrderByExpr>,
+}
+
+/// Convert `QueryFragment` to `String`. `SELECT * from table_name` is lated
```
ra0x3:
```suggestion
/// Convert `QueryFragment` to `String`. `SELECT * from table_name` is later
```

packages/fuel-indexer-plugin/src/find.rs
```diff
@@ -0,0 +1,298 @@
...
+
+impl<T> Filter<T> {
+    fn new(filter: sql::Expr) -> Self {
+        Self {
+            filter,
+            phantom: std::marker::PhantomData,
+        }
+    }
+
+    pub fn and(self, right: Filter<T>) -> Filter<T> {
```
ra0x3, line 85: Elegant usage of `sqlparser` here ❤️

packages/fuel-indexer-plugin/src/find.rs
```diff
@@ -0,0 +1,298 @@
...
+            op: sql::BinaryOperator::Or,
+            right: Box::new(right.filter),
+        };
+        Filter {
+            filter,
+            phantom: std::marker::PhantomData,
+        }
+    }
+
+    pub fn order_by_asc<F>(self, f: Field<T, F>) -> QueryFragment<T> {
```
ra0x3: - So for this,  I think the more ORM way to do `order by` is to follow the SQL syntax.
- In SQL, order by is:

```
SELECT foo
FROM bar
ORDER BY {field} {direction}
```

- Where `{direction}` is `ASC/DESC` and `{field}` is obviously the field you're ordering by.
- I think functionally what you have is roughly the same as that ☝🏼 
- But syntactically, it's not "SQL"
- Some links
  - [`order_by`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.order_by) in Django
    - You first declare your field, then you declare how to order it
  - [`order_by`](https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#ordering-and-grouping) in Sequelize
    - Again, you first declare what to order, then how to order it 
  - Finally, [`order_by`](https://docs.diesel.rs/2.0.x/diesel/prelude/trait.QueryDsl.html#method.order_by) in Diesel 
    - Again, declaring what to order first, then how to order it

So I think a small change to something like:

```rust
Struct::find(Struct::field().eq(5).order_by(Struct::field().desc()); // like diesel
```
OR
```rust
Struct::find(Struct::field().eq(5).order_by(Struct::field(), Order::DESC));
```

packages/fuel-indexer-plugin/src/find.rs
```diff
@@ -0,0 +1,298 @@
...
+            filter: self,
+            order_by: Some(sql::OrderByExpr {
+                expr: sql::Expr::Identifier(sql::Ident::new(f.field)),
+                asc: Some(true),
+                nulls_first: None,
+            }),
+        }
+    }
+
+    pub fn order_by_desc<F>(self, f: Field<T, F>) -> QueryFragment<T> {
```
ra0x3: Same comment as above with regard to declaring the field first, then the ordering

packages/fuel-indexer-tests/contracts/fuel-indexer-test/Forc.lock
```diff
@@ -1,13 +1,13 @@
 [[package]]
-name = 'core'
-source = 'path+from-root-63B5D95B29B128A3'
+name = "core"
```
ra0x3, line 2: @lostman 
- So if you want to upgrade Sway to 0.46 is this PR that's fine with me - I believe beta-4 is on 0.46 now anyway
- But let's make sure to upgrade _all_ smart contracts to 0.46 in a single go (for the ones that aren't on 0.46 already, assuming there are any)

ra0x3 (changes requested):
- Wow @lostman very well done 👏🏼 
- Very clean and sets us up in a good spot to extend and continue the ORM work
- ⭐ ⭐⭐⭐⭐
- I left a comment about the syntax of your order clause. That's my only real feedback
- The rustdoc comments make this pretty easy to follow thankfully 🙏🏼

packages/fuel-indexer-plugin/src/find.rs
lostman: @ra0x3 

Sure, the syntax would be:

```rust
Struct::find(Struct::field().eq(5).order_by(Struct::field()).desc();
```

Or, we can add support for multiple fields from the get-go:

```rust
Struct::find(Struct::field().eq(5).order_by(vec![Struct::field()]).desc();
```

Or, this could also work, although my preference would be for the `vec!` as a cleaner solution:

```rust
Struct::find(Struct::field().eq(5).order_by(Struct::field()).order_by(Struct::another_field()).desc();
```

Finally, the syntax would allow for:

```rust
Struct::find(Struct::field().eq(5).order_by(Struct::field()).asc().desc().desc().asc();
```

Unless I add another struct that blocks it. It shouldn't be a problem for now though.

ra0x3 (approved):
🥇

packages/fuel-indexer-plugin/src/wasm.rs
```diff
@@ -124,6 +127,27 @@ pub trait Entity<'a>: Sized + PartialEq + Eq + std::fmt::Debug {
...
+            if !ptr.is_null() {
+                let len = u32::from_le_bytes(bufflen) as usize;
+                let bytes = Vec::from_raw_parts(ptr, len, len);
+                let data = deserialize(&bytes).unwrap();
+                Some(Self::from_row(data))
+            } else {
+                None
+            }
+        }
+    }
```
deekerno, line 149: Can we get rid of the `.unwrap()`s here? From what I can tell, the other functions in this module don't use them.

deekerno (approved):
Looks great, just left one comment about the usage of `.unwrap()`.

PR #1450, opened 2023-11-02
enhancement: revamp GraphQL query generation

### LINKED ISSUE #1145: Integrate web server, GraphQL layer, and data layer

deekerno:
### Description

This PR rewrites the indexer service's GraphQL query resolution layer.

The following breaking changes have been made:
- If annotated with the `@queryable` directive, each entity type has two queries generated for it which are the entry points for querying the indexer service:
-- a query returning a singular object record, which take at most one argument that matches to a field in the entity. _The chosen field must be annotated with the `@search` and `@unique` directives._
-- a query returning a `Connection` for the entity as described in the [GraphQL Cursor Connections specification](https://relay.dev/graphql/connections.htm).
- Input objects are now programmatically per type; this means that there is no longer general filtering and sorting behavior.
-- A connection query will allow for the use of a _sort_ input object as long as at least one of its entity's fields is sortable.
-- A connection query will allow for the use of a _filter_ input object as long as at least one of its entity's fields is marked with the `@filter` directive.
- Query responses are now spec-compliant with regards to the [Response Format](https://spec.graphql.org/October2021/#sec-Response-Format) section of the GraphQL specification.
- Offset-based pagination has been **retired** in favor of cursor-based pagination.
-- Paginated queries require a limit (by using the `first` keyword) and a sort order (by using the `order` keyword). Requesting the next or previous "page" can be done by passing a value from a `cursor` field from the response as a value for either the `after` or `before` keywords, respectively.

In addition to the above, the following non-breaking changes and fixes have been made:
- Adds support for [query variables](https://spec.graphql.org/October2021/#sec-Language.Variables).
- Adds support for the `@skip` and `@include` directives as required by [the relevant section of the GraphQL specification](https://spec.graphql.org/October2021/#sec-Type-System.Directives.Built-in-Directives).
- Adds support for [connection types](https://relay.dev/graphql/connections.htm); these are automatically generated for an entity.
- Adds a `@filter` directive for fields to denote which fields should be used for filtering connection queries.
- Adds support for executing multiple operations in one request.
- Adds proper introspection data generation for logical operators with filter input objects.
- Adjusts query string generation to use `Display` traits.
-- Queries are modeled through types and then converted to strings at the end.
- Adjusts `hello-world` example to be a simple explorer of blocks with lists of transactions.
- Cleans up organization of `fuel-indexer-graphql` module.
- Fixes querying for list type fields on an object.
- Fixes issue with database joins not being in correct sorted order.
- Moves introspection type generation from `dynamic.rs` to the GraphQL schema parsing step.
- Reduces use of `format!()` calls in query generation.

### Testing steps

First, build the Fuel explorer example; you can use `bash scripts/utils/build_modules.bash` to build all examples. Then, start the Fuel explorer example with the service (be sure to adjust the Postgres details if they differ from the following command):
```
cargo run -p fuel-indexer -- run --fuel-node-host beta-4.fuel.network --fuel-node-port 80 --database postgres --postgres-host 127.0.0.1 --postgres-port 5432 --postgres-user postgres --postgres-password 'my-secret' --postgres-database 'postgres'
```

_Examples coming soon..._

### Notes

Yes, this PR is quite large. However, given that the block explorer will be served by an indexer backend, we needed to support as much GraphQL functionality as we possibly can; we do not have the luxury of enshrining certain types and queries as part of our code so we need to be able to support anything that a user may feasibly create.


examples/hello-world/hello-world/schema/hello_world.schema.graphql
```diff
@@ -1,12 +1,36 @@
-type Block @entity {
+type SubExampleEntity @entity {
```
ra0x3: Can we leave the `hello-world` example as is? Ideally the only indexer that gets update from this is just our test indexer

packages/fuel-indexer-graphql/src/query.rs
```diff
@@ -0,0 +1,684 @@
...
+                    Some((parent_node, child_node)),
+                    Some((referring_field, fk_field)),
+                ) = (self.graph.edge_endpoints(e), self.graph.edge_weight(e))
+                {
+                    if let (Some(parent_table), Some(child_table)) = (
+                        self.graph.node_weight(parent_node),
+                        self.graph.node_weight(child_node),
+                    ) {
+                        let join = format!(
+                            "INNER JOIN {} ON {}.{} = {}.{}",
```
ra0x3: Can we extract this into some struct instead of the string hacking? This would help make things a bit more maintainable as you'd know what are the components of this `InnerJoin` and you could always `impl Display for InnerJoin` in order to get the SQL fragment (or `impl SqlFragment for InnerJoin`)

packages/fuel-indexer-graphql/src/query.rs
```diff
@@ -0,0 +1,684 @@
...
+                            query_parameters.sorts.append(&mut params.sorts);
+
+                            if field_iter.peek().is_some() {
+                                selections.push(", ".to_string());
+                            }
+                        }
+
+                        let joins = dep_graph.get_sorted_joins()?;
+
+                        Ok(format!(
```
ra0x3: Same comment here with regarding to the `String` usage.

packages/fuel-indexer-graphql/src/query.rs
```diff
@@ -0,0 +1,684 @@
...
+                let fqn = schema.fully_qualified_namespace();
+                match self {
+                    SelectionNode::Scalar {
+                        name,
+                        alias,
+                        parent_entity,
+                    } => {
+                        let field_name = alias
+                            .clone()
+                            .map_or(format!("'{}'", name), |a| format!("'{}'", a.node));
```
ra0x3: In general, we would _love_ to avoid as much of this `format!` as possible. As mentioned in the other comments, if we could abstract this stuff behind informative structs or functions that would make it a bit more understandable and maintainable.

ra0x3 (changes requested):
@deekerno 

- Round 1 finished. 
- I just left comments on some high level maintainability stuff - in general I think we want to avoid `format!` here as much as possible and abstract these (what are effectively just `String`s into structs and functions). Seems a bit overkill but there's a lot of business logic in this `queries` module 
- Could you also update *Testing Steps* with instructions on how I would test the block explorer?
  - We will probably have some non-indexer folks test the queries as well

PR #1469, opened 2023-11-22
enhancement: add find_many

### LINKED ISSUE #1467: Implement `.find_many()` on `Entity`
- With the completion of https://github.com/FuelLabs/fuel-indexer/issues/1437 we now have `.find()` usage
- The next thing we want is the implementation of `.find_many()`
- `.find_many()` would work _just_ like `.find()` except it instead of returning `Option<Entity>` it would return `Option<Vec<Entity>>`
- We would have to establish some constraints/parameters around the size of the result set of `.find_many()`
  - Since we allocate space in the runtime for the result of `.find()` (and `.get()`) we would need to set some max size that can be allocated for `.find_many()` results
    - Maybe this is configurable?

lostman:
### Description

Closes #1467.

This PR adds the `find_many()` function to the `Entity` trait, which is like `find()`, except that it fetches multiple results.

### Testing steps

CI tests should pass. This PR adds basic `find_many` tests to the existing `find` test.

### Changelog

* Add `find_many` to the `Entity` trait.
* Rename `QueryFragment` to `OrderedFilter`. This way we have `Filter` and `OrderedFilter`, which easier to understand.
* Clean up some implementation details.


packages/fuel-indexer-plugin/src/wasm.rs
```diff
@@ -128,22 +128,39 @@ pub trait Entity<'a>: Sized + PartialEq + Eq + std::fmt::Debug {
...
     /// Finds the first entity that satisfies the given constraints.
-    fn find(query: impl Into<QueryFragment<Self>>) -> Option<Self> {
-        let query: QueryFragment<Self> = query.into();
+    fn find(query: impl ToFilter<Self>) -> Option<Self> {
+        let result = Self::find_many(1, query);
+        result.into_iter().next()
+    }
+
+    /// Finds the entities that satisfy the given constraints.
+    fn find_many(limit: usize, query: impl ToFilter<Self>) -> Vec<Self> {
```
ra0x3: - So the semantics of this `find_many` are a bit off.
- The `LIMIT` clause in SQL works similar to the `ORDER BY` clause
- In the `find` implementation https://github.com/FuelLabs/fuel-indexer/pull/1446, we added the `order_by` clause as a separate method

```rust
Entity::find(Entity::field().eq(value).order_by(Entity::field()).asc())
```
- The `LIMIT` clause works in a similar way - as a separate method
```rust
Entity::find(Entity::field().eq(value).order_by(Entity::field()).asc().limit(5))
```
- With that said, `.limit()` is a separate piece of functionality from `.find_many`
  - Totally fine if you want to go ahead and throw both of those in the same PR
  - When I mentioned "limit the result set" in the original issue, I was more so making sure we're aware of how much space we're allocating in the runtime for the result set

References
- [Sequelize `limit`](https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#limits-and-pagination)
- [Django](https://docs.djangoproject.com/en/dev/topics/db/queries/#limiting-querysets) is actually notoriously known for not having a `.limit()` API 😆 
- [Diesel `.limit()`](https://docs.diesel.rs/2.0.x/diesel/prelude/trait.QueryDsl.html#method.limit)

ra0x3 (changes requested):
@lostman 

- Left feedback regarding the usage of `limit` in your `find_many` implementation
- Will review again once that's fixed (since I think fixing that will shift some logic around)

packages/fuel-indexer-plugin/src/wasm.rs
lostman: Thanks @ra0x3. I've changed the API to be like you described.

ra0x3 (approved):
Thanks @lostman

deekerno (approved):

PR #1522, opened 2024-01-30
fix: include `std` feature on `strum` dependency

deekerno:
### Description

Includes `std` feature for `strum` in `fuel-indexer-api-server`.

### Testing steps

For some odd reason, I was unable to build the `fuel-indexer` package due to the following error:
```
error[E0599]: the method `as_dyn_error` exists for reference `&ParseError`, but its trait bounds were not satisfied
   --> packages/fuel-indexer-api-server/src/api.rs:105:18
    |
105 |     ParseError(#[from] strum::ParseError),
    |                  ^^^^ method cannot be called on `&ParseError` due to unsatisfied trait bounds
    |
```

Try to build on develop, and then try to build on this branch; you should see that it works in the latter.


ra0x3 (approved):
utACK

